PR #118, opened 2019-06-13
docs: organized code into modules

tjwelde:


tjwelde:
@LeonFLK yes they were redundant. Must've slipped in, while adding more imports to the top of the file.

Issue #144, opened 2019-08-27
Resolve circular dependency in delegation module around `decodeChildNode` in another way


tjwelde:
Right now, it uses inheritance, but the DelegationBaseNode should really just have the method.

PR #145, opened 2019-08-27
feat: move from tslint to typescript-eslint

### LINKED ISSUE #132: Migrate to EsLint
As TsLint [will be deprecated throughout this year](https://github.com/palantir/tslint/issues/4534) in favor of `typescript-eslint`, I would recommend to [move to EsLint](https://blog.matterhorn.dev/posts/learn-typescript-linting-part-1/) instead, or at least keep the migration in mind. See [here](https://github.com/typescript-eslint/typescript-eslint) for more information.

tjwelde:
fixes #132 

This PR moves the codebase from tslint to typescript-eslint. It also refactors the whole codebase to meet new coding standard


src/ctype/CType.chain.ts
```diff
@@ -26,7 +26,10 @@ export async function store(
   > = await blockchain.api.tx.ctype.add(ctype.hash)
   const txStatus: TxStatus = await blockchain.submitTx(identity, tx)
   if (txStatus.type === 'Finalised') {
-    ctype.owner = identity.address
+    txStatus.payload = {
+      ...ctype,
+      owner: identity.address,
+    }
   }
   return txStatus
```
maudnals: If the transaction isn't finalized, then the `txSatus` doesn't have a `payload` property.    
Are we not accessing `payload` and using it somewhere else in the codebase, which could potentially crash if `payload` is `undefined` ?  

(this is just a thought exercise, since I'm guessing that case was covered already in the previous implementation)

src/config/ConfigLog.ts
```diff
@@ -17,6 +17,7 @@ const options = new LoggerFactoryOptions().addLogGroupRule(
 
 // Create a named loggerfactory and pass in the options and export the factory.
 // Named is since version 0.2.+ (it's recommended for future usage)
+// eslint-disable-next-line import/prefer-default-export
 export const factory = LFService.createNamedLoggerFactory(
```
maudnals: Why can't we use default export instead of disabling the rule?

src/ctype/CType.ts
```diff
@@ -32,7 +34,7 @@ export default class CType implements ICType {
     }
   }
 
-  public async store(identity: Identity) {
+  public async store(identity: Identity): Promise<TxStatus> {
```
maudnals: General remark on the whole PR: this update is super helpful, it's great to be able to see the return types 👍

src/ctype/CTypeUtils.ts
```diff
@@ -6,41 +6,38 @@ import Ajv from 'ajv'
...
-  ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-07.json'))
+  const ajv = new Ajv()
   ajv.addMetaSchema(CTypeModel)
   const result = ajv.validate(metaModel, model)
   if (!result && ajv.errors) {
-    ajv.errors.map((error: any) => {
-      if (messages) {
+    if (messages) {
+      ajv.errors.forEach((error: any) => {
         messages.push(error.message)
```
maudnals: Not purely part of this PR, so that's nitpicking / just a question:

Why not using functional instead?
```javascript  
if (messages) {
  messages = [...messages, ...ajv.errors.map(e => e.message)]; 
}
```

src/messaging/Message.ts
```diff
@@ -44,31 +44,61 @@ export interface IEncryptedMessage {
...
+        {
+          const submitClaimsForCtype = message.body
+          submitClaimsForCtype.content.forEach(claim => {
+            if (claim.request.claim.owner !== message.senderAddress) {
+              throw new Error('Sender is not owner of the claims')
+            }
+          })
+        }
         break
+      default:
```
maudnals: Cool, I didn't know that the `default` case could be empty!

src/requestforattestation/RequestForAttestation.ts
```diff
@@ -90,17 +91,17 @@ export default class RequestForAttestation implements IRequestForAttestation {
     this.claimerSignature = this.sign(identity)
   }
 
-  public removeClaimProperties(properties: string[]) {
+  public removeClaimProperties(properties: string[]): void {
     properties.forEach(key => {
-      if (!this.claimHashTree.hasOwnProperty(key)) {
+      if (!this.claimHashTree[key]) {
```
maudnals: Not sure whether that can happen, but what if `this.claimHashTree[key]` exists and is `0` or `false`?

src/attestation/Attestation.chain.ts
```diff
@@ -29,23 +29,31 @@ export async function store(
...
+      claimHash,
+      cTypeHash: attestationTuple[0],
+      owner: attestationTuple[1],
+      delegationId: attestationTuple[2],
+      revoked: attestationTuple[3],
+    }
+    log.info(`Decoded attestation: ${JSON.stringify(attestation)}`)
+    return Attestation.fromObject(attestation)
+  }
+  return undefined
```
maudnals: Nothing that's in the scope of this PR, but:    
General question: why are we using `undefined` over `null` in the SDK codebase?  
 
So far I was told that `undefined` should rather be left to the compiler to signify actual `undefined` errors, and use `null` instead.

.eslintrc.json
```diff
@@ -0,0 +1,57 @@
...
+        }],
+        "@typescript-eslint/interface-name-prefix": ["error", "always"],
+        "@typescript-eslint/prefer-interface": "off",
+        "@typescript-eslint/array-type": ["error", "array-simple"]
+    },
+    "overrides": [
+        {
+            "files": ["**/*.spec.ts"],
+            "env": {
+                "jest": true
```
maudnals: Not a suggestion, but a question: why do we need to list jest in the `"overrides"` and not in the `"env"` above?

src/delegation/DelegationRootNode.ts
```diff
@@ -21,7 +21,9 @@ export default class DelegationRootNode extends DelegationBaseNode
    * @param delegationId unique identifier of the delegation root
    * @returns promise containing [[DelegationRootNode]] or [undefined]
    */
-  public static async query(delegationId: string) {
+  public static async query(
+    delegationId: string
+  ): Promise<DelegationRootNode | undefined> {
```
maudnals: Same question regarding `undefined` vs `null`

src/delegation/DelegationRootNode.ts
```diff
@@ -48,10 +50,11 @@ export default class DelegationRootNode extends DelegationBaseNode
     return Promise.resolve(this)
   }
 
-  // tslint:disable-next-line:prefer-function-over-method
+  /* eslint-disable class-methods-use-this */
```
maudnals: Why do we need to disable the rule instead of making the method static?

src/did/Did.ts
```diff
@@ -121,17 +123,17 @@ export default class Did implements IDid {
       id: this.identifier,
       authentication: {
         type: 'Ed25519SignatureAuthentication2018',
-        publicKey: [this.identifier + '#key-1'],
+        publicKey: [`${this.identifier}#key-1`],
       },
       publicKey: [
         {
-          id: this.identifier + '#key-1',
+          id: `${this.identifier}#key-1`,
```
maudnals: feels gooood   

feelsGood

src/errorhandling/ErrorHandler.ts
```diff
@@ -85,16 +84,16 @@ export class ErrorHandler {
...
+    let index = 0
+    filtered.forEach(m => {
       if (m.name.toString() === ErrorHandler.ERROR_MODULE_NAME) {
-        break
+        return
       }
-      index++
-    }
+      index += 1
+    })
```
maudnals: Not in the scope of this PR, but maybe this could be more compact/declarative: 
  
```javascript  
// find the last module error
return filtered.map(m => m.name.toString()).lastIndexOf(ErrorHandler.ERROR_MODULE_NAME); 
```   

(or on several lines if more readable)

src/ctype/CTypeUtils.ts
```diff
@@ -6,41 +6,38 @@ import Ajv from 'ajv'
...
-      }
-    })
+      })
+    }
   }
-  return result ? true : false
+  return !!result
```
maudnals: so nice ✨

.eslintignore
```diff
@@ -0,0 +1 @@
+__mocks__
```
maudnals: Not a suggestion, but a question: why?

tjwelde: Mocks normally have to abuse the guidelines quite heavily, to mock a class. We could have some mock specific rules, though. Just not that restrictive.

src/delegation/DelegationRootNode.ts
tjwelde: I actually just wanted to move the method to the BaseNode, but there is a problem with circular dependencies. I opened a ticket to resolve this issue: https://github.com/KILTprotocol/sdk-js/issues/144

src/attestation/Attestation.chain.ts
tjwelde: yes, we should switch to null

src/delegation/DelegationRootNode.ts
tjwelde: we should switch to null

.eslintrc.json
tjwelde: Because jest environment only exist in the context of a test

src/ctype/CType.chain.ts
tjwelde: yeah, this code is weird. if it is not finalized, it will actually throw an error. We have to refactor it here.

src/config/ConfigLog.ts
tjwelde: I didn't want to change the API (I know, I already did this with changing the ctype/txstatus/payload thingy)

tjwelde: also the whole naming is weird here. I would rather move this logger to a util folder and use it from there

src/ctype/CType.ts
tjwelde: Awesome

src/requestforattestation/RequestForAttestation.ts
tjwelde: this can't happen. But maybe the claimHashTree should be better defined, type-wise

src/errorhandling/ErrorHandler.ts
tjwelde: uhhh, nice

.eslintignore
maudnals: Got it, thanks!

src/ctype/CType.chain.ts
tjwelde: https://github.com/KILTprotocol/sdk-js/issues/149

src/config/ConfigLog.ts
tjwelde: https://github.com/KILTprotocol/sdk-js/issues/150

src/ctype/CTypeUtils.ts
tjwelde: This will not work, since we don't want to reassign parameters.
I will make a ticket, where we will change the return type of this function, to return either true, or messages.

tjwelde: https://github.com/KILTprotocol/sdk-js/issues/151

src/requestforattestation/RequestForAttestation.ts
tjwelde: https://github.com/KILTprotocol/sdk-js/issues/152

src/attestation/Attestation.chain.ts
tjwelde: https://github.com/KILTprotocol/sdk-js/issues/153

src/delegation/DelegationRootNode.ts
tjwelde: https://github.com/KILTprotocol/sdk-js/issues/153

src/delegation/DelegationRootNode.ts
tjwelde: Moved to https://github.com/KILTprotocol/ticket/issues/4

PR #159, opened 2019-09-02
fix: decrypted message has to support senderboxpublickey

LeonFLK:
IMessage now has optional parameter senderboxpublickey. Should resolve the failed build.

### DIFF
--- a/src/messaging/Message.ts
+++ b/src/messaging/Message.ts
@@ -24,6 +24,7 @@ export interface IMessage {
   createdAt: number
   receiverAddress: IPublicIdentity['address']
   senderAddress: IPublicIdentity['address']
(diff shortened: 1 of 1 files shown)


src/messaging/Message.ts
```diff
@@ -24,6 +24,7 @@ export interface IMessage {
   createdAt: number
   receiverAddress: IPublicIdentity['address']
   senderAddress: IPublicIdentity['address']
+  senderBoxPublicKey?: IPublicIdentity['boxPublicKeyAsHex']
```
tjwelde: Let's have it required here

LeonFLK: okay updated

PR #190, opened 2019-09-13
docs: guidelines

maudnals:
## How to test:
No test steps required, just some reading. 

## Checklist:

No code in this PR, markdown only.

### DIFF
--- a/docs/contribution-guide.md
+++ b/docs/contribution-guide.md
@@ -12,7 +12,7 @@ KILT's API doc is available [here][apidoc].
 
 In the KILT SDK, modules and public methods are documented directly in the code, in what is called *docBlocks*. DocBlocks are special comments which document the code they precede. DocBlocks need to follow some linting rules. Make sure ESLint is activated in your code editor, so the errors and warnings are highlighted.
 
-We also check these rules on commit, so that you're protected: you can't push a linting error to the repo.
+We also check these rules on commit and in our CI pipeline, so that you're protected: you can't push a linting error to the repo.
 
 You don't need to - but if you're interested, you can see all the linting rules for the inline docBlocks in `.eslintrc.json`, and check what they mean on [jsdoc][jsdoc].
 
@@ -55,7 +55,8 @@ On top of the the linting rules mentioned above (must document all parameters, m
     * Expected output;
     * Suggestion for the next step;
     * ...
-  * Don't hesitate to include comments.
+  * Don't hesitate to include comments;
+  * Keep it short! No need to overdo it: just the bit of context required to understand the method call is enough.
 
 💡The linting rules for the example snippet are **not** the same as the SDK codebase linting rules. For example, the example snippet should make use of semicolumns. You can see the full ruleset in `.eslintrc-jsdoc.json`, but the linter should be enough to help you figure the rules out.
 
@@ -69,8 +70,10 @@ Example of a method docBlock that follows these guidelines:
  * @param identity - The identity used to revoke the attestation. It should be an attester identity, or an identity with delegated rights.
  * @returns A promise containing the [[TxStatus]] (transaction status).
  * @example ```javascript
- * // revoke the attestation mapped to the claim hash "0xd8024cdc147c4fa9221cd177" with `identity`. To create `identity`, see `buildFromMnemonic` and `generateMnemonic` in the `Identity` class. The attestation can not be un-revoked.
+ * // revoke the attestation mapped to the claim hash "0xd8024cdc147c4fa9221cd177". To create `identity`, see `buildFromMnemonic` and `generateMnemonic` in the `Identity` class. Note that `identity` should have revokation rights.
  * Attestation.revoke("0xd8024cdc147c4fa9221cd177", identity);
+ * Attestation.query("0xd8024cdc147c4fa9221cd177").then((attestation) => console.log(attestation.revoked));
+ * // should log `true`
  * ```
  */
 ```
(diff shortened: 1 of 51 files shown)


docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
 # Contribution Guide
 
-## NB
+## Documentation
+
+We want the KILT SDK to be easy to use for all fellow developers. We ❤️ well-documented code.
+
+KILT's API doc is available [here][apidoc].
+
+In the KILT SDK, modules and public methods are documented inline, in the code. These inline "doc blocks" need to follow some linting rules. Make sure ESLint is activated in your code editor, so the errors and warnings are highlighted.
```
tjwelde: I think of `inline` as 
```
var bla = 'keks'; // my inline comment
```
not the one on top of a method. Maybe there is a better wording here?

tjwelde: The "doc blocks" are actually `jsdoc` block comments

docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
+
+We also check these rules on commit, so that you're protected: you can't push a linting error to the repo.
```
tjwelde: We also check it in the CI!

docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
+
+You don't need to - but if you're interested, you can see all the linting rules for the inline doc blocks in `.eslintrc.json`, and check what they mean on [jsdoc][jsdoc].
```
tjwelde: ~inline~

docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
+* `@example`:
+  * Create it as valid code 😎;
+  * Illustrate only this method's functionality, but also provide enough context for fellow developers to try it out:
+    * Imports;
+    * Preparation step;
+    * Actual function call;
+    * Expected output;
+    * Suggestion for the next step;
+    * ...
+  * Don't hesitate to include comments.
```
tjwelde: Let's also say, that this doesn't have to be overdone. Just a bit of needed context to understand the method call is enough

docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
+```javascript
+/**
+ * [STATIC] [ASYNC] Revokes an attestation.
+ *
+ * @param claimHash - The hash of the claim that corresponds to the attestation to revoke.
+ * @param identity - The identity used to revoke the attestation. It should be an attester identity, or an identity with delegated rights.
+ * @returns A promise containing the [[TxStatus]] (transaction status).
+ * @example ```javascript
+ * // revoke the attestation mapped to the claim hash "0xd8024cdc147c4fa9221cd177" with `identity`. To create `identity`, see `buildFromMnemonic` and `generateMnemonic` in the `Identity` class. The attestation can not be un-revoked.
+ * Attestation.revoke("0xd8024cdc147c4fa9221cd177", identity);
```
tjwelde: Let's have an example with context here, since we talked about that on top.

docs/contribution-guide.md
```diff
@@ -1,19 +1,104 @@
...
+ */
+```
+
+### Checking locally how the [online API Doc][apidoc] will look like
+
+You probably don't need to do this. You can trust the doc will be rendered properly.
+If you have a doubt:
+
+* Run `yarn build:docs` within the `sdk-js` folder. This generates the doc at `sdk-js/docs/api`.
+* Open any of the generated files in your browser, such as `sdk-js/docs/api/classes/attestation.attestation-1.html`. You can now use the menu or inline links to navigate across modules and classes.
```
tjwelde: It's enough to open `sdk-js/docs/api/index.html`

docs/contribution-guide.md
```diff
@@ -1,19 +1,107 @@
...
+
+We want the KILT SDK to be easy to use for all fellow developers. We ❤️ well-documented code.
+
+KILT's API doc is available [here][apidoc].
+
+In the KILT SDK, modules and public methods are documented directly in the code, in what is called *docBlocks*. DocBlocks are special comments which document the code they precede. DocBlocks need to follow some linting rules. Make sure ESLint is activated in your code editor, so the errors and warnings are highlighted.
+
+We also check these rules on commit and in our CI pipeline, so that you're protected: you can't push a linting error to the repo.
+
+You don't need to - but if you're interested, you can see all the linting rules for the inline docBlocks in `.eslintrc.json`, and check what they mean on [jsdoc][jsdoc].
```
tjwelde: Ok, let's clear this up:
[jsdoc](https://jsdoc.app/)  [(wiki)](https://en.wikipedia.org/wiki/JSDoc) is a spec and parser for javascript docblocks.
[tsdoc](https://github.com/microsoft/tsdoc) is the equivalent to that from microsoft for typescript and   
[eslint-plugin-jsdoc](https://www.npmjs.com/package/eslint-plugin-jsdoc) is an eslint plugin (normally for jsdoc) we use to lint our tsdoc docblock comments.

So to check, what the options mean, they have to look at `eslint-plugin-jsdoc` and not at `jsdoc`

maudnals: Got it, thanks. Fixed - I changed it to display eslint-plugin-jsdoc (the link is still right). The details you've mentioned are perfect for me to understand - however I left them out here since I feel contributors would only need to check eslint-plugin-jsdoc... what do you think?

tjwelde: I think we should mention that we use tsdoc/jsdoc.
It's good, for people who know these and people who don't can just read further.

maudnals: Good idea - Updated!

PR #196, opened 2019-09-23
refactor: updated 'any' to something meaningful

### LINKED ISSUE #164: SDK - Change `any` types to something meaningful

Dudleyneedham:
## fixes
#164 
Updates 'any' types to something meaningful

src/claim/Claim.ts
src/ctype/CType.ts 
src/ctype/CTypeUtlis.ts 
src/requestforattestation/RequestForAttestation.ts
src/types/CType.js
src/types/Claim.js

Updated state setup in test suites according to update

Update the any to match what the types accordingly

Changed some names of arguments in response to checking through the code

During this time, found something I would like to tackle. #200 Changing from Arrays to Objects

## How to test:

yarn install
yarn test -> should be Test Suites: 1 skipped, 17 passed, 17 of 18 total
Tests: 2 skipped, 50 passed, 52 total
## Checklist:

### DIFF
--- a/src/blockchain/Blockchain.ts
+++ b/src/blockchain/Blockchain.ts
@@ -37,7 +37,7 @@ export interface IBlockchainApi {
   api: ApiPromise
 
   getStats(): Promise<Stats>
-  listenToBlocks(listener: (header: Header) => void): Promise<any> // TODO: change any to something meaningful
+  listenToBlocks(listener: (header: Header) => void): SubscriptionResult // Changes this into a Promise function for any, could be better to dig deeper
   submitTx(
     identity: Identity,
     tx: SubmittableExtrinsic<CodecResult, SubscriptionResult>
@@ -50,6 +50,7 @@ export interface IBlockchainApi {
 
 export default class Blockchain implements IBlockchainApi {
   public static asArray(queryResult: QueryResult): any[] {
+    // I think this can be changed. Was originally any.
     const json =
       queryResult && queryResult.encodedLength ? queryResult.toJSON() : null
     if (json instanceof Array) {
--- a/src/blockchain/Blockchain.ts
+++ b/src/blockchain/Blockchain.ts
@@ -37,7 +37,7 @@ export interface IBlockchainApi {
   api: ApiPromise
 
   getStats(): Promise<Stats>
-  listenToBlocks(listener: (header: Header) => void): SubscriptionResult 
+  listenToBlocks(listener: (header: Header) => void): SubscriptionResult
   submitTx(
     identity: Identity,
     tx: SubmittableExtrinsic<CodecResult, SubscriptionResult>
--- a/src/blockchain/TxStatus.ts
+++ b/src/blockchain/TxStatus.ts
@@ -7,7 +7,7 @@
  */
 
 import { Codec } from '@polkadot/types/types'
-import ICType from 'src/types/CType'
+import ICType from '../types/CType'
 
 export type QueryResult = Codec | undefined | null | ICType | string
 
--- a/src/blockchain/TxStatus.ts
+++ b/src/blockchain/TxStatus.ts
@@ -5,11 +5,17 @@
 /**
  * Dummy comment needed for correct doc display, do not remove
  */
+
+import { Codec } from '@polkadot/types/types'
+import ICType from 'src/types/CType'
+
+export type QueryResult = Codec | undefined | null | ICType | string
+
 export default class TxStatus {
   public type: string
-  public payload: any
+  public payload: QueryResult
 
-  public constructor(type: string, payload?: any) {
+  public constructor(type: string, payload?: QueryResult) {
     this.type = type
     this.payload = payload
   }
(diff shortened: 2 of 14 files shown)


Dudleyneedham:
Changed every any into something a little more meaningful. 

I am unsure about one, that seems to need to be an any. 

It is a JSON call again from the blockchain.

I searched the Polkadot api and found a lot of help within there.

Dudleyneedham:
Made changes accordingly.

src/blockchain/Blockchain.ts
```diff
@@ -37,7 +37,7 @@ export interface IBlockchainApi {
   api: ApiPromise
 
   getStats(): Promise<Stats>
-  listenToBlocks(listener: (header: Header) => void): Promise<any> // TODO: change any to something meaningful
+  listenToBlocks(listener: (header: Header) => void): SubscriptionResult
```
Dudleyneedham: The subscriptionResult is a type given from the Polkadot API, I don't think it actually suits here. 

I will look at the type here again.

src/blockchain/Blockchain.ts
```diff
@@ -50,6 +50,7 @@ export interface IBlockchainApi {
 
 export default class Blockchain implements IBlockchainApi {
   public static asArray(queryResult: QueryResult): any[] {
+    // I think this can be changed. Was originally any may need help with this
```
Dudleyneedham: I was unsure how to proceed with this any, it would be a good sparring session.

src/blockchain/TxStatus.ts
```diff
@@ -5,11 +5,17 @@
 /**
  * Dummy comment needed for correct doc display, do not remove
  */
+
+import { Codec } from '@polkadot/types/types'
+import ICType from '../types/CType'
+
+export type QueryResult = Codec | undefined | null | ICType | string
```
Dudleyneedham: After searching deep into the types from Polkadot, I looked into Codec and what it provides, then added some necessaries. Though the null and undefined may not be essential. Sparring on this topic would be ideal.

src/claim/Claim.ts
```diff
@@ -18,7 +18,7 @@ import { verifyClaimStructure } from '../ctype/CTypeUtils'
 import Identity from '../identity/Identity'
 import IClaim from '../types/Claim'
 
-function verifyClaim(claimContents: object, cType: CType): boolean {
+function verifyClaim(claimContents: IClaim['contents'], cType: CType): boolean {
```
Dudleyneedham: I changed the naming of the argument as it wasn't a claim but the Contents. Which allowed me to add the IClaim contents.

src/ctype/CType.chain.ts
```diff
@@ -26,7 +22,7 @@ export async function store(
   log.debug(() => `Create tx for 'ctype.add'`)
   const tx: SubmittableExtrinsic<
     CodecResult,
-    any
+    SubscriptionResult
```
Dudleyneedham: All the SubmittableExtrinsic come from the Polkadot api and the SubscriptionResult is what they used. It seemed a natural fit. It is a Promise<any>.

src/ctype/CType.ts
```diff
@@ -48,8 +47,8 @@ export default class CType implements ICType {
     return store(this, identity)
   }
 
-  public verifyClaimStructure(claim: any): boolean {
-    return CTypeUtils.verifySchema(claim, this.schema)
+  public verifyClaimStructure(claimContents: IClaim['contents']): boolean {
+    return CTypeUtils.verifySchema(claimContents, this.schema)
```
Dudleyneedham: Again, as mentioned above,  changed it to claimContents as it fits.

src/types/CType.ts
```diff
@@ -7,9 +7,9 @@
 import IPublicIdentity from './PublicIdentity'
 
 export interface ICTypeSchema {
-  $id: any
-  $schema: any
-  properties: any
+  $id: string
+  $schema: string
+  properties: object // TO DO: need to refine what properties are
```
Dudleyneedham: I believe this may have been resolved from moving the CTypeUtils into the demo client.

tjwelde:
Please test in demo client

src/blockchain/TxStatus.ts
flechtnergalani:
```suggestion
export type QueryResult = Codec
```
Playing with chain queries it seemed to me that we only ever get Codec types back, even when the query is for something that is not on chain. In fact, it was going through some trouble in [this](https://github.com/KILTprotocol/sdk-js/pull/221) to make sure `null` is returned in these cases. Tests pass when QuerieResult is just `Codec`, including integrations tests.

src/blockchain/TxStatus.ts
```diff
@@ -5,11 +5,17 @@
...
+
 export default class TxStatus {
   public type: string
-  public payload: any
+  public payload: QueryResult
 
-  public constructor(type: string, payload?: any) {
+  public constructor(type: string, payload?: QueryResult) {
```
flechtnergalani: TxStatus.payload is only ever used by Ctype.store - unfortunately, it's used as a plain object, so QueryResult won't do. Would be worthwhile discussing whether payload is needed at all

PR #211, opened 2019-10-30
Refactor: Splitting metadata from the schema

Dudleyneedham:
https://github.com/KILTprotocol/ticket/issues/15

Removed the metadata from the schema
Created a new class in order to create an instance for the metadata object to be created
Added tests suite to check if the metadata matches the metadata Model and included a check to verify if the metadata has the corresponding ctype hash.

Added the changes to my CType Utils PR on the Demo client, due to the changes from it.

Changes to the demo client - https://github.com/KILTprotocol/demo-client/pull/231

## How to test:

Run yarn test
Test Suites: 1 skipped, 18 passed, 18 of 19 total
Tests:       2 skipped, 52 passed, 54 total

## Checklist:


src/ctype/CType.spec.ts
```diff
@@ -73,6 +58,6 @@ describe('CType', () => {
   })
   it('verifies the claim structure', () => {
     expect(claimCtype.verifyClaimStructure(claim)).toBeTruthy()
-    expect(claimCtype.verifyClaimStructure(!claim)).toBeFalsy()
```
Dudleyneedham: I noticed this wasn't doing what I wanted, because of the checks in place regarding the schema interfaces it was hard to find a way around without calling errors in the above code. So, I found a hacky way of fixing it. 

Please, let me know if this is ok.

src/ctype/CType.spec.ts
```diff
@@ -37,13 +29,6 @@ describe('CType', () => {
...
       type: 'object',
     },
-    metadata: {
-      title: { default: 'CType Title' },
-      description: {},
-      properties: {
-        name: { title: { default: 'Name' } },
-      },
-    },
   } as ICType)
```
Dudleyneedham: As above as below, I have take out all the metadata from the test suites in the ctype spec files. I have created a new spec file for the metadata.

src/ctype/CType.ts
```diff
@@ -27,16 +27,13 @@ export default class CType implements ICType {
...
 
   public constructor(ctype: ICType) {
     if (!CTypeUtils.verifySchema(ctype, CTypeWrapperModel)) {
       throw new Error('CType does not correspond to schema')
     }
     this.schema = ctype.schema
-    this.metadata = ctype.metadata
     this.owner = ctype.owner
-
     this.hash = CTypeUtils.getHashForSchema(this.schema)
```
Dudleyneedham: Cleaned it out of the ctype.

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      description: { type: 'string' },
+      properties: {
+        'first-property': { type: 'integer' },
+        'second-property': { type: 'string' },
+      },
+      type: 'object',
+    },
+    ctypeHash: 'string',
+  } as ICtypeMetadata
+
```
Dudleyneedham: This is the general format I was going for with it, but I think it has room for improvement, maybe we don't need the $id or $schema. But, it could be a way to use it again. 

I could remove it altogether. Room for discussion

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  const metadata = new CTypeMetadata(ctype, ctypeMetadata)
+
```
Dudleyneedham: I am not so thrilled about putting the whole ctype within the CTypeMetadata, creating the new instances. 

This was my work-around, maybe there is a better way of doing it. 

Or will we need it to build the ctyeMetadata?
It seems messy and contains too much unnecessary information.

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  it('verifies the metadata of a ctype', async () => {
+    expect(metadata.ctypeHash).not.toHaveLength(0)
```
Dudleyneedham: It seems pointless, but necessary at the same time.

src/ctype/CTypeMetadata.ts
```diff
@@ -0,0 +1,17 @@
...
+  public metadata: ICtypeMetadata['metadata']
+
+  public constructor(ctype: CType, ctypeMetadata: CTypeMetadata) {
+    if (!CTypeUtils.verifySchema(ctypeMetadata, CTypeWrapperMetadata)) {
+      throw new Error('CTypeMetadata does not correspond to schema')
+    }
+    this.metadata = ctypeMetadata.metadata
+    this.ctypeHash = ctype.hash
+  }
+}
```
Dudleyneedham: I am quite happy with this, it is clean and to the point.

Only thing is the CType, as I feel I have too extra information.

src/ctype/CTypeSchema.ts
```diff
@@ -144,5 +158,6 @@ export const CTypeWrapperModel = {
       },
     },
   },
-  required: ['schema'],
+  ctypeHash: { type: 'string', minLength: 1 },
+  required: ['metadata', 'ctypeHash'],
 }
```
Dudleyneedham: Cleaned up inside the CType Schema.

src/types/CType.ts
```diff
@@ -13,19 +13,24 @@ export interface ICTypeSchema {
...
+      type: string
+    }
+    description: {
+      type: string
+    }
+    properties: object
+    type: 'object'
+  }
+  ctypeHash: any
 }
```
Dudleyneedham: This should be string

tjwelde:
Plz try in demo client

Dudleyneedham:
https://github.com/KILTprotocol/demo-client/pull/231

src/types/CTypeMetedata.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @module TypeInterfaces/CType
+ */
+/**
+ * Dummy comment needed for correct doc display, do not remove.
+ */
+export default interface ICTypeMetadata {
+  metadata: IMetadata
```
Dudleyneedham: @maudnals I had to remove the id and schema, because of the demo client. It didn't allow for them.  without changing more of the input data structure in the backend to make it work.

src/types/CTypeMetedata.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @module TypeInterfaces/CType
+ */
+/**
+ * Dummy comment needed for correct doc display, do not remove.
+ */
+export default interface ICTypeMetadata {
+  metadata: {
+    $id: string
```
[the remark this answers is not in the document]
Dudleyneedham: I removed the `$id` and `$schema`, as I needed to manipulate the data to fit the current data structure within the backend on the Demo client.

src/types/CTypeMetedata.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @module TypeInterfaces/CType
+ */
+/**
+ * Dummy comment needed for correct doc display, do not remove.
+ */
+export default interface ICTypeMetadata {
```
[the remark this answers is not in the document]
Dudleyneedham: am I blind or where? :D

src/ctype/CTypeMetadata.ts
Dudleyneedham: I have updated the constructor to take the `hash`

Dudleyneedham:
Merged the develop into branch.

src/types/CTypeMetedata.ts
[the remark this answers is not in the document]
LeonFLK: Metedata
METE EVEN THIS ECRONYM

src/ctype/CType.spec.ts
tjwelde: I think it would be best to modify the claim, so that it actually does not match the ctype

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,46 @@
+import CType from './CType'
+import ICType from '../types/CType'
+import CTypeMetadata from './CTypeMetadata'
+import * as CTypeUtils from './CTypeUtils'
+import { CTypeWrapperMetadata } from './CTypeSchema'
+import ICTypeMetadata from '../types/CTypeMetadata'
+
+jest.mock('../blockchainApiConnection/BlockchainApiConnection')
```
tjwelde: Why do you need the mock? I don't see a call to ctype.store or similar

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      $id: 'http://example.com/ctype-1',
+      $schema: 'http://kilt-protocol.org/draft-01/ctype#',
+      properties: {
+        'first-property': { type: 'integer' },
+        'second-property': { type: 'string' },
+      },
+      type: 'object',
+    },
+  } as ICType)
+
```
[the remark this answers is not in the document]
tjwelde: That's fine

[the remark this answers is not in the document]
tjwelde: But the `as ICtype` should be removed. We don't need to force it to be of type `ICType`, because the class `CType` already is.

src/ctype/CTypeMetadata.spec.ts
tjwelde: Is this comment supposed to be at the Ctype creation?

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  const ctypeHash = ctype.hash
+
+  const ctypeMetadata = {
+    title: { default: 'string' },
+    description: { default: 'string' },
+    properties: {
+      'first-property': { type: 'integer' },
+      'second-property': { type: 'string' },
+    },
+  } as ICTypeMetadata['metadata']
```
tjwelde: It is preferred to define the type like this:
```
const ctypeMetadata: ICTypeMetadata['metadata'] = { ... }
```
instead of with `as`.

The reason is, that with `as`, you are just forcing the type, whereas by defining it like I did here, Typescript actually does a type check

src/ctype/CTypeMetadata.spec.ts
tjwelde: I don't understand. Where are you putting the whole ctype somewhere?

src/ctype/CTypeMetadata.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+        'second-property': { type: 'string' },
+      },
+      type: 'object',
+    },
+  } as ICType)
+
+  const ctypeHash = ctype.hash
+
+  const ctypeMetadata = {
+    title: { default: 'string' },
```
tjwelde: Why are you using `'string'` as the title? I feel. that you are confused about how the metadata looks like and the json-schema.

src/ctype/CTypeMetadata.spec.ts
tjwelde: It is fine

src/ctype/CTypeMetadata.ts
```diff
@@ -0,0 +1,17 @@
+import ICTypeMetadata, { IMetadata } from '../types/CTypeMetadata'
+import CType from './CType'
+import { WrapperMetadata } from './CTypeSchema'
+import * as CTypeUtils from './CTypeUtils'
+
+export default class CTypeMetadata implements ICTypeMetadata {
+  public ctypeHash: ICTypeMetadata['ctypeHash']
+  public metadata: ICTypeMetadata['metadata']
+
+  public constructor(ctype: CType['hash'], ctypeMetadata: IMetadata) {
```
tjwelde: Why are using `IMetadata` here, when you use `ICTypeMetadata['metadata']` at other places?

src/ctype/CTypeSchema.ts
```diff
@@ -133,7 +133,46 @@ export const CTypeWrapperModel = {
...
+          properties: {},
+        },
+      },
+    },
+  },
+  ctypeHash: { type: 'string', minLength: 1 },
+  required: ['metadata', 'ctypeHash'],
+}
+
+export const WrapperMetadata = {
```
tjwelde: Is this used anywhere?

src/ctype/CTypeSchema.ts
```diff
@@ -133,7 +133,46 @@ export const CTypeWrapperModel = {
...
+  },
+  required: ['schema'],
+}
+
+export const CTypeWrapperMetadata = {
+  metadata: {
+    $id: 'http://kilt-protocol.org/draft-01/ctype-metadata',
+    $schema: 'http://json-schema.org/draft-07/schema#',
+    title: {
+      type: ' string',
```
tjwelde: As far as I remember, we always have at least a `default` prop here and are expecting more props for other languages

src/ctype/CTypeSchema.ts
```diff
@@ -133,7 +133,46 @@ export const CTypeWrapperModel = {
       type: 'string',
       minLength: 1,
     },
-    metamodel: {
+  },
+  required: ['schema'],
+}
+
+export const CTypeWrapperMetadata = {
```
tjwelde: Let's come up with a better naming here. It is not wrapping the ctype, it is rather the model or json-schema for the metadata + ctypehash object

src/ctype/CTypeMetadata.spec.ts
Dudleyneedham: Done!

src/ctype/CTypeMetadata.spec.ts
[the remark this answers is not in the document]
Dudleyneedham: Updated!

src/ctype/CTypeMetadata.spec.ts
Dudleyneedham: Done!

src/ctype/CTypeMetadata.spec.ts
Dudleyneedham: I wanted to ask if we should just use the CTypeHash or a whole CType into the metadata?

So, inside the Metadata, we only use the CTypeHash. What do you think?

src/ctype/CTypeMetadata.ts
Dudleyneedham: I guess it seemed like it should be only that. But, I understand it is more consistent.

Dudleyneedham: Also, this seems a little strange that I had `ICTypeMetadata['ctypeHash']` and `CType['hash']` should I just use the `ICTypeMetadata['ctypeHash']`?

src/ctype/CTypeSchema.ts
Dudleyneedham: I think it is used to check the Metadata model.

src/ctype/CTypeSchema.ts
Dudleyneedham: I changed it to the following `CTypeMetadataModel`

src/ctype/CTypeSchema.ts
Dudleyneedham: I changed it to default again.

src/ctype/CTypeMetadata.spec.ts
Dudleyneedham: Understood now.

src/ctype/CTypeMetadata.ts
LeonFLK: It's basically a reference to each other so it makes no difference, right?

tjwelde: right

tjwelde: we should just be consistent

PR #219, opened 2019-12-10
Feat: added signature/verification features to the DID Document methods

maudnals:
## fixes https://github.com/KILTprotocol/ticket/issues/170 

- Added a signature property to the generated Did Document and a verification utility method 
- Created dedicated interfaces
- Made sure to create both static and non-static versions of the new methods

## How to test:
`yarn test`

## Checklist:

### DIFF
--- a/src/did/Did.utils.ts
+++ b/src/did/Did.utils.ts
@@ -98,9 +98,9 @@ export function createDefaultDidDocument(
 
 export function verifyDidDocumentSignature(
   didDocument: IDidDocumentSigned,
-  address: string
+  identifier: string
 ): boolean {
-  if (!didDocument || !didDocument.signature || !address) {
+  if (!didDocument || !didDocument.signature || !identifier) {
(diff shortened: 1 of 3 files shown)


src/did/Did.utils.ts
```diff
@@ -45,3 +57,94 @@ export function getAddressFromIdentifier(
...
+            serviceEndpoint: kiltServiceEndpoint,
+          },
+        ]
+      : [],
+  }
+}
+
+export function verifyDidDocumentSignature(
+  didDocument: IDidDocumentSigned,
+  address: string
```
tjwelde: Can we have the did identifier here, check the did identifier against the one in the did document and use it for checking the signature?

maudnals: I've updated this, it's ready for re-review!

maudnals: sounds good, done 👍

PR #220, opened 2019-12-20
feat: multiple transactions per account

tjwelde:
## fixes KILTprotocol/ticket#820

This PR adds support for doing multiple transactions per account at once, by managing a account specific nonce map and re-signing and re-sending tx that failed with an recoverable error.

Added or Refactored:
 - `IS_RELEVANT_ERROR`: regexp based `ErrorEvaluator` that matches recoverable errors.

 - `parseSubscriptionOptions`: `SubscriptionPromiseOptions` can be completely or partially omitted when calling any of the Blockchain methods.

 - `submitSignedTx`: exported function tries to send tx, catches errors, and rejects either with the original rejection reason, or rejects with ERROR_TRANSACTION_RECOVERABLE if the error can be recovered when re-signing.

 - `submitSignedTx`: Blockchain class method, calls exported function submitSignedTx, catches recoverable errors and re-signs and re-sends the tx up to two times, if identity was supplied.

 - `getNonce`: Uses `api.rpc.system.accountNextIndex` to determine the highest known nonce for the specified account. This nonce might be outdated when using multiple off-chain workers with the same account.

 - `reSignTx`: purges account from map, re-fetches the nonce, generates updated `SignerPayload` and injects it into the existing provided tx.

## How to test:
- Trigger multiple blockchain transactions
- All should go through
- Run CI without --runInBand, with realistic worker count (8 threads:` -w 7`)

## Checklist:

### DIFF
--- a/src/__integrationtests__/Attestation.spec.ts
+++ b/src/__integrationtests__/Attestation.spec.ts
@@ -8,11 +8,10 @@ import { IAttestedClaim, IClaim } from '..'
 import Attestation from '../attestation/Attestation'
 import { revoke } from '../attestation/Attestation.chain'
 import AttestedClaim from '../attestedclaim/AttestedClaim'
-import {
+import Blockchain, {
   IS_IN_BLOCK,
   IS_READY,
   IBlockchainApi,
-  submitSignedTx,
 } from '../blockchain/Blockchain'
 import getCached, { DEFAULT_WS_ADDRESS } from '../blockchainApiConnection'
 import Claim from '../claim/Claim'
@@ -34,8 +33,10 @@ import {
 } from './utils'
 
 let blockchain: IBlockchainApi | undefined
+let alice: Identity
 beforeAll(async () => {
   blockchain = await getCached(DEFAULT_WS_ADDRESS)
+  alice = await Identity.buildFromURI('//Alice')
 })
 
 describe('handling attestations that do not exist', () => {
@@ -45,10 +46,9 @@ describe('handling attestations that do not exist', () => {
 
   it('Attestation.revoke', async () => {
     return expect(
-      Attestation.revoke(
-        '0x012012012',
-        await Identity.buildFromURI('//Alice')
-      ).then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+      Attestation.revoke('0x012012012', alice).then((tx) =>
+        Blockchain.submitSignedTx(alice, tx, { resolveOn: IS_IN_BLOCK })
+      )
     ).rejects.toThrow()
   }, 30_000)
 })
@@ -68,7 +68,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     // console.log(`verify stored: ${await DriversLicense.verifyStored()}`)
     if (!ctypeExists) {
       await DriversLicense.store(attester).then((tx) =>
-        submitSignedTx(tx, IS_READY)
+        Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_READY })
       )
     }
   }, 60_000)
@@ -106,7 +106,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     )
     await attestation
       .store(attester)
-      .then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+      .then((tx) =>
+        Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
+      )
     const cred = await Credential.fromRequestAndAttestation(
       claimer,
       request,
@@ -142,7 +144,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     await expect(
       attestation
         .store(bobbyBroke)
-        .then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+        .then((tx) =>
+          Blockchain.submitSignedTx(bobbyBroke, tx, { resolveOn: IS_IN_BLOCK })
+        )
     ).rejects.toThrow()
     const cred = await Credential.fromRequestAndAttestation(
       bobbyBroke,
@@ -184,7 +188,11 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       attester.getPublicIdentity()
     )
     await expect(
-      attestation.store(attester).then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+      attestation
+        .store(attester)
+        .then((tx) =>
+          Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
+        )
     ).rejects.toThrowError(ERROR_CTYPE_NOT_FOUND)
   }, 60_000)
 
@@ -207,7 +215,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       )
       await attestation
         .store(attester)
-        .then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+        .then((tx) =>
+          Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
+        )
       const cred = await Credential.fromRequestAndAttestation(
         claimer,
         request,
@@ -221,7 +231,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       await expect(
         attClaim.attestation
           .store(attester)
-          .then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+          .then((tx) =>
+            Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
+          )
       ).rejects.toThrowError(ERROR_ALREADY_ATTESTED)
     }, 15_000)
 
@@ -246,7 +258,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     it('should not be possible for the claimer to revoke an attestation', async () => {
       await expect(
         revoke(attClaim.getHash(), claimer).then((tx) =>
-          submitSignedTx(tx, IS_IN_BLOCK)
+          Blockchain.submitSignedTx(claimer, tx, { resolveOn: IS_IN_BLOCK })
         )
       ).rejects.toThrowError('not permitted')
       await expect(attClaim.verify()).resolves.toBeTruthy()
@@ -255,7 +267,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     it('should be possible for the attester to revoke an attestation', async () => {
       await expect(attClaim.verify()).resolves.toBeTruthy()
       await revoke(attClaim.getHash(), attester).then((tx) =>
-        submitSignedTx(tx, IS_IN_BLOCK)
+        Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
       )
       await expect(attClaim.verify()).resolves.toBeFalsy()
     }, 40_000)
@@ -265,7 +277,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     beforeAll(async () => {
       if (!(await CtypeOnChain(IsOfficialLicenseAuthority))) {
         await IsOfficialLicenseAuthority.store(faucet).then((tx) =>
-          submitSignedTx(tx, IS_IN_BLOCK)
+          Blockchain.submitSignedTx(faucet, tx, { resolveOn: IS_IN_BLOCK })
         )
       }
       await expect(
@@ -295,7 +307,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       )
       await licenseAuthorizationGranted
         .store(faucet)
-        .then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+        .then((tx) =>
+          Blockchain.submitSignedTx(faucet, tx, { resolveOn: IS_IN_BLOCK })
+        )
       // make request including legitimation
       const iBelieveICanDrive = Claim.fromCTypeAndClaimContents(
         DriversLicense,
@@ -322,7 +336,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
         attester.getPublicIdentity()
       )
       await LicenseGranted.store(attester).then((tx) =>
-        submitSignedTx(tx, IS_IN_BLOCK)
+        Blockchain.submitSignedTx(attester, tx, { resolveOn: IS_IN_BLOCK })
       )
       const license = await Credential.fromRequestAndAttestation(
         claimer,
--- a/src/__integrationtests__/Attestation.spec.ts
+++ b/src/__integrationtests__/Attestation.spec.ts
@@ -47,7 +47,7 @@ describe('handling attestations that do not exist', () => {
   it('Attestation.revoke', async () => {
     return expect(
       Attestation.revoke('0x012012012', alice).then((tx) =>
-        Blockchain.submitSignedTx(tx, alice, { resolveOn: IS_IN_BLOCK })
+        Blockchain.submitTxWithReSign(tx, alice, { resolveOn: IS_IN_BLOCK })
       )
     ).rejects.toThrow()
   }, 30_000)
@@ -68,7 +68,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     // console.log(`verify stored: ${await DriversLicense.verifyStored()}`)
     if (!ctypeExists) {
       await DriversLicense.store(attester).then((tx) =>
-        Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_READY })
+        Blockchain.submitTxWithReSign(tx, attester, { resolveOn: IS_READY })
       )
     }
   }, 60_000)
@@ -107,7 +107,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     await attestation
       .store(attester)
       .then((tx) =>
-        Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
+        Blockchain.submitTxWithReSign(tx, attester, { resolveOn: IS_IN_BLOCK })
       )
     const cred = await Credential.fromRequestAndAttestation(
       claimer,
@@ -142,11 +142,11 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     )
 
     await expect(
-      attestation
-        .store(bobbyBroke)
-        .then((tx) =>
-          Blockchain.submitSignedTx(tx, bobbyBroke, { resolveOn: IS_IN_BLOCK })
-        )
+      attestation.store(bobbyBroke).then((tx) =>
+        Blockchain.submitTxWithReSign(tx, bobbyBroke, {
+          resolveOn: IS_IN_BLOCK,
+        })
+      )
     ).rejects.toThrow()
     const cred = await Credential.fromRequestAndAttestation(
       bobbyBroke,
@@ -188,11 +188,11 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       attester.getPublicIdentity()
     )
     await expect(
-      attestation
-        .store(attester)
-        .then((tx) =>
-          Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
-        )
+      attestation.store(attester).then((tx) =>
+        Blockchain.submitTxWithReSign(tx, attester, {
+          resolveOn: IS_IN_BLOCK,
+        })
+      )
     ).rejects.toThrowError(ERROR_CTYPE_NOT_FOUND)
   }, 60_000)
 
@@ -213,11 +213,11 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
         request,
         attester.getPublicIdentity()
       )
-      await attestation
-        .store(attester)
-        .then((tx) =>
-          Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
-        )
+      await attestation.store(attester).then((tx) =>
+        Blockchain.submitTxWithReSign(tx, attester, {
+          resolveOn: IS_IN_BLOCK,
+        })
+      )
       const cred = await Credential.fromRequestAndAttestation(
         claimer,
         request,
@@ -229,11 +229,11 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
 
     it('should not be possible to attest the same claim twice', async () => {
       await expect(
-        attClaim.attestation
-          .store(attester)
-          .then((tx) =>
-            Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
-          )
+        attClaim.attestation.store(attester).then((tx) =>
+          Blockchain.submitTxWithReSign(tx, attester, {
+            resolveOn: IS_IN_BLOCK,
+          })
+        )
       ).rejects.toThrowError(ERROR_ALREADY_ATTESTED)
     }, 15_000)
 
@@ -258,7 +258,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     it('should not be possible for the claimer to revoke an attestation', async () => {
       await expect(
         revoke(attClaim.getHash(), claimer).then((tx) =>
-          Blockchain.submitSignedTx(tx, claimer, { resolveOn: IS_IN_BLOCK })
+          Blockchain.submitTxWithReSign(tx, claimer, { resolveOn: IS_IN_BLOCK })
         )
       ).rejects.toThrowError('not permitted')
       await expect(attClaim.verify()).resolves.toBeTruthy()
@@ -267,7 +267,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     it('should be possible for the attester to revoke an attestation', async () => {
       await expect(attClaim.verify()).resolves.toBeTruthy()
       await revoke(attClaim.getHash(), attester).then((tx) =>
-        Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
+        Blockchain.submitTxWithReSign(tx, attester, { resolveOn: IS_IN_BLOCK })
       )
       await expect(attClaim.verify()).resolves.toBeFalsy()
     }, 40_000)
@@ -277,7 +277,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
     beforeAll(async () => {
       if (!(await CtypeOnChain(IsOfficialLicenseAuthority))) {
         await IsOfficialLicenseAuthority.store(faucet).then((tx) =>
-          Blockchain.submitSignedTx(tx, faucet, { resolveOn: IS_IN_BLOCK })
+          Blockchain.submitTxWithReSign(tx, faucet, { resolveOn: IS_IN_BLOCK })
         )
       }
       await expect(
@@ -308,7 +308,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       await licenseAuthorizationGranted
         .store(faucet)
         .then((tx) =>
-          Blockchain.submitSignedTx(tx, faucet, { resolveOn: IS_IN_BLOCK })
+          Blockchain.submitTxWithReSign(tx, faucet, { resolveOn: IS_IN_BLOCK })
         )
       // make request including legitimation
       const iBelieveICanDrive = Claim.fromCTypeAndClaimContents(
@@ -336,7 +336,7 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
         attester.getPublicIdentity()
       )
       await LicenseGranted.store(attester).then((tx) =>
-        Blockchain.submitSignedTx(tx, attester, { resolveOn: IS_IN_BLOCK })
+        Blockchain.submitTxWithReSign(tx, attester, { resolveOn: IS_IN_BLOCK })
       )
       const license = await Credential.fromRequestAndAttestation(
         claimer,
--- a/src/__integrationtests__/Attestation.spec.ts
+++ b/src/__integrationtests__/Attestation.spec.ts
@@ -47,7 +47,7 @@ describe('handling attestations that do not exist', () => {
   it('Attestation.revoke', async () => {
     return expect(
       Attestation.revoke('0x012012012', alice).then((tx) =>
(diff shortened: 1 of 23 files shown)


flechtnergalani:
@tjwelde We're not quite there yet; all transactions are currently rejected with something like 
```    2019-12-20 14:02:31             DRR: Error: submitAndWatchExtrinsic (extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: 1
        at errorHandler (/Users/flechtner/Documents/KILT/integrationstests/sdk-js/node_modules/@polkadot/rpc-core/index.js:276:26)
        at /Users/flechtner/Documents/KILT/integrationstests/sdk-js/node_modules/@polkadot/rpc-core/index.js:249:9
        at processTicksAndRejections (internal/process/task_queues.js:93:5)```

**Comment by LeonFLK** [comment]
I have implemented mutex, that manages concurrent access to the Map `accountNonces` ( previously `currentNonce`) which maps the account's `nonce` to the account `address`.

In the current implementation the `nonceQueue` is account independent, meaning different accounts would wait for their turn to access the `accountNonces`.

We could further implement a Map, which maps `nonceQueue` to the account address, so that retrieving nonces for different accounts could be performed concurrently.

**Comment by tjwelde** [comment]
Can we write tests for this?

File: src/blockchain/Blockchain.ts [hunk 1]
```diff
@@ -82,48 +96,115 @@ export default class Blockchain implements IBlockchainApi {
...
+                log.debug(
+                  `Transaction detail: ${JSON.stringify(result, null, 2)}`
+                )
+                const extrinsicError: ExtrinsicError =
+                  this.errorHandler.getExtrinsicError(result) || ERROR_UNKNOWN
+
+                log.warn(`Extrinsic error ocurred: ${extrinsicError}`)
+                reject(extrinsicError)
+              }
+              if (status.type === 'Finalized') {
```
Dudleyneedham: I have added constants for status.

src/blockchain/Blockchain.ts
```diff
@@ -82,48 +96,115 @@ export default class Blockchain implements IBlockchainApi {
...
-          reject(err)
+    try {
+      const nonce: Index = await this.getNonce(identity.address)
+      if (nonce.toNumber() === -1) {
+        throw new Error('Error while retrieving Nonce')
+      } else {
+        const signed = identity.signSubmittableExtrinsic(tx, nonce.toHex())
+        log.info(`Submitting ${tx.method} with Nonce ${nonce}`)
+
+        return new Promise<TxStatus>((resolve, reject) => {
```
Dudleyneedham: you can shorten `(resolve, reject)` to `(res, rej)`

src/blockchain/Blockchain.ts
```diff
@@ -82,48 +96,115 @@ export default class Blockchain implements IBlockchainApi {
...
+                  `Transaction detail: ${JSON.stringify(result, null, 2)}`
+                )
+                const extrinsicError: ExtrinsicError =
+                  this.errorHandler.getExtrinsicError(result) || ERROR_UNKNOWN
+
+                log.warn(`Extrinsic error ocurred: ${extrinsicError}`)
+                reject(extrinsicError)
+              }
+              if (status.type === 'Finalized') {
+                resolve(new TxStatus(status.type))
```
Dudleyneedham: IF you change the `resolve` to `res`, you will need to change here.

src/blockchain/Blockchain.ts
```diff
@@ -82,48 +96,115 @@ export default class Blockchain implements IBlockchainApi {
...
+              }
+            })
+            .catch((err: Error) => {
+              // just reject with the original tx error from the chain
+              reject(err)
+            })
         })
+      }
+    } catch (err) {
+      return new Promise<TxStatus>((resolve, reject) => reject(err))
```
Dudleyneedham: you can shorten `(resolve, reject)` to `(res, rej)`

src/blockchain/Blockchain.ts
```diff
@@ -82,48 +96,115 @@ export default class Blockchain implements IBlockchainApi {
...
+  public async getNonce(accountAddress: string): Promise<Index> {
+    const unlock = await this.lock(accountAddress)
+    const nonce = await unlock()
+    return nonce
+  }
+
+  private lock(
+    accountAddress: Identity['address']
+  ): Promise<() => Promise<Index>> {
+    const lock = new Promise<() => Promise<Index>>(resolve => {
```
Dudleyneedham: Same as above

src/blockchain/Blockchain.ts
LeonFLK: I think we want our code to be somewhat verbose, and shortening an argument name has IMO only diminishing returns, shorter name but less "readable" code.

src/blockchain/Blockchain.ts
LeonFLK: yes thanks for the heads up, resolved the conflicts

src/blockchain/Blockchain.ts
Dudleyneedham: Sounds good.

tjwelde:
The accountNonce for an address might need to be reset and retrieved from the chain, if the client is kept open, while doing transactions on another client with the same address.
I am thinking about setting it to `null` at the end of the queue.

LeonFLK:
The suggested approach still has some limitations to it, as it only mitigates the risk of accessing an outdated nonce in `accountNonce` after we finished iterating through the queue, but not while we iterate.

`api.query.system.accounNonce`'s return Nonce is only updated after the transaction has been finalized, so even after we finish the queue, the `api.query.system.accounNonce`'s is not up to date within the same block.

Lets discuss in person, as I want to raise a few more issues with that approach.

tjwelde:
As we talked in person, we want to wait for the transactions to be finalized, before we reset the queue and we say, that we do not support multi-clients on the same identity

LeonFLK:
Resets account specific nonce queue, when either the Tx result has isFinalized or isError set, or receives an Chain extrinsicError, and there are no more pending Tx waiting for a Nonce.

LeonFLK:
> Is this mocked separately because you need to mock the api, but not the Blockchain itself?

I need to mock the api only since i need to be able to retrieve a nonce from the api, without actually establishing connection
I resurrected the Blockchain.ts, only to unit test the nonce retrieval logic, I will remove all the already excluded tests.

> This could be
> 
>     if ((this.nonceQueue.get(accountAddress) || []).length > 0) {
> to avoid the linter warning

Good Suggestion

> Why do you return some magic value instead of rejecting the promise and handling the rejection?

It is indeed a magic value, that traverses the universe and spaghetti code like a Planeswalker to reach it's error handling destination, instead of having to reject every single promise, that this magic number visits.

Jokes aside: I refactored it to be less magical, I promise.

LeonFLK:
I changed the mocked and actual `getCached()`,  `buildConnection()` and cached "instance" (ironically not an instance of Blockchain) from `Promise<IBlockchainApi>` to `Promise<Blockchain>` to better represent the code behavior.

weichweich:
Did you guys checkout the [batch transaction](https://polkadot.js.org/api/cookbook/tx.html#how-can-i-batch-transactions)?

> Polkadot/Substrate provides a utility.batch method that can be used to send a number of transactions at once. These are then executed from a single sender (single nonce specified) in sequence. This is very useful in a number of cases, for instance if you wish to create a payout for a validator for multiple eras, you can use this method. Likewise, you can send a number of transfers at once. Or even batch different types of transactions.

weichweich:
If KILTprotocol/mashnet-node/pull/81 gets merged, we have the nextNonce RPC which takes the tx pool into account.

But this is only part of the solution. The next nonce only works if we wait until a transaction was put into the pool if we loop over transactions without waiting for them to be in the pool the next nonce will be wrong.

I guess it would be great to have KILTprotocol/ticket/issues/513. Then we could wait on different stages of the transactions. Maybe something like that:

```ts
const txs = []
for (let i = 0; i < 10; i++) {
  const txHandle = await kilt.attester.attest(...).execute()
  await txHandle.ready
  txs.push(txHandle.finalize)
}
await Promise.all(txs)
```

We should also consider to add the batch pallet to our runtime.

src/blockchain/Blockchain.ts
```diff
@@ -84,46 +99,160 @@ export default class Blockchain implements IBlockchainApi {
...
+   * @returns Promise of the [[Index]] representation of the Transaction nonce of the identity.
+   *
+   */
   public async getNonce(accountAddress: string): Promise<Index> {
-    return this.api.rpc.system.accountNextIndex(accountAddress)
+    // Initiate nonceRetrieval
+    const unlock = await this.lock(accountAddress)
+    // Await execution of the in handleQueue defined resolve function
+    const nonce = await unlock()
+    return nonce
```
weichweich, line 155: I think we could simply check if `accountAddress` is in `this.accountNonces` if not, query next nonce, otherwise increase `this.accountNonces[accountAddress]` by one and return it.

This would not be a [critical section](https://en.wikipedia.org/wiki/Critical_section) because [node.js, v8 and stuff is single threaded](https://stackoverflow.com/questions/5153492/models-of-concurrency-in-nodejs).

Or did I overlook something?

src/blockchain/Blockchain.ts
```diff
@@ -84,46 +99,160 @@ export default class Blockchain implements IBlockchainApi {
...
-    })
+              const extrinsicError: ExtrinsicError =
+                this.errorHandler.getExtrinsicError(result) || ERROR_UNKNOWN
+
+              log.warn(`Extrinsic error ocurred: ${extrinsicError}`)
+              this.resetAccountQueue(identity.address)
+              reject(extrinsicError)
+            }
+            if (result.isFinalized) {
+              this.resetAccountQueue(identity.address)
```
rflechtner, line 124: the fact that you rely on the stuff that happens in the promise we return from submitTx could be a problem later on - we plan to expose the submittable extrinsic, so people are free to call `tx.submit()` and this would never be triggered

rflechtner: Can we solve this any other way?

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +48,103 @@ export interface IBlockchainApi {
...
-  getNonce(accountAddress: string): Promise<Codec>
+  getNonce(accountAddress: string, reset?: boolean): Promise<Index>
+  reSignTx(
+    identity: Identity,
+    tx: SubmittableExtrinsic
+  ): Promise<SubmittableExtrinsic>
 }
 
+export const IS_RELEVANT_ERROR: ErrorEvaluator = (err: Error) => {
+  return /Priority|Transaction Already|outdated/g.test(err.message)
```
LeonFLK, line 69: I hate this.

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +211,158 @@ export default class Blockchain implements IBlockchainApi {
...
+   *
+   * @param accountAddress The address of the identity that we retrieve the nonce for.
+   * @param reset Specify whether the entry for the account is outdated and has to be reset.
+   * @returns [[Index]] representation of the Tx nonce for the identity.
+   *
+   */
+  public async getNonce(accountAddress: string, reset = false): Promise<Index> {
+    if (reset) {
+      this.accountNonces.delete(accountAddress)
+    }
```
weichweich, line 299: maybe we could move that to a separate method? The call side could just call `resetNonce(account)`. That way the method only does one thing , which reduces complexity.

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +211,158 @@ export default class Blockchain implements IBlockchainApi {
...
+      return chainNonce
+    }
+    if (secondQuery) {
+      this.accountNonces.set(
+        accountAddress,
+        this.api.registry.createType('Index', secondQuery.addn(1))
+      )
+      return secondQuery
+    }
+    throw Error(`Nonce Retrieval Failed for : ${accountAddress}`)
```
weichweich, line 331: So in pseudo code this does:

```
nonce := get nonce from store
if nonce was not in store:
    chain_nonce := get nonce from chain
    stored_nonce := get nonce from store
    nonce = max(chain_nonce, stored_nonce)
store nonce + 1
return nonce
```

Might be easier to read since it only returns once and it's written in pseudo code. 😉

weichweich (commented):
looks nice! I think this is better than the first approach. But i would try to simplify it even further. It might be easier to understand if there is only one return in the end. what are your thoughts?  😊

src/blockchain/Blockchain.ts
LeonFLK: Agreed.

src/blockchain/Blockchain.ts
LeonFLK: Yes, love it. I ditched the whole Index notation, we work with BN only now.

src/blockchain/Blockchain.ts
LeonFLK: As we currently only reset inside reSignTx, we reset it there.

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +200,135 @@ export default class Blockchain implements IBlockchainApi {
...
+          log.error(
+            `On-chain nonce retrieval failed for account ${accountAddress}\nwith reason: ${reason}`
+          )
+          throw Error(`Chain failed to retrieve nonce for : ${accountAddress}`)
+        })
+      const secondQuery = this.accountNonces.get(accountAddress)
+      nonce = BN.max(chainNonce, secondQuery || new BN(0))
+    }
+    this.accountNonces.set(accountAddress, nonce.addn(1))
+    return nonce
```
weichweich, line 296: That's a beauty! ;)

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +200,135 @@ export default class Blockchain implements IBlockchainApi {
...
+    opts?: SubscriptionPromiseOptions
+  ): Promise<SubmittableResult> {
+    const options = parseSubscriptionOptions(opts)
+    return submitSignedTx(tx, options)
+      .catch(async (reason: Error) => {
+        if (reason.message === 'Recoverable') {
+          return submitSignedTx(await this.reSignTx(identity, tx), options)
+        }
+        throw reason
+      })
```
weichweich, line 247: Wow that's a really easy retry! Love it! But maybe refactor the async closure into it's own variable and have

```ts
const retry = async (...) => {...}
return submitSignedTx(tx, options).catch(retry).catch(retry)
```

weichweich (approved):
LGTM! Just one minor thing, but that would not block a merge.

This probably still needs to be updated and reviewed by timo...

src/blockchain/Blockchain.ts
LeonFLK: Good suggestion, might even make it possible in the future to have the user specify how often they want to retry.

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +211,130 @@ export default class Blockchain implements IBlockchainApi {
...
+          log.error(
+            `On-chain nonce retrieval failed for account ${accountAddress}\nwith reason: ${reason}`
+          )
+          throw Error(`Chain failed to retrieve nonce for : ${accountAddress}`)
+        })
+      const secondQuery = this.accountNonces.get(accountAddress)
+      nonce = BN.max(chainNonce, secondQuery || new BN(0))
+    }
+    this.accountNonces.set(accountAddress, nonce.addn(1))
+    return nonce
```
weichweich, line 302: Now that i revisited that code, I notice that it's not clear why we do the things we do. I would maybe add some comments

```suggestion
  public async getNonce(accountAddress: string): Promise<BN> {
    let nonce = this.accountNonces.get(accountAddress)
    if (!nonce) {
      // the account nonce is unknown, we will query it from chain
      const chainNonce = await this.api.rpc.system
        .accountNextIndex(accountAddress)
        .catch((reason) => {
          log.error(
            `On-chain nonce retrieval failed for account ${accountAddress}\nwith reason: ${reason}`
          )
          throw Error(`Chain failed to retrieve nonce for : ${accountAddress}`)
        })
      // ensure that the nonce we queried is still up to date and no newer nonce was queried during the await above
      const secondQuery = this.accountNonces.get(accountAddress)
      nonce = BN.max(chainNonce, secondQuery || new BN(0))
    }
    this.accountNonces.set(accountAddress, nonce.addn(1))
    return nonce
```

weichweich (approved):
LGTM

a minor non blocking thing: maybe we should add a few comments to the nonce query part...

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +52,108 @@ export interface IBlockchainApi {
     identity: Identity,
     tx: SubmittableExtrinsic
   ): Promise<SubmittableExtrinsic>
+  submitSignedTx(
+    identity: Identity,
+    tx: SubmittableExtrinsic,
+    opts?: SubscriptionPromiseOptions
+  ): Promise<SubmittableResult>
```
rflechtner, line 59: again, we have `submitTx` for transactions that need to be (re)signed - is there no way we keep the distinction? I thought it was a nice feature that you didn't have to worry about selecting the correct Identity here

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +52,108 @@ export interface IBlockchainApi {
...
+    ErrorHandler.extrinsicFailed(result) &&
+    (ErrorHandler.getExtrinsicError(result) || UNKNOWN_EXTRINSIC_ERROR)
+  )
+}
 
-export const IS_ERROR: ResultEvaluator = (result) =>
-  result.isError && ERROR_UNKNOWN()
-export const EXTRINSIC_FAILED: ResultEvaluator = (result) =>
-  ErrorHandler.extrinsicFailed(result) &&
-  (ErrorHandler.getExtrinsicError(result) || UNKNOWN_EXTRINSIC_ERROR)
```
rflechtner, line 68: oh, did this not work right? weeeeird

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +52,108 @@ export interface IBlockchainApi {
...
+ */
+export function parseSubscriptionOptions(
+  opts?: Partial<SubscriptionPromiseOptions>
+): SubscriptionPromiseOptions {
+  const {
+    resolveOn = IS_FINALIZED,
+    rejectOn = (result: SubmittableResult) =>
+      IS_ERROR(result) || EXTRINSIC_FAILED(result) || IS_USURPED(result),
+    timeout,
+  } = { ...opts }
```
rflechtner, line 116: OMGG YOU CAN SET DEFAULTS LIKE THIS ????? this changes eeeeeverything!

rflechtner: but do we need this as an extra function? Whenever I use an options map on a function, I just apply defaults in the first line(s). It's a one-liner and used once - could be integrated.

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +211,130 @@ export default class Blockchain implements IBlockchainApi {
...
+        nonce,
+        genesisHash: this.api.genesisHash,
+        blockHash: this.api.genesisHash,
+        runtimeVersion: this.api.runtimeVersion,
+        version: this.api.extrinsicVersion,
+      })
+      .toPayload()
+    tx.addSignature(
+      identity.address,
+      this.api
```
rflechtner, line 331: daaaaamn I thought you could just do `tx.sign()` again 😣

rflechtner (approved):
LGTM! Just a few comments and questions.
I was hoping you could still send a signed tx off without having to know which signer was used. Is there no way that could work (I know this would mean the tx could bounce back, but that would be the tradeoff).

src/blockchain/Blockchain.ts
LeonFLK: Only addSignature doesn't check whether the extrinsic was already signed, and just injects the data, regardless of existence of the signature.

src/blockchain/Blockchain.ts
LeonFLK: Yeah, it's pretty sexy, innit?
Yes, this could be a one-liner, though this would also mean that passing the whole object referencing these created vars would be ugly IMO.
instead of `parseSubscriptionOptions(opts)` it would be `{resolveOn, rejectOn, timeout }`.
What are your thoughts?

src/blockchain/Blockchain.ts
LeonFLK: I didn't remove this, it's working just the way you wrote it.

src/util/SubscriptionPromise.ts
```diff
@@ -11,27 +11,36 @@ export interface Evaluator<SubscriptionType> {
...
+}
+
 /**
  * Helps building a promise associated to a subscription callback through which updates can be pushed to the promise.
  * This promise is resolved with the value of latest update when a resolution criterion is met.
  * It is rejected with a custom error/reason if a rejection criterion is met or on timeout (optional). Rejection takes precedent.
  *
- * @param resolveOn Resolution criterion. A function that evaluates an incoming update from the subscription.
+ * @param terminationOptions Object provides:
+ * resolveOn: Resolution criterion. A function that evaluates an incoming update from the subscription.
```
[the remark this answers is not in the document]
LeonFLK: That's correct, updating it!

src/identity/Identity.ts
```diff
@@ -425,21 +427,21 @@ export default class Identity {
...
-  public signSubmittableExtrinsic(
+  public async signSubmittableExtrinsic(
     submittableExtrinsic: SubmittableExtrinsic,
-    nonceAsHex: string
-  ): SubmittableExtrinsic {
-    return submittableExtrinsic.sign(this.signKeyringPair, {
-      nonce: nonceAsHex,
+    nonce: number | Index | BN
+  ): Promise<SubmittableExtrinsic> {
+    return submittableExtrinsic.signAsync(this.signKeyringPair, {
```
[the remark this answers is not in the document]
LeonFLK: yes signAsync basically calls accountNextIndex for you and applies the nonce, if you don't supply one yourself

src/blockchain/Blockchain.ts
LeonFLK: Yeah we could have something, that prevents retrying if no identity is given. This wouldn't actually be very complicated as we'd just have to set the retry to something that just doesn't retry.

src/__integrationtests__/Attestation.spec.ts
```diff
@@ -45,10 +46,9 @@ describe('handling attestations that do not exist', () => {
 
   it('Attestation.revoke', async () => {
     return expect(
-      Attestation.revoke(
-        '0x012012012',
-        await Identity.buildFromURI('//Alice')
-      ).then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
+      Attestation.revoke('0x012012012', alice).then((tx) =>
+        Blockchain.submitSignedTx(alice, tx, { resolveOn: IS_IN_BLOCK })
```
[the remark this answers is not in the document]
LeonFLK: The tx are getting signed already, e.g. makeTransfer() calls signTx.

src/blockchain/Blockchain.ts
rflechtner: maybe we could have a `retryWith` parameter on the opts object?

src/blockchain/Blockchain.ts
rflechtner: I see what you mean... mhh an alternative would be
```js
const defaults = {
     resolveOn: IS_FINALIZED,
     rejectOn: (result: SubmittableResult) =>
       IS_ERROR(result) || EXTRINSIC_FAILED(result) || IS_USURPED(result),
   }
const options = Object.assign(defaults, opts) // unfortunately {...defaults, ...opts} results in weird behaviour
```

src/blockchain/Blockchain.ts
rflechtner: that sucks man, so much code just to get around this

src/blockchain/Blockchain.ts
LeonFLK: I set the identity parameter to optional, and if it's not given, then we don't retry

rflechtner: nice 👍

weichweich:
I hope we merge that before the PR celebrates its first anniversary 😂

rflechtner:
> I hope we merge that before the PR celebrates its first anniversary 😂

We got a full month to go! Eaaasy

But yeah, I don't want to be in the way of merging it, I'm happy with what I see. Go ahead and merge as far as I'm concerned

weichweich:
> > I hope we merge that before the PR celebrates its first anniversary 😂
> 
> We got a full month to go! Eaaasy
> 
> But yeah, I don't want to be in the way of merging it, I'm happy with what I see. Go ahead and merge as far as I'm concerned

Ah sorry that was not meant as an answer to your post. Just noticed the age of the PR. 😁

src/blockchain/Blockchain.ts
LeonFLK: I ment:

`const defaults = {
     resolveOn: IS_FINALIZED,
     rejectOn: (result: SubmittableResult) =>
       IS_ERROR(result) || EXTRINSIC_FAILED(result) || IS_USURPED(result),
   } = {...opts}

const options = {resolveOn, rejectOn, timeout } `

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +52,105 @@ export interface IBlockchainApi {
...
-  getNonce(accountAddress: string): Promise<Codec>
+  getNonce(accountAddress: string): Promise<BN>
+  reSignTx(
+    identity: Identity,
+    tx: SubmittableExtrinsic
+  ): Promise<SubmittableExtrinsic>
 }
+const TxOutdated = '1010: Invalid Transaction: Transaction is outdated'
+const TxPriority = '1014: Priority is too low:'
+const TxAlreadyImported = 'Transaction Already'
```
tjwelde, line 73: we need good integration tests for these

src/blockchain/Blockchain.ts
tjwelde:
```javascript
const defaults = {
  resolve: IS_FINALIZED,
  rejectOn: (result: SubmittableResult) =>
       IS_ERROR(result) || EXTRINSIC_FAILED(result) || IS_USURPED(result),
}

return {
  ...defaults,
  ...opts
}
```

src/blockchain/Blockchain.ts
```diff
@@ -47,56 +52,105 @@ export interface IBlockchainApi {
...
 ): Promise<SubmittableResult> {
-  log.info(`Submitting ${tx.method}`)
-  const { promise, subscription } = makeSubscriptionPromise(
-    resolveOn,
-    rejectOn,
-    timeout
-  )
-  const unsubscribe = await tx.send(subscription)
-  return promise.finally(() => unsubscribe())
+  const { promise, subscription } = makeSubscriptionPromise(opts)
```
tjwelde, line 133: let's keep the log

src/blockchain/Blockchain.ts
```diff
@@ -141,28 +208,135 @@ export default class Blockchain implements IBlockchainApi {
...
+   *
+   * Transaction fees will apply whenever a transaction fee makes it into a block, even if extrinsics fail to execute correctly!
+   *
+   * @param tx The [[SubmittableExtrinsic]] to be submitted. Most transactions need to be signed, this must be done beforehand.
+   * @param identity Optional [[Identity]] to potentially re-sign the Tx with.
+   * @param opts Optional partial criteria for resolving/rejecting the promise.
+   * @returns A promise which can be used to track transaction status.
+   * If resolved, this promise returns the eventually resolved [[SubmittableResult]].
+   */
+  public async submitSignedTx(
```
tjwelde, line 244: Same name as the function export is confusing. What does this method do differently?

tjwelde: `submitTxWithResign`?

src/identity/AttesterIdentity.ts
```diff
@@ -313,7 +313,7 @@ export default class AttesterIdentity extends Identity {
     const bc = await getCached()
     const tx = bc.portablegabi.buildUpdateAccumulatorTX(acc)
     this.accumulator = acc
-    await bc.portablegabi.signAndSend(tx, this.signKeyringPair)
+    await bc.submitTx(this, tx)
```
tjwelde, line 316: Why is this change necessary?

src/blockchain/Blockchain.spec.ts
```diff
@@ -36,3 +53,294 @@ describe('queries', () => {
...
+  describe('reSignTx', () => {
+    const submittable: SubmittableExtrinsic = ({
+      signature: {
+        toHuman: jest.fn(),
+      },
+      addSignature: jest.fn(),
+      nonce: { toHuman: jest.fn() },
+      method: { data: 'unchanged', toHex: jest.fn() },
+    } as unknown) as SubmittableExtrinsic
+    it('fetches updated Nonce and applies updated signature to Extrinsic', async () => {
```
tjwelde, line 190: Please add newlines in front of `it` functions

tjwelde (commented):
Looks very good now.
Just some nitpicks

src/blockchain/Blockchain.ts
tjwelde: Please make a ticket and provide information and ideas on how to trigger the errors

LeonFLK: Created Ticket.

src/blockchain/Blockchain.ts
LeonFLK: I like `submitTxWithResign`, but we have many similarly named functions that makes this class kinda convoluted.
After refactoring of the tx signing and submission procedure, this should be more on point.

src/identity/AttesterIdentity.ts
LeonFLK: Signing and sending Tx with gabi chain can result in nonce collision as well.

src/blockchain/Blockchain.ts
tjwelde: https://github.com/KILTprotocol/ticket/issues/845

src/identity/AttesterIdentity.ts
LeonFLK: Different instances of our Chain will use the `SubmittableExtrinsic` specific methods to sign and or send the tx, and is therefore completely independent from the chain instance, after the tx's creation/instantiation.

src/blockchain/Blockchain.ts
LeonFLK: Was placed in wrong place, now back to original

tjwelde:
LGTM

PR #221, opened 2020-01-09
HOLD: further empty strings decoded to string repr

flechtnergalani:
**This PR is made redundant by changes to the mashnet-node in https://github.com/KILTprotocol/mashnet-node/pull/78
Additionally, the code here may even be incompatible with the changes on the node.**

fixes: [222](https://github.com/KILTprotocol/ticket/issues/222)
the issue fixed on #218 may also exist on other decode methods.
Applied the fix to all that I found, queries returned with no result are now returned as null.
This is only an interim solution, however; a final solution should build on a suitable decoder (see [issue 226](https://github.com/KILTprotocol/ticket/issues/226). 

## How to test:
1. git cherry-pick e6932b7c419671bbbfcdd74719d9af90b78eac55
2. rename `noResultQueries.intspec.ts` > `testme.spec.ts`
3. start mashnet node with `docker run -d --rm -p 9944:9944 kiltprotocol/mashnet-node ./target/release/mashnet-node --dev --ws-port 9944 --ws-external`
4. `yarn test testme`

## Checklist:


tjwelde:
Looks good so far, but lets fix all the places, where it could be an issue

src/did/Did.utils.ts
```diff
@@ -27,7 +27,10 @@ export function decodeDid(
   identifier: string,
   encoded: QueryResult
 ): IDid | null {
-  const json = encoded && encoded.encodedLength ? encoded.toJSON() : null
+  const json =
+    encoded && encoded.encodedLength && !encoded.isEmpty
```
flechtnergalani:
```suggestion
    encoded && !encoded.isEmpty
```
does `encoded.encodedLength` serve any other function besides what `!encoded.isEmpty` does? It seems to me that the former was also meant to assure empty QueryResults are not decoded and null is returned instead, which only the latter reliably does. In that case, we could remove the former?
@tjwelde

src/blockchain/Blockchain.ts
```diff
@@ -44,7 +44,11 @@ export interface IBlockchainApi {
 export default class Blockchain implements IBlockchainApi {
   public static asArray(queryResult: QueryResult): any[] {
     const json =
-      queryResult && queryResult.encodedLength ? queryResult.toJSON() : null
+      queryResult &&
+      !queryResult.isEmpty &&
+      (queryResult instanceof Array ? !queryResult.every(e => e.isEmpty) : true)
```
flechtnergalani: will not work as expected if any element is a bool (isEmpty => always false). Not sure how to proceed

flechtnergalani:
This has been addressed and fixed by https://github.com/KILTprotocol/mashnet-node/pull/78

PR #231, opened 2020-01-31
fix: ctype.verifyStored() always returned false when owner not set

flechtnergalani:
Fixes https://github.com/KILTprotocol/ticket/issues/284

Given a Ctype instance like 
```
  const myCtype = CType.fromCType({
    schema: {
      $id: 'http://example.com/ctype-1',
      $schema: 'http://kilt-protocol.org/draft-01/ctype#',
      properties: {
        name: { type: 'string' },
      },
      type: 'object',
    } as ICType['schema'],
  } as ICType)
```

`myCtype.verifyStored()` would always return `false`, because the owner property is unset. This is the case even after storing the Ctype instance on chain. `true` is returned only if the owner property is set to the same address that is returned by the chain query (`getOwner()`).
With the proposed change,  `myCtype.verifyStored()` returns true if 
a) the owner property is unset and the chain query does not return null (=the ctype exists)
b) the owner property contains an address which is identical to the address returned by the chain query.

## How to test:
```
await expect(myCtype.verifyStored()).resolves.toBeFalsy()
await myCtype.store(identity.adress)
await expect(myCtype.verifyStored()).resolves.toBeTruthy()
myCtype.owner = identity.adress
await expect(myCtype.verifyStored()).resolves.toBeTruthy()
```
## Checklist:


flechtnergalani:
Part of this PR should be to uncomment 
https://github.com/KILTprotocol/sdk-js/blob/5a3e145b3fad850b9970eefd781d149ed3103608/src/__integrationtests__/Ctypes.spec.ts#L29
and the following line

PR #232, opened 2020-02-03
fix: blockchain connection disconnect

maudnals:
## fixes https://github.com/KILTprotocol/ticket/issues/169 
Added an explicit `disconnect` function in Kilt default exports, as a counterpart of `connect`. Disconnect also cleans up the cache.

## How to test:

- Check out this branch
- In the SDK repo: `yarn`, `yarn build`, `yalc publish`
- Create a new test folder and `yarn init --y` and `yalc add @kiltprotocol/sdk-js` and `yarn add tslib`.
- Test out connecting / disconnecting by using the new `disconnect` function. No errors should be thrown when sticking to a regular connect/disconnect flow. Example:

```javascript
  const connection1 = await Kilt.BlockchainApiConnection.getCached(
    "wss://full-nodes.kilt.io:9944"
  );
  await Kilt.default.disconnect("wss://full-nodes.kilt.io:9944");
  const connection2 = await Kilt.BlockchainApiConnection.getCached(
    "wss://full-nodes.kilt.io:9944"
  );
  // No error should be thrown
  connection2.api.disconnect("wss://full-nodes.kilt.io:9944");
```

## Checklist:


src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -46,4 +46,9 @@ export async function getCached(
   return instance
 }
 
+export function clearCache(): null {
+  instance = null
+  return instance
```
wischli: Question: Is it necessary for this function to return something or should it rather be `void`? I am asking as I am working on the chain connection to Portablegabi and my code is very similar except for `clearCache` having no return value in my case.

maudnals: Yup makes sense, I'm removing the return value since indeed `clear`-like methods e.g. in web APIs don't return.

PR #234, opened 2020-02-12
docs: documentation of the quote

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/292

Adds the documentation of the quote.

## How to test:

Test Suites: 1 skipped, 19 passed, 19 of 20 total
Tests:       2 skipped, 63 passed, 65 total

## Checklist:


src/quote/Quote.ts
```diff
@@ -19,6 +19,16 @@ import Identity from '../identity/Identity'
 import { IQuote, IQuoteAgreement, IQuoteAttesterSigned } from '../types/Quote'
 import { hashObjectAsStr, verify } from '../crypto/Crypto'
 
+/**
+ * Validates the quote against the meta schema and quote data against the provided schema.
+ *
+ * @param schema A [[Quote]] schema object.
+ * @param validate [[Quote]] data to be validated against the provided schema.
+ * @param messages The validating function provides an array of errors, in the case of validation failures.
```
maudnals: Suggestion:
Rephrase line 27 to make the sentence/structure consistent with other parameter doc, e.g.:
"messages: Errors provided by..."

Also:
Out-of-scope for this PR, but validate could/should be replaced by a noun such as dataToValidate

src/quote/Quote.ts
```diff
@@ -19,6 +19,16 @@ import Identity from '../identity/Identity'
...
+ *
+ * @returns Boolean to provide a check against the data.
```
maudnals: Suggestion:
Use the same phrasing as for boolean return values elsewhere in the SDK, e.g. "Whether the quote schema is valid".

src/quote/Quote.ts
```diff
@@ -40,6 +50,14 @@ export function validateQuoteSchema(
   return !!result
 }
 
+/**
+ * Deserializes a [[Quote]] object into a readable format.
```
maudnals: Change request:
If this function deserializes as mentioned in line 54, then the input is not already deserialized. 
Also, it's mentioned that `deserializedQuote` is a `serialized` object, isn't that contradictory?

src/quote/Quote.ts
```diff
@@ -77,6 +104,15 @@ export function createAttesterSignature(
   }
 }
 
+/**
+ * Creates a [[Quote]] object with a given [[Identity]].
```
maudnals: Suggestion:
Maybe you can precise that this creates a quote object **signed** by the given identity.

src/quote/Quote.ts
```diff
@@ -87,6 +123,16 @@ export function fromQuoteDataAndIdentity(
   return createAttesterSignature(quoteInput, identity)
 }
 
+/**
+ * Creates a [[Quote]] signed by the Attester and the Claimer.
+ *
+ * @param claimerIdentity [[Identity]] of the Claimer in order to sign.
+ * @param attesterSignedQuote An Attester signed [[Quote]] object.
```
maudnals: Question:
Is the wording of "An attester signed" clear? Or should we say "... signed by an Attester"?

src/quote/Quote.ts
Dudleyneedham: Done

src/quote/Quote.ts
Dudleyneedham: Agreed

src/quote/Quote.ts
Dudleyneedham: Agreed

src/quote/Quote.ts
Dudleyneedham: Agreed

src/quote/Quote.ts
Dudleyneedham: I have updated it. 

But, I am struggling with what I should write.

src/quote/Quote.ts
maudnals: Not really ☺️ Can you please check again?

src/quote/Quote.ts
```diff
@@ -19,6 +19,16 @@ import Identity from '../identity/Identity'
 import { IQuote, IQuoteAgreement, IQuoteAttesterSigned } from '../types/Quote'
 import { hashObjectAsStr, verify } from '../crypto/Crypto'
 
+/**
+ * Validates the quote against the meta schema and quote data against the provided schema.
+ *
+ * @param schema A [[Quote]] schema object.
+ * @param validate [[Quote]] data to be validated against the provided schema.
+ * @param messages Errors provide an array of errors, in the case of validation failures.
```
maudnals: Not really changed as discussed ☺️ 
Can you please check again?
https://github.com/KILTprotocol/sdk-js/pull/234#discussion_r378294636

src/quote/Quote.ts
maudnals: Not really ☺️ 
Can you please check again?

src/quote/Quote.ts
maudnals: Maybe some confusion on serialization/deserialiation?

Serialization is this process:
JS object ---> JSON object (string) // via `JSON.stringify`

Deserialization is this:
JSON object (string) --> JS object // via `JSON.parse`

What your function does is just transformation; it's designed to be called with an argument that has already been deserialized. You know this because you know how the function will be used; but the function itself doesn't need to mention that. Except in the code example maybe, where this use case can be illustrated.

--> Check out the code example and the doc of `fromAttestedClaim`, it will help.

src/quote/Quote.ts
Dudleyneedham: Maybe, I am a little confused.  Because it is what the `AJV` states - https://github.com/epoberezkin/ajv#validation-errors

I have updated and tried again.

src/quote/Quote.ts
Dudleyneedham: What do you think of this iteration?

src/quote/Quote.ts
```diff
@@ -40,6 +50,14 @@ export function validateQuoteSchema(
   return !!result
 }
 
+/**
+ * Builds a [[Quote]] object, from a simple object with the same properties.
+ *
+ * @param deserializedQuote The bae object which is used to create the attester signed [[Quote]] object.
+ *
```
wischli: I would add to the description that this function does not only "clone" the quote object but also some checking. Also, the documentation is missing `@throws` for both errors.

src/quote/Quote.ts
```diff
@@ -77,6 +104,15 @@ export function createAttesterSignature(
   }
 }
 
+/**
+ * Creates a [[Quote]] object signed by the given [[Identity]].
+ *
+ * @param quoteInput A [[Quote]] object.
+ * @param identity [[Identity]] to add the public address of the Attester.
```
wischli: Does it really add the public address? Isn't the identity solely used for signing?

src/quote/Quote.ts
```diff
@@ -87,6 +123,16 @@ export function fromQuoteDataAndIdentity(
   return createAttesterSignature(quoteInput, identity)
 }
 
+/**
+ * Creates a [[Quote]] signed by the Attester and the Claimer.
```
wischli: Missing `@throws`

src/quote/Quote.ts
Dudleyneedham: Yes, we currently don't use `@throws`, which could be useful. 

Something to discuss, at the moment I will not add.

src/quote/Quote.ts
Dudleyneedham: You are right, we are only signing with it. 

I have updated it accordingly.

src/quote/Quote.ts
Dudleyneedham: Mentioned above.

PR #235, opened 2020-02-13
test: updating unit tests to use Option return values

flechtnergalani:
fixes https://github.com/KILTprotocol/ticket/issues/304
Option return values are introduced in KILTprotocol/mashnet-node#78
This change set is compatible with the current SDK. This PR just updates the mocked return values to reflect that objects of type `Option` are returned in most `api.query.` calls to ensure they accurately reflect the behaviour of the actual mashnet node.

## How to test:
`yarn test`

## Checklist:


flechtnergalani:
I enabled integration tests which are expected to fail until https://github.com/KILTprotocol/ticket/issues/304 has been merged and included in a release

PR #238, opened 2020-02-14
refactor: introduces index files and use them

tjwelde:
## fixes KILTProtocol/ticket#311
This PR adds index files to all src subfolders and uses them in the root index file.

## How to test:
- Use the SDK in demo client / services / demo wallet

## Checklist:


src/index.ts
```diff
@@ -2,45 +2,38 @@
...
+import AttestedClaim, { AttestedClaimUtils } from './attestedclaim'
+import {
+  DelegationBaseNode,
+  DelegationNode,
+  DelegationRootNode,
+} from './delegation'
+import Did, { IDid } from './did'
+import Quote, { QuoteUtils, QuoteSchema } from './quote'
+import Message from './messaging'
 import * as Verifier from './actor/Verifier'
```
wischli, line 28: Q: Do we want index files for actors?

src/index.ts
```diff
@@ -2,45 +2,38 @@
...
 import * as Claimer from './actor/Claimer'
 import * as Attester from './actor/Attester'
 import Credential from './credential/Credential'
```
wischli, line 31: Missing index file.

PR #241, opened 2020-02-20
fix: example

weichweich:
## fixes example
The example was broken and should be fixed now.

## How to test:

```
ts-node docs/example.ts
```

## Checklist:

### DIFF
--- a/docs/example.ts
+++ b/docs/example.ts
@@ -6,6 +6,7 @@ import Kilt, {
   Message,
   // IRequestForAttestation,
   ISubmitAttestationForClaim,
+  // ICTypeMetadata,
 } from '../src'
 
 // How to generate an Identity
@@ -35,30 +36,32 @@ const ctypeSchema: ICType['schema'] = {
 // Generate the Hash for it
 const ctypeHash = CTypeUtils.getHashForSchema(ctypeSchema)
 // Build metadata for schema
-const ctypeMetadata: ICType['metadata'] = {
-  title: {
-    default: 'DriversLicense',
-  },
-  description: {
-    default: '',
-  },
-  properties: {
-    name: {
-      title: {
-        default: 'name',
-      },
-    },
-    age: {
-      title: {
-        default: 'age',
-      },
-    },
-  },
-}
-// Put everyhting together
+// const ctypeMetadata: ICTypeMetadata = {
+//   metadata: {
+//     title: {
+//       default: 'DriversLicense',
+//     },
+//     description: {
+//       default: '',
+//     },
+//     properties: {
+//       name: {
+//         title: {
+//           default: 'name',
+//         },
+//       },
+//       age: {
+//         title: {
+//           default: 'age',
+//         },
+//       },
+//     },
+//   },
+//   ctypeHash,
+// }
+// Put everything together
 const rawCtype: ICType = {
   schema: ctypeSchema,
-  metadata: ctypeMetadata,
   hash: ctypeHash,
   owner: claimer.address,
 }
@@ -72,7 +75,7 @@ const ctype = new Kilt.CType(rawCtype)
 
 // Store ctype on blockchain
 // ! This costs tokens !
-// Also note, that the completly same ctype can only be stored once on the blockchain.
+// Also note, that the completely same ctype can only be stored once on the blockchain.
 // ctype.store(claimer)
 
 // How to build a Claim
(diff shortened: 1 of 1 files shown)


weichweich:
@Dudleyneedham [what should I do with the](https://github.com/KILTprotocol/sdk-js/pull/241/files#diff-74824597fce70cad35c6ad88a75cf318R39-R61) `ICTypeMetadata`? Delete it or keep it?

docs/example.ts
```diff
@@ -6,6 +6,7 @@ import Kilt, {
   Message,
   // IRequestForAttestation,
   ISubmitAttestationForClaim,
+  // ICTypeMetadata,
```
Dudleyneedham: Honestly, having the Metadata here could be confusing. As it serves no purpose unless storing and retrieving the CType in different languages. 

At the moment, I could remove it. It has its main application within the Demo-client and we have it here to allow people to use it.

docs/example.ts
```diff
@@ -72,7 +75,7 @@ const ctype = new Kilt.CType(rawCtype)
 
 // Store ctype on blockchain
 // ! This costs tokens !
-// Also note, that the completly same ctype can only be stored once on the blockchain.
+// Also note, that the completely same ctype can only be stored once on the blockchain.
```
Dudleyneedham: I don't like completely here anyway. I would remove it, or change it. 

Suggestion: 

Also note, that the absolutely identical ctype can only be stored once on the blockchain.

docs/example.ts
```diff
@@ -91,26 +98,27 @@ const mnemonicForAttester = Kilt.Identity.generateMnemonic()
...
+Message.ensureHashAndSignature(encrypted, claimer.address)
 // When the Attester receives the message, she can decrypt it
-const decrypted = Message.createFromEncryptedMessage(
-  encrypted,
-  claimer,
-  attester
-)
+const decrypted = Message.createFromEncryptedMessage(encrypted, attester)
 
 // And make sure, that the sender is the owner of the identity
```
Dudleyneedham: I am not sure about this line here. What does it relate to? Why do we need to know the owner of the identity is also the sender?

weichweich: we need to make sure that the sender is the owner of the attested claim. If Alice shares her attested claim with Mallory, Mallory could send it to a verifier. If the verifier doesn't verify that sender == owner he will not notice that the attested claim belongs to Alice and not to Mallory.

docs/example.ts
weichweich: removed it

docs/example.ts
```diff
@@ -82,7 +63,11 @@ const rawClaim = {
   age: 29,
 }
 
-const claim = new Kilt.Claim(ctype, rawClaim, claimer)
+const claim = new Kilt.Claim({
+  cTypeHash: ctypeHash,
+  contents: rawClaim,
+  owner: claimer.address,
+})
```
flechtnergalani: the new constructor since 0.17.0 is 
```suggestion
const claim = Kilt.Claim.fromCTypeAndClaimContents(
  ctype,
  rawClaim,
  claimer.address,
)
```
additionally, there is a new constructor for the Ctype:
`Kilt.CType.fromCType(rawCtype)`

docs/example.ts
```diff
@@ -6,6 +7,7 @@ import Kilt, {
   Message,
   // IRequestForAttestation,
   ISubmitAttestationForClaim,
+  // ICTypeMetadata,
```
flechtnergalani: this can probably go for good?

docs/example.ts
```diff
@@ -91,28 +76,30 @@ const mnemonicForAttester = Kilt.Identity.generateMnemonic()
 const attester = Kilt.Identity.buildFromMnemonic(mnemonicForAttester)
 
 // And we need to build a request for an attestation
-const requestForAttestation = new Kilt.RequestForAttestation(claim, [], claimer)
+const requestForAttestation = Kilt.RequestForAttestation.fromClaimAndIdentity(
+  claim,
+  claimer
+)
 
 // Excourse to the messaging system
```
flechtnergalani: oh and excourse : not a word

weichweich: I'm not sure what they wanted to say. I changed it to `Excursion to the messaging system`.

tjwelde:
Looks good

PR #244, opened 2020-02-28
feat: change `any` to something more meaningful

### LINKED ISSUE #164: SDK - Change `any` types to something meaningful

flechtnergalani:
@Dudleyneedham 
Squashed commits from #196 on top of develop and resolved merge conflicts, then added two more commits replacing some more `any`s. 

fixes #164 

- defined any typed where they could be narrowed down to more specific types
- refactorings to avoid using any return or argument types
- removed await on non-async function calls

## How to test:

- [x] run unit tests
- [x] run integration tests
- [x] test with demo client 

## Checklist:


src/did/Did.ts
```diff
@@ -67,7 +67,7 @@ export interface IDidDocumentPublicKey {
 export interface IDidDocumentPpties {
   authentication: object
   publicKey: IDidDocumentPublicKey[]
-  service: any
+  service: any // can this possibly be narrowed down to something other than any?
```
flechtnergalani: @maudnals I'll need your input here - `any` ideas? Please?

src/ctype/CType.spec.ts
```diff
@@ -63,7 +63,7 @@ describe('CType', () => {
       owner: identityAlice.address,
     }
 
-    const resultTxStatus = new TxStatus(FINALIZED, Crypto.hashStr('987654'))
```
Dudleyneedham: This is out of the scope of the ticket.

src/delegation/Delegation.chain.ts
```diff
@@ -37,7 +37,10 @@ export async function getChildIds(
   id: IDelegationBaseNode['id']
 ): Promise<string[]> {
   const blockchain = await getCached()
-  return Blockchain.asArray(await blockchain.api.query.delegation.children(id))
```
Dudleyneedham: This is out of the scope of the ticket.

src/delegation/DelegationNode.chain.ts
```diff
@@ -23,7 +23,7 @@ export async function store(
   const includeParentId: boolean = delegation.parentId
     ? delegation.parentId !== delegation.rootId
     : false
-  const tx: SubmittableExtrinsic = await blockchain.api.tx.delegation.addDelegation(
```
Dudleyneedham: See other comments about the ticket.

src/messaging/Message.ts
```diff
@@ -281,7 +281,7 @@ export interface IRequestAcceptDelegation extends IMessageBodyBase {
       isPCR: boolean
     }
     metaData?: {
-      [key: string]: any
+      [key: string]: string | object // can this possibly be narrowed down to something other than any?
```
Dudleyneedham: You will need to remove the comments.

src/quote/Quote.ts
```diff
@@ -29,8 +29,10 @@ export function validateQuoteSchema(
   )
   if (!result && ajv.errors) {
     if (messages) {
-      ajv.errors.forEach((error: any) => {
-        messages.push(error.message)
+      ajv.errors.forEach((error: Ajv.ErrorObject) => {
```
Dudleyneedham: Nice

src/types/CType.ts
```diff
@@ -5,9 +5,9 @@
 import IPublicIdentity from './PublicIdentity'
 
 export interface ICTypeSchema {
-  $id: any
-  $schema: any
-  properties: any
+  $id: string
+  $schema: string
+  properties: object // TO DO: need to refine what properties are
```
Dudleyneedham: This needs to be changed.

src/ctype/CTypeUtils.ts
```diff
@@ -8,34 +8,41 @@ import Ajv from 'ajv'
...
-export function verifyClaimStructure(claim: any, schema: any): boolean {
+export function verifyClaimStructure(
+  claimContents: IClaim['contents'], // Checking the SDK and demo-client I feel this should be named claimContents
+  schema: ICType['schema']
+): boolean {
   if (!verifySchema(schema, CTypeModel)) {
     throw new Error('CType does not correspond to schema')
   }
-  return verifySchema(claim, schema)
+  return verifySchema(claimContents, schema)
```
Dudleyneedham: I agree this needs to be changed to this. But, it is out of the scope of the ticket.

src/ctype/CType.spec.ts
flechtnergalani: A change will be necessary though if we’re changing the type of TxStatus.payload to object. We could pass an object instead, but this is dead code anyway, it is never used afterwards.

src/delegation/Delegation.chain.ts
flechtnergalani: I wouldn't think it is - it's a necessary change to get rid of the return type `any[]` of `Blockchain.asArray()` 
What we probably could do instead is convert each element to string in order to match the `string[]` return type of `getChildIds()`

src/delegation/DelegationNode.chain.ts
flechtnergalani: I agree this could be merged separately

weichweich:
https://github.com/KILTprotocol/sdk-js/blob/2dde442d0e87d948f08638ed83120d5390364a28/src/types/RequestForAttestation.ts#L20

We could add a type for that. 🙂 
```ts
export type NonceHashTree = NonceHash | { [key: string]: NonceHashTree }
```

flechtnergalani:
> https://github.com/KILTprotocol/sdk-js/blob/2dde442d0e87d948f08638ed83120d5390364a28/src/types/RequestForAttestation.ts#L20
> 
> We could add a type for that. 🙂
> 
> ```ts
> export type NonceHashTree = NonceHash | { [key: string]: NonceHashTree }
> ```

I like that ! But https://github.com/KILTprotocol/sdk-js/blob/2dde442d0e87d948f08638ed83120d5390364a28/src/requestforattestation/RequestForAttestation.ts#L308 would technically require it to be 
```ts
export type NonceHashTree = { [key: string]: NonceHash }
```
Would also work!

weichweich:
> ```ts
> export type NonceHashTree = { [key: string]: NonceHash }
> ```
> Would also work!

But that's not a tree. If we want to be able to hide arbitrarily nested fields we need a tree (see KILTprotocol/ticket/issues/386 for details).

weichweich:
https://github.com/KILTprotocol/sdk-js/blob/2dde442d0e87d948f08638ed83120d5390364a28/src/types/Claim.ts#L10

could also have this type, which is pretty close to `any`. 🤣 

```ts
export type ClaimContent =
  | string
  | number
  | null
  | boolean
  | { [key: string]: ClaimContent }
  | ClaimContent[]
```

flechtnergalani:
> ```ts
> ```ts
> export type NonceHashTree = { [key: string]: NonceHash }
> ```
> 
> 
> Would also work!
> 
> But that's not a tree. If we want to be able to hide arbitrarily nested fields we need a tree (see KILTprotocol/ticket/issues/386 for details).
> ```

It is not, but I'd say we stick to this until the code supports an actual tree - that should be happening as part of the issue you opened, not as part of this PR

flechtnergalani:
> https://github.com/KILTprotocol/sdk-js/blob/2dde442d0e87d948f08638ed83120d5390364a28/src/types/Claim.ts#L10
> 
> could also have this type, which is pretty close to `any`. 🤣
> 
> ```ts
> export type ClaimContent =
>   | string
>   | number
>   | null
>   | boolean
>   | { [key: string]: ClaimContent }
>   | ClaimContent[]
> ```

An issue similar to the ClaimHashTree: It can't be a scalar or Array because the property definitions in the Ctype require a Struct / Object, so maybe
```ts
[...]
contents: { [key: string]: ClaimProperty }
[...]
export type ClaimProperty =
  | string
  | number
  | null
  | boolean
  | { [key: string]: ClaimProperty }
  | ClaimProperty[]
```
But it seems like the `CTypeModel` currently only allows four types: 
```ts
          properties: {
            type: {
              type: 'string',
              enum: ['string', 'integer', 'number', 'boolean'],
            },
```
in which case 
```ts
contents: { [key: string]: string | number | boolean }
```
covers it all - it doesn't even allow `null` when validating a claim against a ctype.

src/did/Did.ts
```diff
@@ -67,7 +68,14 @@ export interface IDidDocumentPublicKey {
...
-  service: any
+  service: IDidService[]
+}
+
+export interface IDidService {
+  type: string
+  serviceEndpoint: string | AnyJsonObject
+  id?: string
+  [key: string]: AnyJson
 }
```
flechtnergalani: we implemented this on the basis of the [W3C draft](https://www.w3.org/TR/did-core/#service-endpoints), hope that makes sense to you @maudnals

src/ctype/CTypeUtils.ts
flechtnergalani: given that the name of the parameter should not have any consequences for the function call, can we not just leave it like this?

PR #248, opened 2020-03-11
WIP: replace transaction status constants with substrate SDK status checks

flechtnergalani:
## fixes https://github.com/KILTprotocol/ticket/issues/384
Proposal to replace transaction status constants with properties from Substrate's `ExtrinsicStatus`. See ticket.
Draft contains two possible ways of implementing this; either by implementing `ExtrinsicStatus` or by extending `SubmittableResult`.
The latter seems less hacky, but instances of the former are far easier to create, thus making for better usability. 
I'd appreciate your feedback here.

## How to test:
We'll worry about this later

## Checklist:


flechtnergalani:
I'm closing this because we have merged #250 which removes status constants along with the TxStatus class

PR #249, opened 2020-03-19
fix: ctype compress accepts nulled owner

LeonFLK:
## fixes [KILTProtocol/ticket#312](https://github.com/KILTprotocol/ticket/issues/397)

Currently compress throws an Error when given a ctype with owner set to null, even though `CType['owner']` is of union type `IPublicIdentity['address'] | null`

It should instead check, if the owner is a valid address or null

checking with checkAddress from '@polkadot/util-crypto' and prefix 42

## Checklist:

### DIFF
--- a/src/ctype/CType.utils.ts
+++ b/src/ctype/CType.utils.ts
@@ -107,7 +107,11 @@ export function decompressSchema(
  */
 
 export function compress(cType: ICType): CompressedCType {
-  if (!cType.hash || !cType.owner || !cType.schema) {
+  if (
+    !cType.hash || typeof cType.owner === 'string'
+      ? true
+      : cType.owner === null || !cType.schema
+  ) {
     throw new Error(
       `Property Not Provided while building cType: ${JSON.stringify(
         cType,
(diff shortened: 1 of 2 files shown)


src/ctype/CType.utils.ts
```diff
@@ -107,7 +108,13 @@ export function decompressSchema(
  */
 
 export function compress(cType: ICType): CompressedCType {
-  if (!cType.hash || !cType.owner || !cType.schema) {
+  if (
+    !cType.hash ||
+    (typeof cType.owner === 'string'
+      ? !checkAddress(cType.owner, 42)[0]
```
Dudleyneedham: Is `42` the length of the address or the maximum number it can be?

LeonFLK: as per https://github.com/paritytech/substrate/wiki/External-Address-Format-(SS58): 00101010b (42) Generic Substrate wildcard (SS58 checksum preimage).

This this is basically the address prefix, which is ofc required parameter.

PR #254, opened 2020-04-09
feat: dockerfile for running integration tests

flechtnergalani:
## relates to KILTProtocol/ticket#415
This allows easy testing of the sdk with a mashnet-node image pulled from dockerhub. 
`docker-compose run sdk && docker-compose down` will run tests and then reset chain state (`docker-compose run sdk` is an alias for `docker-compose run sdk yarn test:integration`).  `docker-compose run sdk yarn test:integration:run Attestation` will run specific tests (here: Attestations).
In order to test agains the current state of the develop branch, you can build a mashnet-node image locally by telling docker-compose to use the override file `dc.build.node.yml` like so: 
`docker-compose -f docker-compose.yml -f dc.build.node.yml run sdk`. Alternatively you could rename `dc.build.node.yml` to `docker-compose.override.yml`, in which case docker-compose loads it by default

## How to test:
Go crazy with integration testing. Be aware that the current code in the develop branch may be incompatible with the latest mashnet-node release available on dockerhub.

## Checklist:


src/__integrationtests__/utils.ts
```diff
@@ -5,6 +5,9 @@ import Identity from '../identity/Identity'
 import CType from '../ctype/CType'
 import ICType from '../types/CType'
 import { getOwner } from '../ctype/CType.chain'
+import { DEFAULT_WS_ADDRESS } from '../blockchainApiConnection'
+
+export const WS_HOST = process.env.WS_HOST || DEFAULT_WS_ADDRESS
```
flechtnergalani, line 10: A nice alternative would be to make the global default configurable via an environment variable by changing line 16 on BlockchainApiConnection.ts to something like
```typescript
export const DEFAULT_WS_ADDRESS = process.env.DEFAULT_WS_ADDRESS || 'ws://127.0.0.1:9944'
```

flechtnergalani:
It would also be possible to run tests against the current develop branch of the mashnet-node, but that would require an extra build of the mashnet-node, which can take more than 30 minutes.

src/__integrationtests__/utils.ts
flechtnergalani: @tjwelde Is this a security relevant change? Would it be potentially dangerous to have the default set via environment variable, because it opens up new ways to have the sdk connect to a different endpoint?
It would allow me to revert many of the proposed changes on the tests, so I'd prefer this way of handling endpoint configuration.

flechtnergalani: implemented this now, lemme know if you think that's a problem

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:alpine
+
+WORKDIR /app
+
+COPY package.json yarn.lock ./
+
+RUN yarn install
+
+COPY ./src ./src
+COPY . ./
```
tjwelde, line 10: Doesn't this also copy `src`?

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '3'
+
+services:
+  dev-node:
+    image: kiltprotocol/mashnet-node
```
tjwelde, line 5: Can you add a commented out alternative, to use the develop branch on github?

flechtnergalani: sure! I'd set it up so the image is built locally, so we don't shut people out who can't login to out AWS?

Dockerfile
flechtnergalani: you're right, doesn't really help caching here

docker-compose.yml
flechtnergalani: I set this up in an override file which you can tell docker-compose to include with docker-compose -f dc.build.node.yml. For example docker-compose -f dc.build.node.yml up --build runs tests with a force rebuild of the mashnet-node from the current develop branch.
The built image is tagged local/mashnet-node, which we can hopefully re-use in other docker-compose files

tjwelde: Looks good, can you please document it somewhere?

flechtnergalani: done!

src/__integrationtests__/utils.ts
tjwelde: Looks good

PR #255, opened 2020-04-17
feat: better blockchain api mocking

flechtnergalani:
## NO-TICKET
This is a suggestion for better mocking of the blockchain api connection. Instead of mocking the entire `Blockchain` module, we would be able to mock only api calls, thus allowing us to use the original functions on the module. We would also only mock `getCached` on the `BlockchainApiConnection` module to inject our mocked API and retain originals of the remaining functions. 
Currently only the spec file I included is written to use this correctly, others will fail. Let me know what you think, if you like what you see I could proceed to refactor our existing tests.

## How to test:
run the dedicated spec file with `yarn jest Mocktest` and play around with the tests included in `Mocktest.spec.ts` 

## Checklist:

### DIFF
--- a/src/blockchain/Mocktest.spec.ts
+++ b/src/blockchain/Mocktest.spec.ts
@@ -1,33 +1,67 @@
 import BN from 'bn.js'
+import { Tuple, Option, U8a, Text } from '@polkadot/types'
 import { Identity } from '..'
 import { makeTransfer } from '../balance/Balance.chain'
 import getCached from '../blockchainApiConnection/BlockchainApiConnection'
+import { queryByAddress } from '../did/Did.chain'
 
 jest.mock('../blockchainApiConnection/BlockchainApiConnection')
 
-it('succeeds', async () => {
-  require('../blockchainApiConnection/BlockchainApiConnection').__setDefaultResult(
-    true
-  )
-  const alice = Identity.buildFromURI('//Alice')
-  const amount = new BN(1000)
-  const blockchain = await getCached()
-  const transfer = blockchain.api.tx.balances.transfer(alice.address, amount)
-  const result = await blockchain.submitTx(alice, transfer)
-  expect(result).toMatchObject({ isFinalized: true })
+describe('tx mocks', () => {
+  it('tx succeeds', async () => {
+    require('../blockchainApiConnection/BlockchainApiConnection').__setDefaultResult(
+      true
+    )
+    const alice = Identity.buildFromURI('//Alice')
+    const amount = new BN(1000)
+    const blockchain = await getCached()
+    const transfer = blockchain.api.tx.balances.transfer(alice.address, amount)
+    const result = await blockchain.submitTx(alice, transfer)
+    expect(result).toMatchObject({ isFinalized: true })
+  })
+
+  it('tx succeeds then fails', async () => {
+    require('../blockchainApiConnection/BlockchainApiConnection').__queueResults(
+      [true, false]
+    )
+    const alice = Identity.buildFromURI('//Alice')
+    const amount = new BN(1000)
+    await expect(
+      makeTransfer(alice, alice.address, amount)
+    ).resolves.toMatchObject({
+      isFinalized: true,
+    })
+    await expect(makeTransfer(alice, alice.address, amount)).rejects.toThrow(
+      'Transaction failed'
+    )
+  })
 })
 
-it('succeeds then fails', async () => {
-  require('../blockchainApiConnection/BlockchainApiConnection').__queueResults([
-    true,
-    false,
-  ])
-  const alice = Identity.buildFromURI('//Alice')
-  const amount = new BN(1000)
-  expect(makeTransfer(alice, alice.address, amount)).resolves.toMatchObject({
-    isFinalized: true,
+describe('mock query result', () => {
+  beforeAll(async () => {
+    require('../blockchainApiConnection/BlockchainApiConnection').__mocked_api.query.did.dIDs.mockReturnValue(
+      new Option(
+        Tuple,
+        new Tuple(
+          // (publicBoxKey, publicSigningKey, documentStore?)
+          [Text, Text, U8a],
+          ['0x987', '0x123', '0x687474703a2f2f6d794449442e6b696c742e696f']
+        )
+      )
+    )
+    // would also work:
+    // ```ts
+    // const bc: any = await getCached()
+    // bc.api.query.did.dIDs.mockReturnValue(...)
+    // ```
   })
-  expect(makeTransfer(alice, alice.address, amount)).rejects.toThrow(
-    'Transaction failed'
-  )
+
+  it('works', async () => {
+    await expect(queryByAddress('0xwertzui')).resolves.toMatchObject({
+      documentStore: 'http://myDID.kilt.io',
+      identifier: 'did:kilt:0xwertzui',
+      publicBoxKey: '0x123',
+      publicSigningKey: '0x987',
+    })
+  })
 })
--- a/src/blockchain/Mocktest.spec.ts
+++ b/src/blockchain/Mocktest.spec.ts
@@ -1,0 +1,33 @@
+import BN from 'bn.js'
+import { Identity } from '..'
+import { makeTransfer } from '../balance/Balance.chain'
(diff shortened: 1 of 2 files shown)


src/blockchain/Mocktest.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  })
+})
+
+describe('mock query result', () => {
+  beforeAll(async () => {
+    require('../blockchainApiConnection/BlockchainApiConnection').__mocked_api.query.did.dIDs.mockReturnValue(
+      new Option(
+        Tuple,
+        new Tuple(
+          // (publicBoxKey, publicSigningKey, documentStore?)
```
wischli, line 46: Order is publicSigningKey, publicBoxKey, documentStore

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,141 @@
...
+
+  public send(callable: Function) {
+    if (callable) {
+      callable(this.result)
+    }
+  }
+}
+
+function __getMockSubmittableExtrinsic(): SubmittableExtrinsic {
+  const result: SubmittableResult = TxresultsQueue.shift() || defaultTxResult
```
wischli, line 51: Would replace `TxresultsQueue` with `TxResultsQueue`

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,141 @@
...
+    },
+    delegation: {
+      root: jest.fn((rootId: string) => new Option(Tuple)),
+      delegations: jest.fn((delegationId: string) => new Option(Tuple)),
+      children: jest.fn((id: string) => new Vec(Text, [])),
+    },
+    did: {
+      dIDs: jest.fn(id => new Option(Tuple)),
+    },
+  },
```
wischli, line 136: Needed to add of `runtimeMetadata` to get rid of an unhandled promise rejection
```javascript
  runtimeMetadata: {
    asV4: {
      modules: [],
    },
  },
```

src/blockchain/Mocktest.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    )
+    const alice = Identity.buildFromURI('//Alice')
+    const amount = new BN(1000)
+    const blockchain = await getCached()
+    const transfer = blockchain.api.tx.balances.transfer(alice.address, amount)
+    const result = await blockchain.submitTx(alice, transfer)
+    expect(result).toMatchObject({ isFinalized: true })
+  })
+
+  it('tx succeeds then fails', async () => {
```
wischli, line 23: Would also add a single test for failed transaction, just to keep it simple.

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,141 @@
...
+    ctype: {
+      cTYPEs: jest.fn(hash => true),
+    },
+    delegation: {
+      root: jest.fn((rootId: string) => new Option(Tuple)),
+      delegations: jest.fn((delegationId: string) => new Option(Tuple)),
+      children: jest.fn((id: string) => new Vec(Text, [])),
+    },
+    did: {
+      dIDs: jest.fn(id => new Option(Tuple)),
```
wischli, line 134: I will submit a proposal for a default query here. There, one can add explanatory comments and reference to this in custom mocks.

flechtnergalani:
@wischli Thank you for your thorough review! Just so that we're on the same page: I added the `Mocktest.spec.ts` just for illustrative purposes and to make it easier to play around with the mock. Probably should have stated this clearly. Would you you advocate keeping this file in the repo as you made suggestions on it? Or is that just a hint to how I should refactor the unit tests that broke with this mock?

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
flechtnergalani: absolutely

src/blockchain/Mocktest.spec.ts
flechtnergalani: Oh, thanks, good catch! I copied that from somewhere, I'll try to keep this in mind for later.

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
flechtnergalani: Nice, good addition

wischli:
>  Would you you advocate keeping this file in the repo as you made suggestions on it? Or is that just a hint to how I should refactor the unit tests that broke with this mock?

Do as you wish. My suggestions can be treated as refactor hints 😇

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
flechtnergalani: My approach was to return null result queries by default (what would be returned by a query for something that is not there), because it seems odd to return a specific result as a default value. Would prefer to stick to that approach, but you're right that documentation and example cases must be given. I could comment this out and leave it as an example return value, and also add similar examples to the other queries?

wischli: Makes sense and I like your proposal.

flechtnergalani: As for documentation on how to set return values, I could also add a comment at the top of the mock or file that illustrates this once.

wischli: Sounds good if it's not too obvious as in this case!

flechtnergalani:
@wischli  I added documentation, is this better understandable now?

wischli:
> @wischli I added documentation, is this better understandable now?

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,264 @@
...
+          null,                                                // parent-id?
+          '5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu',  // Account
+          0,                                                   // permissions
+          false,                                               // revoked flag
+        ]
+      )
+    ) */
+
+      // default return value decodes to [], represents: no children found
+      children: jest.fn((id: string) => new Vec(H256, [])),
```
flechtnergalani, line 230:
```suggestion
      children: jest.fn((id: string) => new Vec(H256)),
```

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,264 @@
...
+import { SubmittableExtrinsic } from '@polkadot/api/promise/types'
+import { ExtrinsicStatus } from '@polkadot/types/interfaces'
+import AccountId from '@polkadot/types/primitive/Generic/AccountId'
+
+const BlockchainApiConnection = jest.requireActual('../BlockchainApiConnection')
 
-export async function getCached() {
-  return Promise.resolve(Blockchain)
+async function getCached(
+  _: string = BlockchainApiConnection.DEFAULT_WS_ADDRESS
```
wischli, line 59: The input can be removed as it is not used at all.

src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -5,12 +5,264 @@
...
+
+class MockSubmittableExtrinsic {
+  result: SubmittableResult
+  method: string = 'mock tx'
+
+  constructor(result: SubmittableResult) {
+    this.result = result
+  }
+
+  public sign(_a: any, _b: any, _c: any) {
```
wischli, line 80: The input variables could be removed as they are not used at all.

flechtnergalani: These are mock implementations, I thought they have to have a signature identical to the original? Won't it just break otherwise?

flechtnergalani: you're right, that works. Have updated!

PR #260, opened 2020-04-29
refactor: use generics on chain queries 

flechtnergalani:
## fixes KILTProtocol/ticket#305
This adds generics to polkadot api calls for chain queries, which communicates the expected return type to our typescript code. It also removes the custom `QueryResult` type as a return type for chain queries which included `null` and `undefined`, which was simply not correct - queries will always return `Codec` types.

This also implements type checks at runtime for codec types provided to decoder methods, making sure that they are only used in appropriate contexts.

## How to test:
Run integration tests with mashnet-node v 0.22 and a develop version

## Checklist:


src/blockchain/Blockchain.ts
```diff
@@ -43,12 +43,11 @@ export interface IBlockchainApi {
...
-    if (json instanceof Array) {
-      return json
-    }
+  public static asArray(queryResult: Codec): AnyJson[] {
+    if (queryResult instanceof Array) return decodeArray(queryResult)
+    // new but in the spirit of the method's name: Struct to Array (drops keys)
+    if (queryResult instanceof Map)
+      return decodeArray(Array.from(queryResult.values()))
     return []
   }
```
flechtnergalani, line 52: This is no longer used internally but as it was part of the public API I kept it. A decoder method like this seemed not to belong here, so I moved functionality out to the new Decode module. The behavior changes a little: We now not only decode, but also check for non-emptiness first. Otherwise this method would only do what .toJSON() already does.

flechtnergalani:
Timo requested I open this for review and implement type checks on a separate ticket: https://github.com/KILTprotocol/ticket/issues/486

src/attestation/Attestation.chain.ts
```diff
@@ -37,9 +36,17 @@ export async function store(
...
 }
 
-function decode(encoded: QueryResult, claimHash: string): Attestation | null {
-  if (encoded && encoded.encodedLength) {
+function decode(
+  encoded: Option<Tuple> | Tuple,
+  claimHash: string
+): Attestation | null {
+  if (
+    encoded instanceof Option ||
```
wischli, line 44: Since this check is done three times, I would recommend to move it to a function like `isNotEmpty`. Also, I would add a comment why it suffices to check for indices 0 and 1 of encoded not to be empty for better clarity.

src/blockchain/Blockchain.ts
```diff
@@ -121,12 +120,8 @@ export default class Blockchain implements IBlockchainApi {
...
 
-  public async getNonce(accountAddress: string): Promise<Codec> {
-    const nonce = await this.api.query.system.accountNonce(accountAddress)
-    if (!nonce) {
-      throw Error(`Nonce not found for account ${accountAddress}`)
-    }
-
+  public async getNonce(accountAddress: string): Promise<u64> {
+    const nonce = await this.api.query.system.accountNonce<u64>(accountAddress)
```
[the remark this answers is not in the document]
flechtnergalani: 1. removed this because it never occurs - the result is always a Codec and never falsy, not even if it is a U64(0)
2. ohh sure, you're right

src/util/Decode.ts
```diff
@@ -0,0 +1,34 @@
...
+export function hasNonNullByte(codec: Codec): boolean {
+  return !codec.toU8a().some(e => e !== 0)
+}
+
+export function decodeArray(codec: Codec[]): AnyJson[] {
+  return codec.map(el => {
+    if (el instanceof Array) {
+      return decodeArray(el)
+    }
+    if (el.isEmpty && el.toJSON) {
```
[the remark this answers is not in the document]
flechtnergalani: Some Codec types return true for `isEmpty` on falsy values; a U64(0) would return true for `isEmpty`, as would a Vec(..., []). If a decoded value is falsy, I see no point in replacing it with `null`. In fact, this may even lead to buggy behavior, I thought. 
Do you think that's to fiddly ? I should deffo explain it well in a doc comment

src/attestation/Attestation.chain.ts
flechtnergalani: There is nothing systematic about checking indices `0` and `1` here; they just happen in these specific cases to be of types that would give away if the whole thing was invalid (we always expect a hash and an account id). I should add comments.
In fact, I think these checks are superfluous; at least they are if any one of the following assumptions are true:
1. A valid Hash will always have at least one non-zero byte when represented as a byte array
2. A valid AccountId will always have at least one non-zero byte when represented as a byte array

Tuples are just concatenations of the individual encoded values (https://substrate.dev/docs/en/conceptual/core/codec), so if either byte representation can tell us whether we have empty values or not, we'd see it in the Tuple's byte representation

flechtnergalani: I'm tempted to remove this last part but not 100 % sure the checking of non-null byte is reliable. What do you think I should do?

src/util/Decode.ts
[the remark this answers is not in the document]
wischli: Should definitely be explained. But I still think the check of `el.toJSON` is insufficient here in that it should be at least `el.toJSON()`?!

src/attestation/Attestation.chain.ts
wischli: I would try to break it in a test and check whether the second check for indices 0 and 1 is really needed. Secondly, I would check your assumptions, e.g. by checking, whether you can create a SHA-256 hash and AccountId from a 0-byte array.

src/util/Decode.ts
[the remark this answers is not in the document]
flechtnergalani: f*** it, decided to remove the function again and reverted `asArray()` - there's more problems to it, and as long as I can't produce a clean implementation, I'd rather not add this at all. It's not part of the ticket anyways.

src/attestation/Attestation.chain.ts
flechtnergalani: Well, you can, and that is the issue:
not so much for a hash, where the string representation is just the hex:
`0x0000000000000000000000000000000000000000000000000000000000000000`
but quite prominently for an AccountId, as this is the string representation of a 0-byte array:
`5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqi1eSuyUpnhM` 
I guess your questions concerns whether this could be any actual account's address, but this I just don't know... 
I do know that both are basically just byte arrays of fixed length (`Ua8Fixed`) for which the `isEmpty` property is defined as "Returns true if the type wraps an empty/default all-0 value".

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Checks nested codec types against a type description string. Uses `codec.toRawType()` internally.
+ *
+ * @param codec Codec to type check.
+ * @param types String or array of strings to check against.
+ * @throws `TypeError` If codec type is not contained in types.
+ */
+export function assertCodecIsType(
```
wischli, line 44: Is it necessary to distinguish between an array and a single value for `types`? It seems to me that we could solely focus on an array and in case of a single value just use an array of length one.

src/delegation/DelegationDecoder.ts
```diff
@@ -69,24 +85,40 @@ function verifyRoot(rootId: string): boolean {
   )
 }
 
+export type DelegationNodeRecord = Pick<
```
wischli, line 88: I guess you have already noticed that this causes multiple problems currently. Nevertheless, I am a fan of picking 👍

src/delegation/DelegationDecoder.ts
```diff
@@ -69,24 +85,40 @@ function verifyRoot(rootId: string): boolean {
...
+  DelegationNode,
+  'rootId' | 'parentId' | 'account' | 'permissions' | 'revoked'
```
wischli, line 90: See above.

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Dummy comment needed for correct doc display, do not remove.
+ */
+
+/**
+ * Checks for a non-null byte in the Uint8Array representation
+ * of the codec. This is a pretty sound indicator that the
+ * codec does not hold an empty default value. The reverse inference
+ * does not apply! Valid values could be encoded only with 0 bytes.
+ *
+ * @param codec Codec type to check for non-empty bytes.
```
wischli, line 17: The general jsdoc style in the SDK includes starting with _The_ or _A_ (e.g. _The codec type..._). We should keep this consistent.

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @returns True if codec is non-empty; false if it may be empty.
```
wischli, line 18: Proposal 
```javascript
`* @returns Whether the codec is non-empty, e.g. in case of `false` it may be empty.
```

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export function hasNonNullByte(codec: Codec): boolean {
+  return !codec.toU8a().some(e => e !== 0)
+}
+
+/**
+ * Checks nested codec types against a type description string. Uses `codec.toRawType()` internally.
+ *
+ * @param codec Codec to type check.
```
wischli, line 27: See above (jsdoc guidelines in SDK).

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param types String or array of strings to check against.
```
wischli, line 28: See above (jsdoc guidelines in SDK).

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @returns `boolean` true if codec type is contained in types, false otherwise.
```
wischli, line 29: Proposal:
```javascript
* @returns Whether the codec type is contained in the input types, e.g. in case of `false` the codec type is missing.
```

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+export function codecIsType(codec: Codec, types: string[] | string): boolean {
+  return types instanceof Array
+    ? types.includes(codec.toRawType())
+    : types === codec.toRawType()
+}
+
+/**
+ * Checks nested codec types against a type description string. Uses `codec.toRawType()` internally.
+ *
+ * @param codec Codec to type check.
```
wischli, line 40: See above (jsdoc guidelines in SDK).

src/util/Decode.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param types String or array of strings to check against.
```
wischli, line 41: See above (jsdoc guidelines in SDK).

src/util/Decode.ts
flechtnergalani: sure, if that helps clarity!

src/delegation/DelegationDecoder.ts
flechtnergalani: I solved this

src/delegation/Delegation.ts
```diff
@@ -77,11 +76,9 @@ export default abstract class DelegationBaseNode
     const queryResults: CodecWithId[] = await fetchChildren(childIds)
     const children: DelegationNode[] = queryResults
       .map((codec: CodecWithId) => {
-        const decoded = this.decodeChildNode(codec.codec)
-        if (decoded) {
-          decoded.id = codec.id
-        }
-        return decoded
+        const decoded = decodeDelegationNode(codec.codec)
+        if (decoded) return { ...decoded, id: codec.id } as DelegationNode
```
flechtnergalani, line 80: this should probably use the constructor - otherwise we'd circumvent the constructor input checks

src/actor/Attester.spec.ts
```diff
@@ -149,7 +149,7 @@ describe('Attester', () => {
     }
   })
 
-  it('Revoke privacy enhanced attestation', async () => {
+  xit('Revoke privacy enhanced attestation', async () => {
```
wischli, line 152: This test should be fixed before merging. Is there a problem Albi or I should be helping with?

src/delegation/Delegation.ts
wischli: I agree!

src/util/Decode.ts
```diff
@@ -0,0 +1,49 @@
...
+export function hasNonNullByte(codec: Codec): boolean {
+  return !codec.toU8a().some(e => e !== 0)
+}
+
+/**
+ * Checks nested codec types against a type description string. Uses `codec.toRawType()` internally.
+ *
+ * @param codec The codec to type check.
+ * @param types An array of strings denoting types to check against.
+ * @returns Whether the codec of any of the allowed `types`.
```
wischli, line 29: At least one word is missing here 😅

src/blockchain/Blockchain.ts
```diff
@@ -128,12 +123,7 @@ export default class Blockchain implements IBlockchainApi {
...
 
-  public async getNonce(accountAddress: string): Promise<Codec> {
-    const nonce = await this.api.query.system.accountNonce(accountAddress)
-    if (!nonce) {
-      throw Error(`Nonce not found for account ${accountAddress}`)
-    }
-
-    return nonce
+  public async getNonce(accountAddress: string): Promise<u64> {
+    return this.api.query.system.accountNonce<u64>(accountAddress)
```
wischli, line 127: What happens now when there is no nonce or can this not occur anymore?

jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   testEnvironment: '../jest.env.js',
   clearMocks: true,
   runner: 'groups',
+  testTimeout: 10000,
```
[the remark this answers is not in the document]
flechtnergalani: I get failing tests because the `beforeAll()` steps time out on many occasions. So you'd argue for setting the timeouts there? My only issue is that when we realise it's not enough for some application, we would have to update this in a lot of places. I could use a const for this though.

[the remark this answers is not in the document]
wischli: In the `CI` or also on the unit tests? How many tests are failing?

src/delegation/Delegation.ts
flechtnergalani: fixed that now!

jest.config.js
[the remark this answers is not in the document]
flechtnergalani: on my local machine, actually. 68 are failing, 19 / 25 test suites

src/actor/Attester.spec.ts
flechtnergalani: Yeah I'd need help with this. Albi told me to open a ticket and disable this instead. Do you have a quick solution? This is also silently failing on develop

src/blockchain/Blockchain.ts
flechtnergalani: Can't happen - it will return the fallback value which is `0`.

flechtnergalani: but this was always the case. The code was just misleading so far

src/util/Decode.ts
flechtnergalani: ouups

src/actor/Attester.spec.ts
wischli: We will investigate this in a ticket, thanks for telling us!

flechtnergalani: nice! you saw https://github.com/KILTprotocol/ticket/issues/583 right?

wischli: Yes!

jest.config.js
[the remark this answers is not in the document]
wischli: Wow this is severe. Then keep it as is!

src/util/Decode.ts
wischli: After this is fixed, everything has been resolved for me.

flechtnergalani: fixed it!

PR #268, opened 2020-06-16
fix: KILT-AC demo

weichweich:
# closes KILTProtocol/ticket#579

- small improvements to the example file
- check in the integration tests that we have a blockchain connection. This improves the error messages (not printing `undefined has no api` but the real error message).

# How to test:

# Checklist:


src/__integrationtests__/ChainConnectivity.spec.ts
```diff
@@ -8,20 +8,24 @@ import { Header } from '@polkadot/types/interfaces/types'
...
-      nodeName: 'substrate-node',
-      nodeVersion: expect.stringMatching(/.+\..+\..+/),
-    })
+      expect(stats).toMatchObject({
+        chain: 'Development',
+        nodeName: 'substrate-node',
+        nodeVersion: expect.stringMatching(/.+\..+\..+/),
+      })
+    } else {
+      expect(false).toBeTruthy()
```
LeonFLK, line 27: We could explicitly test for the existence of blockchain, instead of indirectly failing the tests if equal to undefined:

expect( typeof blockchain !== undefined ).toBeTruthy()

weichweich: The problem there is that we then have no typeguard. `blockchain` could be undefined for the rest of the test and the linting would show that as an error. Although the test would never move beyond the `expect()`.

LeonFLK: This problem can be circumvented with a non-null assertion { `blockchain!.getStats()` } once we are certain that it exists.

IMO it would be just a visibility improvement, since failed tests would state why.

weichweich: Yes I would also like to have the `expect( typeof blockchain !== undefined ).toBeTruthy()`. But the non-null assertion also triggers a warning. But ok maybe we just mark that warning as ignored...

LeonFLK: yeah just disable ts warning for this file for non null assterions

weichweich: seems like you don't event need to disable the warning because there is no warning. 🤷

weichweich: > seems like you don't event need to disable the warning because there is no warning. 🤷

ok i take that back!

PR #271, opened 2020-06-18
test: fix mocked types

weichweich:
## Checklist:

### DIFF
--- a/src/actor/Verifier.spec.ts
+++ b/src/actor/Verifier.spec.ts
@@ -64,7 +64,7 @@ describe('Verifier', () => {
 
     blockchainApi.query.attestation.attestations.mockReturnValue(
       new Option(Tuple.with([H256, AccountId, 'Option<H256>', Bool]), [
-        '0x1234', // ctype hash
+        '"0xde9f624875aa620d06434603787a40c8cd02cc25c7b775cf50de8a3a96bbeafa"', // ctype hash
         attester.getAddress(), // Account
         null, // delegation-id?
         false, // revoked flag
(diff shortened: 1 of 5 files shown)


flechtnergalani:
`any` chance you can get rid of the `any[]` in the signatures on `Verifier.ts` ?

src/actor/Verifier.spec.ts
```diff
@@ -63,13 +63,12 @@ describe('Verifier', () => {
...
     blockchainApi.query.attestation.attestations.mockReturnValue(
-      new Option(
-        Tuple,
-        new Tuple(
-          [H256, AccountId, Text, Bool],
-          [cType.hash, attester.getAddress(), undefined, false]
-        )
-      )
+      new Option(Tuple.with([H256, AccountId, 'Option<H256>', Bool]), [
+        '0x1234', // ctype hash
```
LeonFLK, line 67: We should try to only use real hashes, and not these mocks, as this will be decoded to something unequal to 0x1234

weichweich:
> `any` chance you can get rid of the `any[]` in the signatures on `Verifier.ts` ?

good catch! That's not that easy to solve. I bodged that a little bit for now. The gabi and public credentials are not the same. We should make them the same.  This is [tracked in an issue](KILTprotocol/ticket/issues/585)

src/actor/Verifier.spec.ts
weichweich: I DID IT! 06e712f069524cbf1dd4ca1e8ebab7d63a4b996d

PR #272, opened 2020-06-18
feat: required fields in ctype wrapper schema

tjwelde:
## fixes KILTProtocol/ticket#204
This PR gets the required fields from the ctype schema and adds it in the schema section of the wrapper schema. Before that, some fields were not marked as required.

## How to test:
- Tests should run
- Making a ctype and using it to make a claim in demo-client has to be possible
  - To test, cherry-pick this commit onto sdk master branch and use it in the demo-client

## Checklist:

### DIFF
--- a/src/ctype/CTypeSchema.ts
+++ b/src/ctype/CTypeSchema.ts
@@ -16,12 +16,11 @@ export const CTypeModel = {
     $schema: {
       type: 'string',
       format: 'uri',
-      default: 'http://kilt-protocol.org/draft-01/ctype#',
-      enum: ['http://kilt-protocol.org/draft-01/ctype#'],
+      const: 'http://kilt-protocol.org/draft-01/ctype#',
     },
     type: {
       type: 'string',
-      enum: ['object'],
+      const: 'object',
     },
     properties: {
       type: 'object',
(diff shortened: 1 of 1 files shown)


src/ctype/CTypeSchema.ts
```diff
@@ -16,12 +16,11 @@ export const CTypeModel = {
...
       type: 'string',
       format: 'uri',
-      default: 'http://kilt-protocol.org/draft-01/ctype#',
-      enum: ['http://kilt-protocol.org/draft-01/ctype#'],
+      const: 'http://kilt-protocol.org/draft-01/ctype#',
     },
     type: {
       type: 'string',
-      enum: ['object'],
+      const: 'object',
```
Dudleyneedham, line 23: Is this a new JSON schema keyword?

tjwelde: It was already available in json draft-07, which we are still using. We just didn't use `const`.

Dudleyneedham: Awesome!

Dudleyneedham: We will need to change the demo-client accordingly. 

Do we need a ticket?

tjwelde: I checked the demo-client and no change is needed.

PR #274, opened 2020-06-25
feat: publishing to github packages from develop

flechtnergalani:
## fixes KILTProtocol/ticket#599
Builds and publishes the sdk as a npm package to the GitHub Packages repo.
These releases are versioned with the current commit id and tagged `dev`.

Possible issues:
 - [x] there is a (almost empty) `.npmrc` in this repo that may interfere with the publishing process
     - removed it after talking to Dudley, who added it
 - [x] the `package.json` mentions a target for publishing (`publishConfig`) that could potentially interfere
     - does not seem to interfere

## How to test:
By adding this branch to the triggers I have successfully triggered a release. You can find the published package in https://github.com/KILTprotocol/sdk-js/packages/286306.

## Checklist:

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-tag-version-prefix=""
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+tag-version-prefix=""
--- a/.github/workflows/devpackpublish.yml
+++ b/.github/workflows/devpackpublish.yml
@@ -1,0 +1,19 @@
+name: Publish package to github packages
+on:
+  push:
+    branches:
+      - develop
+jobs:
+  build:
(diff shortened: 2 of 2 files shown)


.npmrc
```diff
@@ -1 +0,0 @@
-tag-version-prefix=""
```
tjwelde, line 1: The tag version prefix was necessary for doing the public release to npm registry. Is it obsolete?

flechtnergalani: I don't actually know - it looked like Dudley introduced this with his last PR and he said it was irrelevant, so I removed it. If we still need it, I can put it back in

flechtnergalani: image

PR #275, opened 2020-06-26
Feat: Debug mode and modify logLevel

LeonFLK:
## fixes KILTProtocol/ticket#496

Changes the default logLevel from Debug to Error, with the option to enable Debug mode with env variable DEBUG.

Adds new exported function to change the logLevel of all factories to the specified level.

## How to test:

start with or without DEBUG env set to true, call modifyLogLevel, see many logs or not.

## Checklist:

### DIFF
--- a/src/config/ConfigLog.ts
+++ b/src/config/ConfigLog.ts
@@ -11,6 +11,8 @@ import {
   LoggerFactoryOptions,
   LogGroupRule,
   LogLevel,
+  getLogControl,
+  LogGroupControlSettings,
 } from 'typescript-logging'
 
 // Create options instance and specify 2 LogGroupRules:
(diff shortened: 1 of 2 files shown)


LeonFLK:
tests were failing due to unreached 80% function coverage

src/config/ConfigLog.ts
```diff
@@ -11,19 +11,44 @@ import {
...
 // eslint-disable-next-line import/prefer-default-export
 export const factory = LFService.createNamedLoggerFactory(
   'LoggerFactory',
   options
 )
+
+/**
+ *  Changes all existing Loggers of our default Factory with id 0 to the intended Level.
+ *
+ * @param level The intended LogLevel.
```
wischli, line 39: Could potentially add that this ranges from 0 to 5 but do as you wish.

wischli:
> tests were failing due to unreached 80% function coverage

Yeah I removed the config from the coverage.

LeonFLK:
Tests run ~8% faster with LogLevel reduced from Debug to Error yey

LeonFLK:
> > tests were failing due to unreached 80% function coverage
> 
> Yeah I removed the config from the coverage.

I added Config.spec.ts that checks the modify function and the default level.

src/config/ConfigLog.ts
LeonFLK: Yes good addition.

I wish though the LogLevel enum had this kind of information at runtime.

PR #276, opened 2020-07-01
feat: added query of delegation root id

LeonFLK:
## fixes KILTProtocol/ticket#490

Adds static as well as class method for retrieving the rootNode from the delegationId supplied with Attestation
## Checklist:


LeonFLK:
Function test coverage failing

PR #278, opened 2020-07-02
test: attested claim coverage

LeonFLK:
Fixes AttestedClaim.ts uncovered functions

Related to: https://github.com/KILTprotocol/ticket/issues/606

### DIFF
--- a/src/attestation/Attestation.spec.ts
+++ b/src/attestation/Attestation.spec.ts
@@ -106,7 +106,7 @@ describe('Attestation', () => {
       requestForAttestation,
       identityAlice.getPublicIdentity()
     )
-    expect(await Attestation.verify(attestation)).toBeFalsy()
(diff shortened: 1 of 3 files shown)


LeonFLK:
> Looks good but we should also include tests for lines _143_ and _152-164_ of `AttestedClaim.ts` to get full coverage (currently 83.33%), i.e. `verifyData` and `getHash`
> 
> Screenshot 2020-07-03 at 08 46 42

With the error case I agree, we definitely should have tests including this case.
The getHash() one I will also include, but the Class method that just calls the static is not needed I reckon.
What do you think?

wischli:
> With the error case I agree, we definitely should have tests including this case.
> The getHash() one I will also include, but the Class method that just calls the static is not needed I reckon.
> What do you think?

Agree!

PR #281, opened 2020-07-06
test: add privacy enhanced attestation to CI

wischli:
## fixes KILTProtocol/ticket#516
- adds CI tests for privacy enhanced attestation
 - basic privacy enhanced attestation setup w/ and w/o revocation
 - basic privacy enhanced verification setup w/ and w/o revocation

Note: A privacy enhanced verification should not reveal any unique claimer data. Unfortunately, when sending a message to the verifier, both the claimer's `senderAddress` and the `senderBoxPublicKey` are revealed. However, this is an unrelated issue and [tracked here](https://github.com/KILTprotocol/ticket/issues/609).

## How to test:
[Check CI](https://github.com/KILTprotocol/sdk-js/pull/281/checks?check_run_id=841667988#step:6:62). If you still want to run it on your machine, do any of these two

### 1) fast + dirty
```
docker pull kiltprotocol/mashnet-node:develop
docker run -d -p 9944:9944 kiltprotocol/mashnet-node:develop ./target/release/mashnet-node --dev --ws-port 9944 --ws-external --rpc-external
yarn test:integration:run AttestationPE
docker kill $(docker ps -q)
```

### 2) slow + tidy

```
docker build -t local/sdk-js .
docker-compose run sdk yarn test:integration:run AttestationPE && docker-compose down
```

## Checklist:


wischli:
> 2\. it's not great when it comes to interpreting failing tests. Everything that depends on a failing `beforeAll` will be marked as failing with the error encountered in `beforeAll` or with some cryptic `undefined` error.

Definitely agree. I am also not happy with this... could run the `beforeAll` setup in a test beforehand to catch this. What do you think?

rflechtner:
You are talking about having it run in an `it` instead of a `beforeAll`, right? While that is of course committing the cardinal sin of building non-independent tests, the solution with `beforeAll` just veils the fact that tests are, in effect, also not independent... so I'd almost prefer having it in an `it`.
Test execution order of course is something we need to consider in this case. By default tests are executed from top to bottom, so that's fine; as long as we don't use the random execution order functionality that jest also offers.

wischli:
> You are talking about having it run in an `it` instead of a `beforeAll`, right? While that is of course committing the cardinal sin of building non-independent tests, the solution with `beforeAll` just veils the fact that tests are, in effect, also not independent... so I'd almost prefer having it in an `it`.

What I meant was running the setup in an "independent" test executing the setup (if run sequentially). Then, errors in the setup (which are sometimes hard to nail down) will at least be caught in the test. Unfortunately, I still need the setup for different verifications. But I removed the attestation one and moved everything from initiating the attestation to building the presentation to the same setup.

Please have a look and if you don't have any objections, I would be happy to merge this 😊

wischli:
Merging as just had a brief chat about it.

PR #282, opened 2020-07-07
test: improve coverage for BlockchainApiConnection

wischli:
## related to KILTProtocol/ticket#606
- improves coverage for `BlockchainApiConnection` by connecting to testnet

## How to test:
```
yarn jest BlockchainApiConnection
```

## Checklist:

### DIFF
--- a/src/blockchainApiConnection/BlockchainApiConnection.spec.ts
+++ b/src/blockchainApiConnection/BlockchainApiConnection.spec.ts
@@ -1,0 +1,36 @@
+import Blockchain from '../blockchain'
+import getCached, { clearCache } from './BlockchainApiConnection'
+
+const DEVNET_WS_ADDRESS = 'ws://full-nodes.devnet.kilt.io:9944/'
+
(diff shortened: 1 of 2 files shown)


wischli:
> I'm not sure I agree with grouping this under unit tests; shouldn't those be able to run without any sort of connection?

I wasn't sure either and assume you are right. I like the idea of improving grouping but I guess this should be done after the Substrate update. 

Do you know why we exclude CI tests from the coverage? Is this the default go-to?

rflechtner:
> I wasn't sure either and assume you are right. I like the idea of improving grouping but I guess this should be done after the Substrate update.

For now we could group this under integration, or does that defy the purpose?

> Do you know why we exclude CI tests from the coverage? Is this the default go-to?

I wasn't aware we are doing this... We also don't collect coverage from the `blockchain` directory, which we should. It may be related to the old way of mocking things?

wischli:
Closing this as this both outdated and irrelevant.

PR #283, opened 2020-07-07
test: improve actor coverage

wischli:
## related to KILTProtocol/ticket#606
- pushes actor test coverage from ~70% to nearly 100%
- adds new SDK error for privacy enhanced verification
- fixes uncaught error in privacy enhanced verification

## How to test:
```
yarn jest actor/
```

## Checklist:


src/actor/Verifier.spec.ts
```diff
@@ -207,6 +239,117 @@ describe('Verifier', () => {
     expect(owner).toBeDefined()
     expect((claims[0] as IRequestForAttestation).claim).toEqual(unownedClaim)
   })
+  it('should throw ... during privacy enhanced presentation', async () => {
```
wischli, line 242: Note: I kept all throws in the same test to avoid re-running the "setup". Also, most tests are basically the same with permutated input.

src/actor/Verifier.ts
```diff
@@ -215,26 +239,24 @@ export async function verifyPresentation(
...
   if (message.body.type === MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_PE) {
-    if (
-      typeof latestAccumulators === 'undefined' ||
-      typeof attesterPubKeys === 'undefined'
-    ) {
-      throw new Error(`Received privacy enhanced presentation. Require:
-      Session: ${session}
-      accumulators: ${latestAccumulators}
-      public keys: ${attesterPubKeys}`)
+    const accFailure = !verificationCheck(_latestAccumulators, true)
```
wischli, line 242: Note: This function fixes the case in which any of `latestAccumulators` or `attesterPubKeys` are anything but `undefined`

src/actor/Claimer.spec.ts
```diff
@@ -296,4 +302,103 @@ describe('Claimer', () => {
...
+              ({ privacyCredential: null } as unknown) as Credential,
+            ],
+          })
+        ).rejects.toThrowError(ERROR_PE_CREDENTIAL_MISSING())
+      })
+      it('Should throw when PE is required but not allowed', () => {
+        return expect(fakePresentation({})).rejects.toThrowError(
+          ERROR_PE_MISMATCH()
+        )
+      })
```
rflechtner, line 362: this looks like a duplicate from line 343++

src/actor/Verifier.spec.ts
```diff
@@ -142,13 +147,40 @@ describe('Verifier', () => {
...
+        legitimations: true,
+        delegation: true,
+      })
+      .finalize(true, verifier, claimer.getPublicIdentity())
+    expect(session).toBeDefined()
+    expect(session.requestedProperties[0].properties).not.toContain(
+      'claim.cTypeHash'
+    )
+    expect(session.requestedProperties[0].properties).toContain('delegationId')
+    expect(session.requestedProperties[0].properties).toContain('delegationId')
```
rflechtner, line 176: one of them was supposed to be the legitimations, right?

src/actor/Verifier.ts
```diff
@@ -182,13 +185,34 @@ async function verifyPublicPresentation(
...
+ * Checks whether each value is either an instance of an Accumulator or a [[PublicAttesterIdentity]].
+ *
+ * @param values Either an array or undefined which should be checked.
+ * @param isAcc A boolean to display whether we check for Accumulators.
+ * @returns True if the supplied values are correct.
+ */
+function verificationCheck(
+  values:
+    | Array<gabi.Accumulator | PublicAttesterIdentity | undefined>
+    | undefined,
```
rflechtner, line 198: what is the reason this is undefined and not `unknown` ? That's usually what you would use on a type guard, isn't it.

src/actor/Verifier.ts
```diff
@@ -182,13 +185,34 @@ async function verifyPublicPresentation(
...
+  isAcc: boolean
+): boolean {
```
rflechtner, line 200: you've written a type guard, so why not make it explicit for typescript: 
```suggestion
): values is Array<gabi.Accumulator | PublicAttesterIdentity> {
```
However it would be most useful if you'd split this in two different type guards - one for `values is Array<PublicAttesterIdentity>` and one for `values is Array<gabi.Accumulator>` - while also dropping the `isAcc` argument.

src/actor/Claimer.spec.ts
wischli: Oops, you are correct.

src/actor/Verifier.ts
wischli: That is correct, thanks. Fixedin 5db18c9. 

I tried to add a general array typeguard with generics but this does not work since type information are removed at compile time such that `instanceof T` (whereas `T` is some generic) does not exist at runtime.

**Example**

```typescript
function arrayTypeguard<T>(values: T[] | unknown): values is T[] {
  if (Array.isArray(values) && values.length) {
    return values.every((a) => a instanceof T)
  }
  return false
}
```

**Error**
```
'T' only refers to a type, but is being used as a value here.ts(2693)
```

src/actor/Verifier.ts
wischli: Correct, fixed in 5db18c9

src/actor/Verifier.spec.ts
wischli: Fixed in 5db18c9

src/actor/Claimer.spec.ts
wischli: Fixed in 5db18c9

PR #284, opened 2020-07-08
fix: expected encoded types

LeonFLK:
## related to KILTProtocol/ticket#621

I just changed the expected types to the ones my dev chain reported to the demo-client.
Maybe this can help figure out what's really going on, this will help update and troubleshoot the demo client.

The tests will fail as they expect the previously expected types.

## Checklist:


LeonFLK:
Okay the integration tests do not like this.

demo-client y u do dis?

LeonFLK:
The latest commit just added the original expected types as part of the possible types.

wischli:
Closing in favor of #284

PR #286, opened 2020-07-09
fix: decoding types

wischli:
## fixes KILTProtocol/ticket#621

One reason for the error was that we did not register our custom chain types for local tests. For most mockings we did not match the return type of the chain. This PR matches the mocking types and expected types after decoding with the return types from the chain 1:1.

- fixed errors for expected decoding types
- refactored chain mocking for less inconsistencies and boilerplate
- added function `mockChainQueryReturn` which wraps your mocking values into Option/Vec with the correct types
- replaced `H256` with `Hash` as the on-chain types are actually `Hash` which point to `H256`, so either of both are fine but `Hash` is the desired endpoint to be used in case we switch the hash at some point
- replaced multiple meaningless `TypeRegistry`s  with a global one `TYPE_REGISTRY`
- had to upgrade to latest polkadot versions to make use of a fix for the actual types we get from the chain
- this PR also updates to latest Portablegabi version (includes latest Polkadot dependencies and WASM webpack fix)

## How to test:
Best way would be to test this build with `yalc` against your issues in the DemoClient etc. My problem when verifying a non-PE attested claim was solved.

### default
```
yarn test
docker-compose run sdk yarn test:integration:run && docker-compose down
```

## Checklist:


src/delegation/DelegationRootNode.spec.ts
```diff
@@ -11,31 +9,25 @@ describe('Delegation', () => {
...
     )
     require('../blockchainApiConnection/BlockchainApiConnection').__mocked_api.query.delegation.delegations.mockReturnValue(
-      new Option(
-        registry,
-        Tuple.with(
-          // Root-Delegation: delegation-id -> (root-id, parent-id?, account, permissions, revoked)
-
-          ['H256', AccountId, Bool]
-        ),
-        [ctypeHash, identityAlice.getAddress(), false]
```
wischli, line 37: Wondering why this worked out in the first place. The mock here was based on `root` and should not have worked for `delegations`.

src/blockchainApiConnection/__mocks__/BlockchainQuery.ts
```diff
@@ -0,0 +1,165 @@
...
+} from '@polkadot/types'
+import AccountId from '@polkadot/types/generic/AccountId'
+import Bool from '@polkadot/types/primitive/Bool'
+import U64 from '@polkadot/types/primitive/U64'
+import U8 from '@polkadot/types/primitive/U8'
+import { Codec } from '@polkadot/types/types'
+import { Constructor } from '@polkadot/util/types'
+
+const TYPE_REGISTRY = new TypeRegistry()
+TYPE_REGISTRY.register({
```
wischli, line 17: Without this, the local tests would not know what `DelegationNodeId` etc. point to.

LeonFLK:
Just went over it, LGTM.

Will use this for testing purposes.

PR #290, opened 2020-07-16
feat: option for disabling gabi

tjwelde:
## fixes KILTProtocol/ticket#634
This PR adds options to methods to disable calls to portablegabi, so that the wasm file is not loaded.

## How to test:
- Tests should run

## Checklist:

### DIFF
--- a/src/actor/Claimer.spec.ts
+++ b/src/actor/Claimer.spec.ts
@@ -16,7 +16,11 @@ import {
 } from '../errorhandling/SDKErrors'
 import AttesterIdentity from '../identity/AttesterIdentity'
 import Identity from '../identity/Identity'
-import Message, { MessageBodyType } from '../messaging/Message'
+import Message, {
+  MessageBodyType,
+  IRequestClaimsForCTypes,
+  ISubmitClaimsForCTypesPublic,
+} from '../messaging/Message'
 import constants from '../test/constants'
 import { ClaimerAttestationSession } from './Claimer'
 
@@ -301,6 +305,33 @@ describe('Claimer', () => {
       MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_PUBLIC
     )
     expect(Array.isArray(presentation.body.content))
+  })
+  it('create public presentation from request without peRequest', async () => {
+    const body: IRequestClaimsForCTypes = {
+      type: MessageBodyType.REQUEST_CLAIMS_FOR_CTYPES,
+      content: {
+        allowPE: false,
+        ctypes: ['this is a ctype hash'],
+      },
+    }
+    const request = new Message(body, verifier, claimer.getPublicIdentity())
+
+    const presentation = await Claimer.createPresentation(
+      claimer,
+      request,
+      verifier.getPublicIdentity(),
+      [credentialPE],
+      [attester.getPublicIdentity()],
+      false
+    )
+    expect(presentation.body.type).toEqual(
+      MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_PUBLIC
+    )
+    expect(Array.isArray(presentation.body.content))
+    const { content } = presentation.body as ISubmitClaimsForCTypesPublic
+    expect(Object.keys(content[0].request.claim.contents)).toEqual(
+      Object.keys(content[0].request.claimHashTree)
+    )
   })
   describe('Negative tests', () => {
     describe('create presentation', () => {
--- a/src/actor/Claimer.spec.ts
+++ b/src/actor/Claimer.spec.ts
@@ -13,6 +13,7 @@ import {
   ERROR_MESSAGE_TYPE,
   ERROR_PE_CREDENTIAL_MISSING,
   ERROR_PE_MISMATCH,
+  ERROR_IDENTITY_NOT_PE_ENABLED,
 } from '../errorhandling/SDKErrors'
 import AttesterIdentity from '../identity/AttesterIdentity'
 import Identity from '../identity/Identity'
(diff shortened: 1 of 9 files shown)


src/actor/Claimer.spec.ts
```diff
@@ -302,6 +307,33 @@ describe('Claimer', () => {
...
+      request,
+      verifier.getPublicIdentity(),
+      [credentialPE],
+      [attester.getPublicIdentity()],
+      false
+    )
+    expect(presentation.body.type).toEqual(
+      MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_PUBLIC
+    )
+    expect(Array.isArray(presentation.body.content))
```
wischli, line 331: Missing `.toBeTruthy()` or `.toBe(true)` here. 

```suggestion
    expect(Array.isArray(presentation.body.content)).toBe(true)
```

src/actor/Claimer.spec.ts
```diff
@@ -355,6 +387,28 @@ describe('Claimer', () => {
           })
         ).rejects.toThrowError(ERROR_PE_CREDENTIAL_MISSING())
       })
+      it('Should throw, if PE is allowed, but claimer has no PE-enabled identity', async () => {
+        const claimerWithoutPE = await Identity.buildFromURI('//bob', {
```
wischli, line 391: Not important, but this creates a different identity than from the `//Bob` URI.

```suggestion
        const claimerWithoutPE = await Identity.buildFromURI('//Bob', {
```

src/actor/Claimer.spec.ts
```diff
@@ -395,5 +449,86 @@ describe('Claimer', () => {
...
+      blockchainApi.query.attestation.attestations.mockReturnValue(
+        mockChainQueryReturn('attestation', 'attestations', [
+          cType.hash,
+          attester.getAddress(),
+          undefined,
+          0,
+        ])
+      )
+
+      const claimerWithoutPE = await Identity.buildFromURI('//bob', {
```
wischli, line 462: See above.
```suggestion
      const claimerWithoutPE = await Identity.buildFromURI('//Bob', {
```

src/actor/Claimer.spec.ts
```diff
@@ -395,5 +449,86 @@ describe('Claimer', () => {
...
```
wischli, line 493: See above.

```suggestion
      const claimerWithoutPE = await Identity.buildFromURI('//Bob', {
```

src/actor/Claimer.spec.ts
```diff
@@ -395,5 +449,86 @@ describe('Claimer', () => {
...
+      )
+
+      await expect(
+        Claimer.buildCredential(
+          claimerWithoutPE,
+          attestationMessage,
+          claimerSession
+        )
+      ).rejects.toThrowError(ERROR_IDENTITY_NOT_PE_ENABLED())
+    })
```
wischli, line 532: I would also add an expected success when using the claimer with PE

```suggestion
      await expect(
        Claimer.buildCredential(claimer, attestationMessage, claimerSession)
      ).resolves.toBeTruthy()
    })
```

src/actor/Claimer.ts
```diff
@@ -77,6 +79,12 @@ export async function createPresentation(
     if (!noNulls(peCreds)) {
       throw ERROR_PE_CREDENTIAL_MISSING()
     }
+    if (!identity.claimer) {
+      throw ERROR_IDENTITY_NOT_PE_ENABLED()
+    }
+    if (!request.content.peRequest) {
+      throw ERROR_PE_MISSING()
```
wischli, line 86: Currently, we are not testing this throw. Maybe we should.

src/identity/Identity.ts
```diff
@@ -108,29 +117,36 @@ export default class Identity {
...
    * @example ```javascript
    * const seed =
    *   '0x6ce9fd060c70165c0fc8da25810d249106d5df100aa980e0d9a11409d6b35261';
    * await Identity.buildFromSeedString(seed);
    * ```
    */
-  public static async buildFromSeedString(seedArg: string): Promise<Identity> {
+  public static async buildFromSeedString(
+    seedArg: string,
+    { peEnabled = true }: IdentityBuildOptions = {}
```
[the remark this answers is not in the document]
wischli: I am not sure whether I am a fan of removing the default value here. 
- If you have it, you don't have to think about the options if you don't care.
- If you don't have it, you always have to provide an options object, forcing you to think about it.

src/identity/Identity.ts
```diff
@@ -92,7 +98,10 @@ export default class Identity {
    * await Identity.buildFromMnemonic(mnemonic);
    * ```
    */
-  public static async buildFromMnemonic(phraseArg?: string): Promise<Identity> {
+  public static async buildFromMnemonic(
+    phraseArg?: string,
+    { peEnabled = true }: IdentityBuildOptions = {}
```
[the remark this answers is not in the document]
tjwelde: There are two different defaults involved.
First, we have a default, when the second parameter is not given (`... = {}`) and second, we have a default argument, when an object was given, but doesn't contain `peEnabled `, so we set it there to `peEnabled = true`

src/identity/Identity.ts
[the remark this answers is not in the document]
tjwelde: see above. This would fail, if the second parameter is not given (and thus `undefined`)

src/identity/Identity.ts
```diff
@@ -142,13 +158,19 @@ export default class Identity {
    * await Identity.buildFromSeed(seed);
    * ```
    */
-  public static async buildFromSeed(seed: Uint8Array): Promise<Identity> {
+  public static async buildFromSeed(
+    seed: Uint8Array,
+    { peEnabled = true }: IdentityBuildOptions = {}
```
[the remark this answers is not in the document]
tjwelde: see above.

src/identity/Identity.ts
```diff
@@ -157,34 +179,42 @@ export default class Identity {
...
+   * @param options.peEnabled - If the identity should be privacy enhanced, or not (default: true).
    * @returns  An [[Identity]].
    * @example ```javascript
    * Identity.buildFromURI('//Bob');
    * ```
    */
-  public static async buildFromURI(uri: string): Promise<Identity> {
+  public static async buildFromURI(
+    uri: string,
+    { peEnabled = true }: IdentityBuildOptions = {}
```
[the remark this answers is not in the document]
tjwelde: see above.

src/actor/Claimer.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

src/actor/Claimer.spec.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

src/actor/Claimer.spec.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

src/actor/Claimer.spec.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

src/actor/Claimer.spec.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

src/actor/Claimer.spec.ts
tjwelde: fixed in ea38166a4ce1b9c442db0264098e7f7ff080847f

PR #292, opened 2020-07-23
feat: gabi opt-in, rename public to classic & require mnemonic

tjwelde:
## fixes KILTProtocol/ticket#644
This PR makes portablgabi opt-in and renames `SUBMIT_CLAIMS_FOR_CTYPES_PUBLIC` to `SUBMIT_CLAIMS_FOR_CTYPES_CLASSIC`

## How to test:
- Tests should run succesfully

## Checklist:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -88,7 +88,7 @@
         "jsdoc/check-tag-names": [
           "warn",
           {
-            "definedTags": ["group"]
+            "definedTags": ["group", "packageDocumentation"]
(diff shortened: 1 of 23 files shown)


weichweich:
It's now a bit confusing that we have an `AttesterIdentity`. The `AttesterIdentity` is actually a `IdentityPEButOnlyForAttester`?

Maybe we just change the default for the verification process? That all claimers by default always show a public credential?

PR #294, opened 2020-07-29
refactor: check for ctype registrations seperately

rflechtner:
## fixes KILTProtocol/ticket#630
This adds a method to check whether a CType's owner property is equal to the owner registered to chain. It also changes the `verifyStored` method to check only whether the ctype is registered to chain without checking the owner. 

## How to test:
...

## Checklist:

### DIFF
--- a/src/ctype/CType.ts
+++ b/src/ctype/CType.ts
@@ -104,14 +104,25 @@ export default class CType implements ICType {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


LeonFLK (approved):
LGTM

rflechtner:
@wischli seems like this was meant for v 0.19 but not merged in time - merging it now, hope we can push it in 0.19.1

PR #298, opened 2020-09-11
Refactor: Change Attestation of verify 

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/520
method return false, when it was revoked, which seems to be counterintuitive, when the function is just called verify

## How to test:
N/A
## Checklist:


tjwelde (approved):
LGTM

PR #299, opened 2020-09-24
explicit transactions & awaiting custom tx status changes

rflechtner:
### fixes KILTProtocol/ticket#691

With this change, functions that require blockchain writes (store Attestation / Ctype, revoke Attestation, make transfer, etc.) will no longer submit the transaction, but only prepare it. A signed SubmittableExtrinsic is returned instead, which the end user then submits by calling its `send()` method (optionally attaching a listener to it) or by passing it to the `submitSignedTx()` helper function, which allows you to promisify and await any tx status change (e.g. `Finalised`, `inBlock`).
This also significantly speeds up integration testing, as we now await only `inBlock`, which is 3 times faster than awaiting finalisation.

We in part solve issues here that we wanted the tx subscription handler for and allow people to implement their own, so this may be closed as well - we could still think about publishing what we already created though, as a goodie or template.

As of now, anything in the `/actor` module will still await finalisation, both because those are fairly high-level and because KILT integrates with portable gabi at this point - in order to change this, we would probably have to make changes there as well. Input on how to handle this issue is appreciated.

## How to test:
unit & integration tests

## Checklist:

### DIFF
--- a/src/__integrationtests__/Balance.spec.ts
+++ b/src/__integrationtests__/Balance.spec.ts
@@ -14,6 +14,7 @@ import {
   AWAIT_IN_BLOCK,
   AWAIT_READY,
   IBlockchainApi,
+  submitTx,
 } from '../blockchain/Blockchain'
 import getCached, { DEFAULT_WS_ADDRESS } from '../blockchainApiConnection'
 import Identity from '../identity/Identity'
@@ -70,7 +71,7 @@ describe('when there is a dev chain with a faucet', () => {
     listenToBalanceChanges(ident.address, funny)
     const balanceBefore = await getBalance(faucet.address)
     const tx = await makeTransfer(faucet, ident.address, MIN_TRANSACTION)
-    await blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+    await submitTx(tx, AWAIT_IN_BLOCK)
     const [balanceAfter, balanceIdent] = await Promise.all([
       getBalance(faucet.address),
       getBalance(ident.address),
@@ -99,7 +100,7 @@ describe('When there are haves and have-nots', () => {
       richieRich,
       stormyD.address,
       MIN_TRANSACTION
-    ).then((tx) => blockchain.submitTx(tx, AWAIT_IN_BLOCK))
+    ).then((tx) => submitTx(tx, AWAIT_IN_BLOCK))
     const balanceTo = await getBalance(stormyD.address)
     expect(balanceTo.toNumber()).toBe(MIN_TRANSACTION.toNumber())
   }, 40_000)
@@ -108,7 +109,7 @@ describe('When there are haves and have-nots', () => {
     const originalBalance = await getBalance(stormyD.address)
     await expect(
       makeTransfer(bobbyBroke, stormyD.address, MIN_TRANSACTION).then((tx) =>
-        blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+        submitTx(tx, AWAIT_IN_BLOCK)
       )
     ).rejects.toThrowError('1010: Invalid Transaction')
     const [newBalance, zeroBalance] = await Promise.all([
@@ -123,7 +124,7 @@ describe('When there are haves and have-nots', () => {
     const RichieBalance = await getBalance(richieRich.address)
     await expect(
       makeTransfer(richieRich, bobbyBroke.address, RichieBalance).then((tx) =>
-        blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+        submitTx(tx, AWAIT_IN_BLOCK)
       )
     ).rejects.toThrowError()
     const [newBalance, zeroBalance] = await Promise.all([
@@ -138,10 +139,10 @@ describe('When there are haves and have-nots', () => {
     const listener = jest.fn()
     listenToBalanceChanges(faucet.address, listener)
     await makeTransfer(faucet, richieRich.address, MIN_TRANSACTION).then((tx) =>
-      blockchain.submitTx(tx, AWAIT_READY)
+      submitTx(tx, AWAIT_READY)
     )
     await makeTransfer(faucet, stormyD.address, MIN_TRANSACTION).then((tx) =>
-      blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+      submitTx(tx, AWAIT_IN_BLOCK)
     )
 
     expect(listener).toBeCalledWith(
@@ -157,10 +158,10 @@ describe('When there are haves and have-nots', () => {
     listenToBalanceChanges(faucet.address, listener)
     await Promise.all([
       makeTransfer(faucet, richieRich.address, MIN_TRANSACTION).then((tx) =>
-        blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+        submitTx(tx, AWAIT_IN_BLOCK)
       ),
       makeTransfer(faucet, stormyD.address, MIN_TRANSACTION).then((tx) =>
-        blockchain.submitTx(tx, AWAIT_IN_BLOCK)
+        submitTx(tx, AWAIT_IN_BLOCK)
       ),
     ])
     expect(listener).toBeCalledWith(faucet.address)
--- a/src/__integrationtests__/Balance.spec.ts
+++ b/src/__integrationtests__/Balance.spec.ts
@@ -71,7 +71,7 @@ describe('when there is a dev chain with a faucet', () => {
     listenToBalanceChanges(ident.address, funny)
     const balanceBefore = await getBalance(faucet.address)
     await makeTransfer(faucet, ident.address, MIN_TRANSACTION).then((tx) =>
-      submitSignedTx(tx, [IS_IN_BLOCK])
+      submitSignedTx(tx, IS_IN_BLOCK)
     )
     const [balanceAfter, balanceIdent] = await Promise.all([
       getBalance(faucet.address),
@@ -101,7 +101,7 @@ describe('When there are haves and have-nots', () => {
       richieRich,
       stormyD.address,
       MIN_TRANSACTION
-    ).then((tx) => submitSignedTx(tx, [IS_IN_BLOCK]))
+    ).then((tx) => submitSignedTx(tx, IS_IN_BLOCK))
     const balanceTo = await getBalance(stormyD.address)
     expect(balanceTo.toNumber()).toBe(MIN_TRANSACTION.toNumber())
   }, 40_000)
@@ -110,7 +110,7 @@ describe('When there are haves and have-nots', () => {
     const originalBalance = await getBalance(stormyD.address)
     await expect(
       makeTransfer(bobbyBroke, stormyD.address, MIN_TRANSACTION).then((tx) =>
-        submitSignedTx(tx, [IS_IN_BLOCK])
+        submitSignedTx(tx, IS_IN_BLOCK)
       )
     ).rejects.toThrowError('1010: Invalid Transaction')
     const [newBalance, zeroBalance] = await Promise.all([
@@ -125,7 +125,7 @@ describe('When there are haves and have-nots', () => {
     const RichieBalance = await getBalance(richieRich.address)
     await expect(
       makeTransfer(richieRich, bobbyBroke.address, RichieBalance).then((tx) =>
-        submitSignedTx(tx, [IS_IN_BLOCK])
+        submitSignedTx(tx, IS_IN_BLOCK)
       )
     ).rejects.toThrowError()
     const [newBalance, zeroBalance] = await Promise.all([
@@ -140,10 +140,10 @@ describe('When there are haves and have-nots', () => {
     const listener = jest.fn()
     listenToBalanceChanges(faucet.address, listener)
     await makeTransfer(faucet, richieRich.address, MIN_TRANSACTION).then((tx) =>
-      submitSignedTx(tx, [IS_READY])
+      submitSignedTx(tx, IS_READY)
     )
     await makeTransfer(faucet, stormyD.address, MIN_TRANSACTION).then((tx) =>
-      submitSignedTx(tx, [IS_IN_BLOCK])
+      submitSignedTx(tx, IS_IN_BLOCK)
     )
 
     expect(listener).toBeCalledWith(
@@ -159,10 +159,10 @@ describe('When there are haves and have-nots', () => {
     listenToBalanceChanges(faucet.address, listener)
     await Promise.all([
       makeTransfer(faucet, richieRich.address, MIN_TRANSACTION).then((tx) =>
-        submitSignedTx(tx, [IS_IN_BLOCK])
+        submitSignedTx(tx, IS_IN_BLOCK)
       ),
       makeTransfer(faucet, stormyD.address, MIN_TRANSACTION).then((tx) =>
-        submitSignedTx(tx, [IS_IN_BLOCK])
+        submitSignedTx(tx, IS_IN_BLOCK)
       ),
     ])
     expect(listener).toBeCalledWith(faucet.address)
--- a/src/__integrationtests__/Balance.spec.ts
+++ b/src/__integrationtests__/Balance.spec.ts
@@ -70,8 +70,9 @@ describe('when there is a dev chain with a faucet', () => {
     const funny = jest.fn()
     listenToBalanceChanges(ident.address, funny)
     const balanceBefore = await getBalance(faucet.address)
-    const tx = await makeTransfer(faucet, ident.address, MIN_TRANSACTION)
-    await submitSignedTx(tx, AWAIT_IN_BLOCK)
+    await makeTransfer(faucet, ident.address, MIN_TRANSACTION).then((tx) =>
+      submitSignedTx(tx, AWAIT_IN_BLOCK)
+    )
     const [balanceAfter, balanceIdent] = await Promise.all([
       getBalance(faucet.address),
       getBalance(ident.address),
--- a/src/__integrationtests__/Balance.spec.ts
+++ b/src/__integrationtests__/Balance.spec.ts
@@ -14,7 +14,7 @@ import {
   AWAIT_IN_BLOCK,
   AWAIT_READY,
   IBlockchainApi,
-  submitTx,
+  submitSignedTx,
 } from '../blockchain/Blockchain'
 import getCached, { DEFAULT_WS_ADDRESS } from '../blockchainApiConnection'
 import Identity from '../identity/Identity'
@@ -71,7 +71,7 @@ describe('when there is a dev chain with a faucet', () => {
     listenToBalanceChanges(ident.address, funny)
     const balanceBefore = await getBalance(faucet.address)
     const tx = await makeTransfer(faucet, ident.address, MIN_TRANSACTION)
-    await submitTx(tx, AWAIT_IN_BLOCK)
+    await submitSignedTx(tx, AWAIT_IN_BLOCK)
     const [balanceAfter, balanceIdent] = await Promise.all([
       getBalance(faucet.address),
       getBalance(ident.address),
@@ -100,7 +100,7 @@ describe('When there are haves and have-nots', () => {
       richieRich,
       stormyD.address,
       MIN_TRANSACTION
-    ).then((tx) => submitTx(tx, AWAIT_IN_BLOCK))
+    ).then((tx) => submitSignedTx(tx, AWAIT_IN_BLOCK))
     const balanceTo = await getBalance(stormyD.address)
     expect(balanceTo.toNumber()).toBe(MIN_TRANSACTION.toNumber())
   }, 40_000)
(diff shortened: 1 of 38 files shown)


src/blockchain/Blockchain.ts
```diff
@@ -80,46 +88,53 @@ export default class Blockchain implements IBlockchainApi {
...
-              )
-            )
-          }
+      tx.send((result) => {
+        log.info(`Got tx status '${result.status.type}'`)
+        if (ErrorHandler.extrinsicFailed(result)) {
+          log.warn(`Extrinsic execution failed`)
+          log.debug(`Transaction detail: ${JSON.stringify(result, null, 2)}`)
+          const extrinsicError: ExtrinsicError =
+            this.errorHandler.getExtrinsicError(result) || ERROR_UNKNOWN
```
rflechtner, line 117: I would love to turn this into a static function, but the error handler is in the way. Any suggestions whether and how we could get rid of this?

rflechtner: see 0a4a8127a18d75fa29ab6efdf680569af34d75fa 👏

src/blockchain/Blockchain.ts
```diff
@@ -27,19 +27,70 @@ export type Stats = {
...
+        reject(extrinsicError)
+      } else if (result.isError) {
+        reject(
+          new Error(`Transaction failed with status '${result.status.type}'`)
+        )
+      } else if (resolveOn(result)) {
+        resolve(result)
+      }
+    })
+      .then((cb) => {
```
Dudleyneedham, line 81: can we change the name from `cb`? I know it is a callback, but it seems not verbose enough.

src/blockchain/Blockchain.ts
```diff
@@ -27,19 +27,70 @@ export type Stats = {
...
+        unsubscribe = cb
+      })
+      // not sure we need this final catch block
+      .catch((err: Error) => {
+        // just reject with the original tx error from the chain
+        reject(err)
```
Dudleyneedham, line 87: What errors are you trying to catch with this?

src/blockchain/Blockchain.ts
```diff
@@ -27,19 +27,70 @@ export type Stats = {
...
+      })
+  }).finally(() => {
+    if (unsubscribe) unsubscribe()
+  })
+}
+
```
Dudleyneedham, line 93: Love it!

src/blockchain/Blockchain.ts
```diff
@@ -80,47 +126,26 @@ export default class Blockchain implements IBlockchainApi {
...
   }
 
-  public async submitTx(
+  public async signTx(
     identity: Identity,
     tx: SubmittableExtrinsic
-  ): Promise<SubmittableResult> {
-    const accountAddress = identity.address
-    const nonce = await this.getNonce(accountAddress)
+  ): Promise<SubmittableExtrinsic> {
```
Dudleyneedham, line 132: This makes a lot more sense.

src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -38,9 +38,7 @@ export async function buildConnection(
     provider,
     types: CUSTOM_TYPES,
   })
-  const bc = new Blockchain(api)
-  await bc.ready
-  return bc
+  return new Blockchain(api)
```
Dudleyneedham, line 41: Shouldn't this be new `Blockchain(api).ready`?

Or do we not need the `ready` now since you removed it from the blockchain constructor? 

`this.ready = this.errorHandler.ready`

src/blockchain/Blockchain.ts
```diff
@@ -80,47 +126,26 @@ export default class Blockchain implements IBlockchainApi {
...
-        .catch((err: Error) => {
-          // just reject with the original tx error from the chain
-          reject(err)
-        })
-    })
+    return signed
+  }
+
+  // TODO: should this be renamed signAndSubmitTx ?
+  public async submitTx(
```
Dudleyneedham, line 142: Yea, I think this is better. 

`signAndSubmitTx`

src/blockchain/Blockchain.ts
```diff
@@ -80,47 +126,26 @@ export default class Blockchain implements IBlockchainApi {
     return this.api.rpc.chain.subscribeNewHeads(listener)
   }
 
-  public async submitTx(
+  public async signTx(
```
Dudleyneedham, line 129: how about `signableTx`

src/blockchain/Blockchain.ts
```diff
@@ -27,19 +27,70 @@ export type Stats = {
...
 }
 
+export const AWAIT_READY: txStatusPromiseResolver = (result) =>
+  result.status.isReady
+export const AWAIT_IN_BLOCK: txStatusPromiseResolver = (result) =>
+  result.isInBlock
+export const AWAIT_FINALIZED: txStatusPromiseResolver = (result) =>
+  result.isFinalized
+
+export async function submitTx(
```
Dudleyneedham, line 57: how about `submitableTx`

src/blockchain/Blockchain.ts
Dudleyneedham: This could be `signableAndSubmitableTx`

src/errorhandling/ErrorHandler.ts
```diff
@@ -29,46 +29,29 @@ export class ErrorHandler {
...
-      events.find((eventRecord: EventRecord) => {
-        return (
-          !eventRecord.phase.asApplyExtrinsic.isEmpty &&
-          eventRecord.event.index.toHex() === SystemEvent.ExtrinsicFailed
-        )
-      }) !== undefined
-    )
-  }
-
-  public constructor(apiPromise: ApiPromise) {
```
Dudleyneedham, line 42: If you removed the constructor. 

Why should this be a class now? 

Shouldn't we just have these as functions?

src/__integrationtests__/Balance.spec.ts
```diff
@@ -20,7 +25,7 @@ import {
   wannabeFaucet,
 } from './utils'
 
-let blockchain: IBlockchainApi | undefined
```
Dudleyneedham, line 23: I hate `undefined`

src/__integrationtests__/Balance.spec.ts
```diff
@@ -65,7 +70,8 @@ describe('when there is a dev chain with a faucet', () => {
     const funny = jest.fn()
     listenToBalanceChanges(ident.address, funny)
     const balanceBefore = await getBalance(faucet.address)
-    await makeTransfer(faucet, ident.address, MIN_TRANSACTION)
+    const tx = await makeTransfer(faucet, ident.address, MIN_TRANSACTION)
```
Dudleyneedham, line 73: This is better already. It is more verbose. 

Rolling this out will make things clearer when we action a `tx` to be written on the chain.

src/__integrationtests__/Balance.spec.ts
```diff
@@ -65,7 +70,8 @@ describe('when there is a dev chain with a faucet', () => {
...
+    await submitTx(tx, AWAIT_IN_BLOCK)
```
Dudleyneedham, line 74: You have gone down a different route with the following `makeTransfer` see line 99.

Is the `const = tx` needed. Else, it would be better if you make it to a `then()`.

src/__integrationtests__/Balance.spec.ts
```diff
@@ -111,7 +123,9 @@ describe('When there are haves and have-nots', () => {
   it('should not accept transactions when sender cannot pay gas, but will keep gas fee', async () => {
     const RichieBalance = await getBalance(richieRich.address)
```
Dudleyneedham, line 124: Never saw the names before. But, its great XD.

src/__integrationtests__/ErrorHandler.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+  const tx = await attestation.store(attester)
+  await expect(submitTx(tx, AWAIT_IN_BLOCK)).rejects.toThrow(
+    ERROR_CTYPE_NOT_FOUND
+  )
+}, 40_000)
+
+afterAll(() => {
+  if (typeof blockchain !== 'undefined') blockchain.api.disconnect()
+})
```
Dudleyneedham, line 45: Thank you, I have been having issues with the Error handling from the chain.

Dudleyneedham (changes requested):
Looks great and should help with the `Substrate 2.0` update.

A few minor things that need altering or questions.

src/blockchain/Blockchain.ts
rflechtner: I don't know... that's old code, I added the comment `not sure we need this final catch block`

src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -38,9 +38,7 @@ export async function buildConnection(
     provider,
     types: CUSTOM_TYPES,
   })
-  const bc = new Blockchain(api)
-  await bc.ready
```
Dudleyneedham, line 42: This is required for the use of PortableGabi. 

Is there another way around?

src/blockchainApiConnection/BlockchainApiConnection.ts
rflechtner: yeah, no more .ready needed, this was just to initialise the error handler

src/blockchain/Blockchain.ts
rflechtner: in order not to break the old api, I would want to keep this name although its not super descriptive. I have renamed the other function instead to signal that it requires previous signing (unlike this one)

src/blockchain/Blockchain.ts
rflechtner: this is actually nice and descriptive: here the tx is signed

src/errorhandling/ErrorHandler.ts
rflechtner: you're right, will change

src/__integrationtests__/Balance.spec.ts
rflechtner: actually that is a change that is no longer necessary... should I revert?

src/__integrationtests__/Balance.spec.ts
rflechtner: yesss that was the point. now noone can say they didn't know it would cost tokens. In theory we should also be able to check transaction costs beforehand

src/blockchainApiConnection/BlockchainApiConnection.ts
Dudleyneedham: Nevermind.

src/__integrationtests__/Balance.spec.ts
rflechtner: it is not, it is just a matter of style. couldn't really decide what I like better

src/blockchain/Blockchain.ts
rflechtner: we actually have failing tests if that's not there, so I guess we need it

src/blockchainApiConnection/BlockchainApiConnection.ts
Dudleyneedham: Ok, good to know.

src/blockchain/Blockchain.ts
Dudleyneedham: Ok, that is fine.  I was just wondering as well. It seemed like everything was handled. 

Do you know if there is a better way to catch if no tokens are available.

Dudleyneedham: This isn't a big deal, maybe we would like to make a ticket to see if we can have a better way of handling this.

src/__integrationtests__/Balance.spec.ts
Dudleyneedham: So, you will need undefined.

src/blockchain/Blockchain.ts
rflechtner: it's just weird for me as I expected anything that throws inside of a promise to be implicitly handled and to reject the promise, but it seems we have to make it explicit here

src/blockchain/Blockchain.ts
```diff
@@ -50,19 +100,21 @@ export default class Blockchain implements IBlockchainApi {
     return []
   }
 
+  public static submitSignedTx(
```
Dudleyneedham, line 103: Great addition.

src/ctype/CType.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { SubmittableResult } from '@polkadot/api'
 import { TypeRegistry } from '@polkadot/types'
 import { Option } from '@polkadot/types/codec'
-
+import { submitSignedTx } from '../blockchain/Blockchain'
```
Dudleyneedham, line 4: You should only need the following.

`import { submitSignedTx } from '../blockchain'

Since we are accessing the index, if I am not mistaken.

src/delegation/DelegationNode.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { submitSignedTx } from '../blockchain/Blockchain'
```
Dudleyneedham, line 1: again see above. In the `CType.spec.ts`.

src/__integrationtests__/Balance.spec.ts
rflechtner: the original is indeed the better description of what's going on, if that's what you mean

src/errorhandling/ErrorHandler.spec.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
-import { ApiPromise, SubmittableResult } from '@polkadot/api'
+import { SubmittableResult } from '@polkadot/api'
 import { Tuple } from '@polkadot/types'
-import { ErrorHandler } from './ErrorHandler'
+import { ErrorHandler } from '.'
```
Dudleyneedham, line 4: I think it might be better to have it as `import { ErrorHandler } from './ErrorHandler'`, since it is more verbose.

src/errorhandling/index.ts
```diff
@@ -3,4 +3,5 @@
  * @ignore
  */
 
+export * as ErrorHandler from './ErrorHandler'
 export * from './ExtrinsicError'
```
Dudleyneedham, line 7: Shouldn't we handle it as the following `export * as ExtrinsicError from './ExtrinsicError'` too.

Dudleyneedham (changes requested):
Nice job, just some cleaning up.

As a rule I think we should handle the `from '../blockchain/Blockchain'` as the following since we are using the `Index.ts` as  `from '../blockchain'`

Dudleyneedham:
Like the changes from a class factory to the function. Let's see if the team approves.

src/ctype/CType.spec.ts
rflechtner: true, I recently added this to the index. Wasn't actually sure what to export from there, but this seems an essential function

src/errorhandling/ErrorHandler.spec.ts
rflechtner: that won't work - we only export as ErrorHandler from the index

src/errorhandling/index.ts
rflechtner: mhhh I that's code I don't need to mess with, so I'd refrain from doing so and possibly break stuff

Dudleyneedham (approved):
LGTM

src/did/Did.spec.ts
```diff
@@ -59,19 +60,18 @@ describe('DID', () => {
...
 
-  it('query by identifier invalid identifier', async (done) => {
-    try {
-      await Did.queryByIdentifier('invalidIdentifier')
-      done.fail('should have detected an invalid DID')
-    } catch (err) {
-      done()
-    }
+  it('query by identifier invalid identifier', async () => {
+    await expect(Did.queryByIdentifier('invalidIdentifier')).rejects.toThrow()
```
wischli, line 64: Can we specify the error?

wischli (approved):
LGTM! We should discuss whether we want to implement this in Portablegabi but I would not put too much effort into that, thus keep Portablegabi as is.

src/did/Did.spec.ts
rflechtner: we can !

src/errorhandling/ErrorHandler.ts
```diff
@@ -17,90 +17,51 @@ export enum SystemEvent {
...
-  public static extrinsicFailed(extrinsicResult: SubmittableResult): boolean {
-    const events: EventRecord[] = extrinsicResult.events || []
+/**
+ * Checks if there is `SystemEvent.ExtrinsicFailed` in the list of
+ * transaction events within the given `extrinsicResult`.
+ *
+ * @param extrinsicResult The result of a submission.
+ * @returns Whether the extrinsic submission failed.
+ */
+export function extrinsicFailed(extrinsicResult: SubmittableResult): boolean {
```
tjwelde, line 29: I think you will have to explain to me, why the errorhandler shrinked

tjwelde (commented):
I like the API overall. My only concern would be the inability to wait for both `IN_BLOCK` and `FINALIZED`. Would you recommend making your own `resolveOn` handler in that case?

src/errorhandling/ErrorHandler.ts
wischli: > My only concern would be the inability to wait for both `IN_BLOCK` and `FINALIZED`. 

Doesn't `FINALIZED` already imply `IN_BLOCK`?

rflechtner: I think we are talking about individual promises that resolve at different stages. Should be possible, I'm working on something right now

Dudleyneedham (changes requested):
`export { SubmittableExtrinsic } from '@polkadot/api/promise/types'`
 needs to be added to the `Index.ts` .

rflechtner:
> Maybe, to solve the problem with only being able to wait for one status change right now, we can give the user the possibility to add a subscription handler to `submitSignedTx` or have another function for that and share some code between the new one and `submitSignedTx`. Could also help as preparation for the subscription handler we want to do in the future...

Boy do I have something for you :)
To redirect a subscription to a promise, you can now use a SubscriptionPromise factory:
```ts
  const { promise, subscription } = makeSubscriptionPromise(
   [IN_BLOCK], // when to resolve
   [EXTRINSIC_FAILED, IS_ERROR], // when to reject
   30_000 // timeout in ms
  )
  const unsubscribe = await tx.send(subscription)
  await promise
  unsubscribe()
```
There is even a factory that makes it easy to subscript multiple promises:
```ts
  const { promises, subscription } = makeSubscriptionPromiseMulti([
  {
   resolveOn: [IN_BLOCK],
   rejectOn: [EXTRINSIC_FAILED, IS_ERROR], 
  }, {
   resolveOn: [IS_FINALIZED],
   rejectOn: []
   timeout: 60_000
  }]
  )
  const unsubscribe = await tx.send(subscription)
  await promises[0]
  await promises[1]
  unsubscribe()
```
The only thing that's different now in a txHandler is that you have to create the Promises first here, then subscribe them to transaction status updates.

Dudleyneedham (approved):
LGTM.

src/util/SubscriptionPromise.ts
```diff
@@ -0,0 +1,63 @@
+import { ERROR_TIMEOUT } from '../errorhandling/SDKErrors'
+
+export interface Evaluator<I, O> {
+  (value: I): [boolean, O]
+}
+
+export function makeSubscriptionPromise<SubscriptionType, PromiseType>(
+  resolveOn: Array<Evaluator<SubscriptionType, PromiseType>>,
```
tjwelde, line 8: Let's just have one Evaluator (which internally can combine different conditions)

tjwelde (changes requested):

src/util/SubscriptionPromise.ts
rflechtner: I have simplified the evaluator, it's now way easier to just chain multiple evaluators to produce a combined one. You'll like it!

src/util/SubscriptionPromise.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * A function that determines whether a new incoming value should reject or resolve the promise.
+ *
+ * @param value A new incoming subscription value.
+ * @returns A truthy value if the promise should be rejected/resolved on this value. Return value is used
+ * as the reason in a rejected promise.
+ */
+export interface Evaluator<SubscriptionType> {
+  (value: SubscriptionType): any
```
tjwelde, line 11: What do you think about
```suggestion
  (value: SubscriptionType): Boolean | any
```

rflechtner: was thinking about that... it's a nice hint but would not have any effects. could do though

tjwelde (approved):
LGTM

PR #301, opened 2020-09-30
Unit denomination and conversion

LeonFLK:
## fixes KILTProtocol/ticket#632

Adds `Balance.utils.ts` to the sdk to handle unit denomination and conversion in a central place instead of our separate components.

## How to test:

Create local instance, start demo client and services, add faucet and transfer funds to test if the portrayal of units is correct.

## Checklist:

### DIFF
--- a/src/__integrationtests__/utils.ts
+++ b/src/__integrationtests__/utils.ts
@@ -9,9 +9,6 @@ import CType from '../ctype/CType'
 import { getOwner } from '../ctype/CType.chain'
 import Identity from '../identity/Identity'
 
-// FIXME: check with weights
-// export const GAS = new BN(1_000_000)
-export const GAS = new BN(125_000_000)
 export const MIN_TRANSACTION = new BN(100_000_000)
 export const ENDOWMENT = MIN_TRANSACTION.mul(new BN(100))
 
--- a/src/balance/Balance.chain.ts
+++ b/src/balance/Balance.chain.ts
@@ -15,6 +15,7 @@ import BN from 'bn.js'
 import { getCached } from '../blockchainApiConnection'
 import Identity from '../identity/Identity'
 import IPublicIdentity from '../types/PublicIdentity'
+import BalanceUtils from './Balance.utils'
 
 /**
  * Fetches the current balance of the account with [accountAddress].
@@ -93,7 +94,8 @@ export async function listenToBalanceChanges(
  *
  * @param identity Identity to use for token transfer.
  * @param accountAddressTo Address of the receiver account.
- * @param amount Amount of Femto-Kilt (1e-15) to transfer.
+ * @param amount Amount of Units to transfer.
+ * @param exponent Magnitude of the amount.
  * @returns Promise containing the transaction status.
  *
  * @example
@@ -117,9 +119,16 @@ export async function listenToBalanceChanges(
 export async function makeTransfer(
   identity: Identity,
   accountAddressTo: IPublicIdentity['address'],
-  amount: BN
+  amount: BN,
+  exponent = -15
 ): Promise<SubmittableResult> {
   const blockchain = await getCached()
-  const transfer = blockchain.api.tx.balances.transfer(accountAddressTo, amount)
+  const transfer = blockchain.api.tx.balances.transfer(
+    accountAddressTo,
+    BalanceUtils.convertToTxUnit(
+      amount,
+      (exponent >= 0 ? 1 : -1) * Math.floor(Math.abs(exponent))
+    )
+  )
   return blockchain.submitTx(identity, transfer)
 }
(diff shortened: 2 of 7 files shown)


src/__integrationtests__/utils.ts
```diff
@@ -9,9 +9,6 @@ import CType from '../ctype/CType'
 import { getOwner } from '../ctype/CType.chain'
 import Identity from '../identity/Identity'
 
-// FIXME: check with weights
-// export const GAS = new BN(1_000_000)
-export const GAS = new BN(125_000_000)
```
Dudleyneedham, line 14: With the new Substrate two update, we may be able to handle the weights better. 

https://github.com/paritytech/substrate/pull/6789

At the moment, I am not sure what will do about the whole weights and transaction costs on the demo-client.

src/balance/Balance.chain.ts
```diff
@@ -117,9 +119,17 @@ export async function listenToBalanceChanges(
...
   const blockchain = await getCached()
-  const transfer = blockchain.api.tx.balances.transfer(accountAddressTo, amount)
+  const cleanExponent =
+    (exponent >= 0 ? 1 : -1) * Math.floor(Math.abs(exponent))
+  const transfer = blockchain.api.tx.balances.transfer(
+    accountAddressTo,
+    cleanExponent === -15
+      ? amount
+      : BalanceUtils.convertToTxUnit(amount, cleanExponent)
+  )
```
Dudleyneedham, line 133: This might be worth waiting for the changes that the other PR will come out with some much needed refactoring of how we handle `tx`s.

Dudleyneedham: It looks very good though.

src/balance/Balance.spec.ts
```diff
@@ -65,4 +66,21 @@ describe('Balance', () => {
     expect(status).toBeInstanceOf(SubmittableResult)
     expect(status.isFinalized).toBeTruthy()
   })
+  it('should make transfer of amount with arbitrary exponent', async () => {
+    const alice = await Identity.buildFromURI('//Alice')
```
Dudleyneedham, line 70: you should check out the conventional way we are doing the tests.

Having a let at the top. Would be better if we keep things in line with all the other tests.

src/balance/Balance.utils.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(new BN(convertToTxUnit(new BN(1), 12).toString())).toEqual(
+      new BN('1000000000000000000000000000')
+    )
+    expect(new BN(convertToTxUnit(new BN(1), 15).toString())).toEqual(
+      new BN('1000000000000000000000000000000')
+    )
+    expect(new BN(convertToTxUnit(new BN(1), 18).toString())).toEqual(
+      new BN('1000000000000000000000000000000000')
+    )
+  })
```
Dudleyneedham, line 81: I bet that took a long time.

src/balance/Balance.utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const TRANSACTION_FEE = convertToTxUnit(new BN(125), -9)
+
+export default {
+  KILT_COIN,
+  TRANSACTION_FEE,
+  formatKiltBalance,
+  asFemtoKilt,
+  convertToTxUnit,
+}
```
Dudleyneedham, line 36: Much needed for the demo-client.

Dudleyneedham: You will need to add this to the Index file of the `SDK`.

Or is that already being done by the `src/balance/index.ts`?

Dudleyneedham (changes requested):
LGTM

src/__integrationtests__/utils.ts
LeonFLK: Okay lets remove this for now, and maybe read later when we can play with the weights.

src/balance/Balance.utils.spec.ts
LeonFLK: copy and paste, friendoo

src/balance/Balance.utils.ts
LeonFLK: Yes it's included in Balance module, we could separate it I guess.

src/balance/index.ts
```diff
@@ -3,4 +3,8 @@
  * @ignore
  */
 
-export * from './Balance.chain'
+import BalanceUtils from './Balance.utils'
+import * as Balance from './Balance.chain'
+
+export { Balance, BalanceUtils }
+export default Balance
```
Dudleyneedham, line 10: Better 👍

src/index.ts
```diff
@@ -66,6 +66,7 @@ export {
   IBlockchainApi,
   BlockchainApiConnection,
   Balance,
+  BalanceUtils,
```
Dudleyneedham, line 69: Yes, that is what I meant.

Dudleyneedham (approved):
LGTM!

PR #303, opened 2020-10-20
fix claim error check

rflechtner:
## fixes KILTProtocol/ticket#781
Fixes Claim errorCheck function to allow falsy values in Claim contents like `0`, `false`, `''`.
Improves readability as well.

## How to test:
create Claim containing the above values.

## Checklist:


wischli (approved):
Your explanation makes sense (e.g. for empty comment section in DID etc). Thus, LGTM.

PR #304, opened 2020-10-20
Proof of Concept: Exporting KILT credentials to VC format

rflechtner:
## closes KILTProtocol/ticket#767
This is a first attempt at coming up with a data structure for KILT credentials that complies with the emerging Verifiable Credential standard. 
The proposed solution is based on multiple custom proof types that perform 3 different verification tasks:
1. Verify content hashes agains root hash and verify signature over root hash against claimer signature (self-signed proof). This proof is equivalent to integrity protection and authentication provided by the `RequestForAttestation`
2. Verify disclosed properties (actual claim contents) and "salt" (nonces) against content hashes.
3. Use root hash (=credential id) to query & verify on-chain attestation and check revocation status.

## How to test:
Some unit tests are included to demonstrate the functionality. As illustrated in these tests, you may:
1. export a credential to a vc
2. run validation functions over the resulting credential and proofs
3. generate a proof for property disclosure
4. run validation functions over the credential and generated proof

## Checklist:

### DIFF
--- a/src/util/toVerifiableCredential.spec.ts
+++ b/src/util/toVerifiableCredential.spec.ts
@@ -1,0 +1,160 @@
+import { Attestation, IClaim } from '..'
+import AttestedClaim from '../attestedclaim'
+import CType from '../ctype'
+import attClaimToVC, {
+  makeRevealPropertiesProof,
+  verifyAttestedProof,
+  verifyRevealPropertyProof,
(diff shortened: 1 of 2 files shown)


rflechtner:
Closing as we have decided against this VC format. #307 is where this work is continued using a different format. 
I'd want to keep this branch though for future reference.

PR #306, opened 2020-11-12
feat: json-ld claim property hashing for selective disclosure

rflechtner:
## fixes KILTProtocol/ticket#782

This proposal introduces a new hashing scheme for selective disclosure of claim contents, as the one currently used has its flaws (see tickets).

Note that I have yet to integrate this into the RequestForAttestation (it is currently unused); at this point I merely want your feedback on the approach before moving forward.

In short, claim contents are hashed by extracting stringified statements consisting of key-value pairs and computing a salted hash over them. When selecting the format/structure for statements, I relied on json-ld's ideas for producing canonical representations of data. This is for the simple reason that this canonical representation is directly understandable in the context of Verifiable Credentials, as they are based on JSON-LD; this also means it is quite easy to generate the canonical hashing representation from a VC built from a KILT claim as long as it contains claim properties in a JSON-LD compatible format. We thereby greatly ease our task to extend VC support.    
A core feature of this representation is describing the relationship of subject (claimer) and arguments (values of claim content properties) via globally unique strings (URIs/IRIs). Because the claim properties in KILT get their semantics from the CType associated with a claim, these URIs are composed here by prepending the property value with the CType schema Id.

Also see [this doc](https://docs.google.com/document/d/1gtkoZUHb7TFXP6nNiW0dZ6CcMLROweYe0g8JWx7AZ7U) for more detailed descriptions of the solutions and rationales behind them.

I have been debating several features of the solution presented here with myself quite a bit; most notably these:
- JSON-LD proofs generally choose an n-quads normalisation of data before signing and hashing, where statements look like this: **<_subject id_> <_predicate URI_> <_value (argument)_> <_data type_> .** To produce this format, I would need to rely on the `jsonld.js` library which I have tried to avoid for now. This is for the simple reason of not wanting to introduce another dependency at this point. The format I use now is a bit more 'custom' but should still be readily understandable for JSON-LD processors: it consists of a simple JSON key value pair **{"_predicate URI_": "_value (argument)_"}**. Of course this format could be changed in the future; I imagine we could introduce flexibility by indicating the canonicalisation algorithm in an additional field on credentials (on the VC, this would sit on the proof; many proof types already do this). 
- To validate hashes, the validation needs to associate the nonces used to hash with the respective statements in order to produce the correct hashes. Because keys are translated, using the original keys as we currently do is not a good option. We may also want to remain open to changes in the canonicalisation, in which case we might not even have such a thing as json keys. A solution that caters for both problems could be provided by using the unsalted hash of the statement as a nonce key. This is sensitive information, of course, but so is the nonce in combination with the original hash. As long as we always remove both key and nonce from the nonce map when removing properties from a claim, we should be safe.
- Computing unsalted hashes as nonce keys gives us two options for computing the final hash: either combine statement and nonce, then hash, or combine the unsalted hash (nonce key) and nonce, then hash. The second option could help building fail-fast validation and I currently see no problems or exploits around this approach, so I pick this one.

## How to test:
There are a few unit tests included to demonstrate how the scheme operates.

## Checklist:


src/claim/Claim.utils.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    nonces[item.key] = item.nonce
+  })
+
+  it('validates hashes of full claim', () => {
+    const rehashed = hashClaimContents(claim, { nonces })
+    expect(hashes).toEqual(rehashed.map((item) => item.hash))
+  })
+
+  it('validates hashes of partial claims', () => {
+    Object.keys(claim.contents).forEach((property) => {
```
tjwelde, line 119: A bit hard to understand. I would suggest adding more comments or making and using helper functions.

tjwelde (changes requested):
Looking good so far. We will have to explain the concepts in-code a bit, so a third-party can understand what is happening.

src/claim/Claim.utils.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+      },
+    ]
+  `)
+})
+
+describe('compute hashes & validate', () => {
+  const hashed = hashClaimContents(claim)
+  const hashes = hashed.map((item) => item.hash)
+  const nonces = {}
+  hashed.reverse().forEach((item) => {
```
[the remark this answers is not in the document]
rflechtner: I tried to find it but it doesn’t seem to be on the array prototype in ES6... I could use some sort hack to do basic shuffling

src/claim/Claim.utils.spec.ts
rflechtner: just pushed a refactor that is supposed to improve readability. What do you think?

rflechtner:
@tjwelde I improved readability by splitting code into several functions and adding comments and docstrings. I also added a `nonceGenerator` option as requested. Finally I changed the output of `hashClaimContents` to produce an array of salted hashes plus a map of unsalted hashes to nonces. Hope you agree that this improves things!

src/claim/Claim.utils.spec.ts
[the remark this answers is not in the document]
tjwelde: Oh, I might have confused languages here. Don't bother than.

src/claim/Claim.utils.ts
```diff
@@ -51,52 +65,129 @@ export function toJsonLD(
...
+  const { hasher, nonces, nonceGenerator, canonicalisation } = {
+    ...defaults,
+    ...options,
+  }
+  // use canonicalisation algorithm to make hashable statement strings
+  const statements = canonicalisation(claim)
+  // generate unsalted hashes from statements as a first step
+  const unsalted = hashStatements(statements, {
+    hasher,
+    nonceGenerator: undefined,
```
tjwelde, line 109: This is the same as just leaving it out. Why did you add it explicitly?

src/requestforattestation/RequestForAttestation.spec.ts
```diff
@@ -540,58 +479,43 @@ describe('RequestForAttestation', () => {
...
-    ).toThrowError(
-      ERROR_NONCE_HASH_INVALID(
-        builtRequestMalformedCtypeHash.cTypeHash,
-        'CType'
-      )
+      RequestForAttestationUtils.errorCheck(builtRequestMalformedHashes)
     )
+      .toThrowError
+      // TODO: fix error type
+      // ERROR_NONCE_HASH_INVALID()
```
tjwelde, line 517: What needs to be fixed here?

src/claim/Claim.utils.ts
```diff
@@ -51,52 +65,129 @@ export function toJsonLD(
...
+  })
+  // to simplify validation, the salted hash is computed over unsalted hash (nonce key) & nonce
+  // the unsalted hash will work has the nonce key later, so we feed a hash:hash map to the hasher
```
tjwelde, line 112:
```suggestion
  // the unsalted hash will work as the nonce key later, so we feed a hash:hash map to the hasher
```

tjwelde (commented):
some nitpicks

tjwelde:
I really like the new output of `hashClaimContents`, btw

src/claim/Claim.utils.spec.ts
tjwelde: love it

src/claim/Claim.utils.ts
rflechtner: in case we would set a default generator, this would force it to undefined (=no nonces)

tjwelde: ah ok

src/requestforattestation/RequestForAttestation.spec.ts
rflechtner: I haven't figured out error reporting for this. It seems like the old sdk errors don't fully apply any more. We had `ERROR_CLAIM_HASHTREE_MALFORMED`, `ERROR_NONCE_HASH_INVALID`, `ERROR_NONCE_HASH_MALFORMED`, `ERROR_ROOT_HASH_UNVERIFIABLE`, `ERROR_CLAIM_UNVERIFIABLE`
Now there's either an unsubstantiated statement for which there is no matching digest in the nonce map, or there is a digest and nonce for a statement, but it they do not combine into a hash that's included on the signed credential. I probably have to create new error types for this. Maybe something like NO_PROOF_FOR_STATEMENT and INVALID_PROOF_FOR_STATEMENT ?
Also the way verification currently works I'm not even sure I can properly differentiate between the two

src/crypto/Crypto.ts
```diff
@@ -213,3 +213,62 @@ export function decryptAsymmetricAsStr(
...
+  // getNonce <- if nonces map then take from map, else use nonceGenerator function
+  // if no nonceGenerator given, unsalted hashes are returned
+  const getNonce: HashingOptions['nonceGenerator'] =
+    typeof nonces === 'object'
+      ? (key) => {
+          const nonce = nonces[key]
+          // nonce map must be exhaustive
+          // TODO: use sdk error module
+          if (!nonce)
+            throw new Error(
```
tjwelde, line 259: Can you please use our errorHandler workflow for this?

src/claim/Claim.utils.ts
```diff
@@ -4,10 +4,129 @@
...
+ */
+export type PartialClaim = Partial<IClaim> & Pick<IClaim, 'cTypeHash'>
+
+function JsonLDcontents(
+  claim: PartialClaim,
+  expanded = true
+): Record<string, AnyJson> {
+  const { cTypeHash, contents, owner } = claim
+  if (!cTypeHash)
+    throw new Error('ctype hash is required for conversion to json-ld')
```
tjwelde, line 30: please also provide an error according to error handler here

tjwelde (changes requested):

src/requestforattestation/RequestForAttestation.spec.ts
tjwelde: The high level errors should be handled in another ticket. Please make one.
Please provide the low level errors as part of this PR and remove the now unthrown low level errors.

src/claim/Claim.utils.ts
tjwelde: Here you still not use our error workflow

rflechtner: ooooupsie

src/claim/Claim.utils.ts
```diff
@@ -91,41 +88,80 @@ export function hashClaimContents(
   options: HashingOptions & {
     canonicalisation?: (claim: PartialClaim) => string[]
   } = {}
-): { hashes: string[]; nonceMap: Record<string, string> } {
+): {
+  hashes: string[]
+  nonceMap: Record<string, string>
+  statements: string[]
```
tjwelde, line 94: let's remove it again

src/claim/Claim.utils.ts
```diff
@@ -91,41 +88,80 @@ export function hashClaimContents(
...
   // apply defaults
-  const defaults = {
-    nonceGenerator: () => uuid(),
-    canonicalisation: makeStatementsJsonLD,
-  }
-  const { hasher, nonces, nonceGenerator, canonicalisation } = {
-    ...defaults,
-    ...options,
-  }
+  const { canonicalisation = makeStatementsJsonLD } = options
```
tjwelde, line 97: I prefer the old version

src/crypto/Crypto.ts
```diff
@@ -224,51 +225,46 @@ export interface HashingOptions {
...
+}> {
   // apply default hasher if undefined
-  const { hasher, nonces, nonceGenerator } = {
-    ...options,
-    hasher: options.hasher || saltedBlake2b256,
-  }
-  // getNonce <- if nonces map then take from map, else use nonceGenerator function
-  // if no nonceGenerator given, unsalted hashes are returned
+  const {
+    hasher = saltedBlake2b256,
```
tjwelde, line 253: I prefer more explicit defaults

src/claim/Claim.utils.ts
```diff
@@ -91,41 +88,80 @@ export function hashClaimContents(
...
+  proof: {
+    nonces: Record<string, string>
+    hashes: string[]
+  },
+  options: Pick<HashingOptions, 'hasher'> & {
+    canonicalisation?: (claim: PartialClaim) => string[]
+  } = {}
+): { verified: boolean; errors: SDKErrors.SDKError[] } {
+  // apply defaults
+  const { canonicalisation = makeStatementsJsonLD } = options
```
tjwelde, line 139: defaults

src/requestforattestation/RequestForAttestation.ts
```diff
@@ -288,14 +289,15 @@ export default class RequestForAttestation implements IRequestForAttestation {
...
-    }
+    // verify properties against selective disclosure proof
+    const verificationResult = ClaimUtils.verifyDisclosedAttributes(
+      input.claim,
+      {
+        nonces: input.claimNonceMap,
+        hashes: input.claimHashes,
+      }
+    )
+    if (!verificationResult.verified) throw verificationResult.errors[0]
```
tjwelde, line 300: Please add TODO

tjwelde (commented):
just some nitpicks

src/crypto/Crypto.ts
rflechtner: I did a little research and this is actually the only concise vanilla js solution for the `key : undefined` problem.
We could thus either stick with it, use [`lodash.merge`](https://lodash.com/docs/4.17.15#merge) or go with something a little more ugly:
```js
  const defaults = {
    hasher : saltedBlake2b256,
    nonceGenerator : () => uuid(),
  }
  const {nonces} = options
  const hasher =  options.hasher || defaults.hasher
  const nonceGenerator = options.nonceGenerator || defaults.nonceGenerator
```

rflechtner: (of course this would also work:
```js
  const {nonces} = options
  const hasher =  options.hasher || saltedBlake2b256
  const nonceGenerator = options.nonceGenerator || () => uuid()
```
but I wouldn't say this improves readability)

tjwelde: I like the first proposal from you, where I added the like

rflechtner: I think I have another favourite option: we purge the `options`:
```js
  // apply defaults
  Object.keys(options).forEach((key) => {
    if (typeof options[key] === 'undefined') delete options[key]
  })
  const defaults = {
    hasher: saltedBlake2b256,
    nonceGenerator: () => uuid(),
  }
  const { nonces, nonceGenerator, hasher } = { ...defaults, ...options }
```

tjwelde: I think it is a bit overkill, just for applying defaults.

LeonFLK: I like the proposal too with the likes, though I don't see how it's clearer than:

```js
const {hasher : saltedBlake2b256,
    nonceGenerator : () => uuid(),
    } = { 
    hasher: undefined,
    nonceGenerator: undefined
    }

return {hasher, nonceGenerator}
```

tjwelde (approved):
LGTM

PR #307, opened 2020-11-20
feat: VC export package

rflechtner:
## fixes KILTProtocol/ticket#767
This branch is meant to integrate changes from #304 & #306 to quickly make it available for use by the DENA.
I rebased #304 on top of #306 and changed one core principle in how claims are represented on the VC: Here, we no longer have claim hashes on the `credentialSubject` of the VC, but the actual values. Salted hashes are contained within an array on the credential body instead. Unsalted hashes mapped to nonces are contained within the proof.

## How to test:
As before (#304) unit tests are included.

## Checklist:


rflechtner:
This is now nearing completion. I have split up exporting, verification and presentation related functionality in separate files. Only those that consume the VC require `jsonld`, while the exporting does not. We may find a way to remove the mandatory `jsonld` dependency, especially when we are splitting the repo up in sub-packages.
Another change I could imagine relates to the proofs. We could potentially integrate the selective disclosure and claimer signature proof, so that things like the hash array do not need to be on the credential body. In this case we'd only have the full digest on the VC body in the form of its `@id` property.

rflechtner:
This lives in a separate package now, which removes the dependency on `jsonld` in core

packages/vc-export/src/types.ts
```diff
@@ -0,0 +1,100 @@
...
+export const KILT_ATTESTED_PROOF_TYPE = 'KILTAttestation2020'
+export const KILT_REVEAL_PROPERTY_TYPE = 'KILTRevealProperties2020'
+
+type ArrayElement<A> = A extends ReadonlyArray<infer T> ? T : never
+type IDidDocumentPublicKey = ArrayElement<IDidDocument['publicKey']>
+
+export type IPublicKeyRecord = Partial<IDidDocumentPublicKey> &
+  Pick<IDidDocumentPublicKey, 'publicKeyHex' | 'type'>
+
+export interface selfSignedProof extends Proof {
```
tjwelde, line 35: I think all interfaces should start with a capital letter

packages/vc-export/src/exportToVerifiableCredential.ts
```diff
@@ -0,0 +1,106 @@
...
+  // add attestation proof
+  const attester = input.attestation.owner
+  VC.proof.push({
+    type: KILT_ATTESTED_PROOF_TYPE,
+    attesterAddress: attester,
+  } as attestedProof)
+
+  // add hashed properties proof
+  VC.proof.push({
+    type: KILT_REVEAL_PROPERTY_TYPE,
```
tjwelde, line 99: As discussed: Let's make this a `Digest` proof, which also includes the `claimHashes`.
This would just verify the `id`, whereas our other proofs just sign this `id`

tjwelde (commented):
We went until `presentationUtils`

packages/vc-export/src/verificationUtils.ts
```diff
@@ -0,0 +1,253 @@
...
+      onChain.delegationId !== delegationId
+    )
+      throw new Error(
+        `proof not matching on-chain data: proof ${{
+          attester: attesterAddress,
+          delegation: delegationId,
+        }}`
+      )
+    // if proof data is valid but attestation is flagged as revoked, credential is no longer valid
+    if (onChain.revoked) throw new Error('attestation revoked')
```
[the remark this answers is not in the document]
tjwelde: As discussed, let's add a status field, which shows the statuses: 'invalid', 'valid', 'revoked'

packages/vc-export/src/verificationUtils.ts
```diff
@@ -0,0 +1,253 @@
...
+): Promise<VerificationResult> {
+  const {
+    hasher = (value, nonce?) => blake2AsHex((nonce || '') + value, 256),
+  } = options
+  const result: VerificationResult = { verified: true, errors: [] }
+  try {
+    // check proof
+    if (proof.type !== KILT_CREDENTIAL_DIGEST_PROOF_TYPE)
+      throw new Error('Proof type mismatch')
+    if (typeof proof.nonces !== 'object') {
```
tjwelde, line 164: proof also has to contain claimHashes now

tjwelde (changes requested):

packages/vc-export/src/exportToVerifiableCredential.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    const presentation = await claimerUtils.makePresentation(VC, ['name'])
+    const { contents, owner } = credential.request.claim
+    expect(presentation).toHaveProperty(
+      'verifiableCredential.credentialSubject',
+      {
+        '@context': expect.any(Object),
+        '@id': owner,
+        name: contents.name,
+      }
+    )
```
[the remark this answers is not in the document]
rflechtner: we have an exact and recursive matcher before, so this is covered

rflechtner:
I assume changes on the node broke integration tests (also fails on develop rn with the latest mash net-node). They pass here with an older node version

packages/vc-export/src/verificationUtils.ts
```diff
@@ -113,6 +128,7 @@ export async function verifyAttestedProof(
     // query on-chain data by credential id (= claim root hash)
     const onChain = await Attestation.query(claimHash)
     // if not found, credential has not been attested, proof is invalid
+    status = AttestationStatus.invalid
```
tjwelde, line 131: Please move into `ifs` for better understandability

tjwelde (changes requested):
Last little thing

tjwelde (approved):
LGTM

PR #308, opened 2020-11-26
feat: upgrade yarn to v2

tjwelde:
## fixes KILTProtocol/ticket#834
This PR upgrade our repo to yarn v2

## How to test:
- Update global yarn to latest version (global still uses a v1 version, v2 is used only locally)
- `yarn install`
- `yarn test`

## Checklist:


rflechtner:
I'm having trouble executing `yarn test:integration` - can you check that?

rflechtner:
> I'm having trouble executing `yarn test:integration` - can you check that?

I get this:
```
command not found: if
command not found: then
command not found: else
```
for a bunch of shells (sh, fish, bash, zsh...)

rflechtner (approved):
Works on my machine, package build works with demo client, tests are running.
Adding new packages is as easy as before. Love it!

We may have to update the integrations tests script thought

weichweich (approved):
Seems ok

rflechtner:
just one last thing: is the way deps are currently cached (locally, in the repo) the best we can do? No way to link this to yarn's global cache?

tjwelde:
yarn will redo the cache after first install

PR #309, opened 2020-12-02
fix github packages publishing workflow

rflechtner:
## NO TICKET
Fixes the publishing of preliminary packages to GitHub packages.

## How to test:
I will trigger a test run here

## Checklist:

### DIFF
--- a/.github/workflows/devpackpublish.yml
+++ b/.github/workflows/devpackpublish.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - develop
-      - fix-ghp-release
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/devpackpublish.yml
```diff
@@ -14,7 +14,7 @@ jobs:
           registry-url: https://npm.pkg.github.com
           scope: '@kiltprotocol'
       - run: yarn install --immutable
-      - run: yarn version --no-git-tag-version --prerelease --preid $(git rev-parse --short HEAD)
+      - run: npm version --no-git-tag-version prerelease --preid $(git rev-parse --short HEAD)
```
rflechtner, line 17: There is a way to do this with yarn v2 but it involves some bash magic:
```suggestion
      - run: yarn version patch
      - run: PACKAGE_VERSION=$(cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g')
      - run: yarn version $PACKAGE_VERSION-$(git rev-parse --short HEAD)
```

tjwelde:
we decided to overhaul our versioning workflow, so this PR is not relevant anymore.
See https://github.com/KILTprotocol/ticket/issues/851

PR #311, opened 2020-12-02
Update CI script logic for yarn custom shell

LeonFLK:
## fixes KILTProtocol/ticket#854

yarn 2 berry uses a new custom shell to provide consistent behaviour across systems. This breaks the test:integration script, that checks whether local chain is running on port 9944.
if -> then -> is written {condition && command} || exit

Also updates to latest berry version of 2.4.0

## Checklist:


LeonFLK:
Not sure if updating to 2.4.0 is neccessary.

tjwelde:
updating yarn to 2.4.0 is not necessary, but also doesn't hurt, so let's keep it

package.json
```diff
@@ -8,7 +8,7 @@
     "test": "jest --coverage --group=-integration",
     "test:ci": "jest --ci --coverage --group=-integration",
     "test:integration:run": "jest --group=integration -b -w 3 --silent",
-    "test:integration": "if lsof -i :9944 > /dev/null; then yarn test:integration:run; else echo 'Can not connect to chain. Is it running?'; exit 1;fi",
+    "test:integration": "{lsof -i :9944 > /dev/null && yarn test:integration:run;} || echo 'Can not connect to chain. Is it running?'; exit 1;fi",
```
tjwelde, line 11: since you got rid of `if`, you can also remove `fi`

tjwelde: Also, can you add a space after `{` and before `}`? It would make it more standard conform

tjwelde (changes requested):

package.json
LeonFLK: whoopsie

LeonFLK: and yes ofc

PR #312, opened 2020-12-03
feat: configuration service for LogLevel and chain ws address

LeonFLK:
## fixes KILTProtocol/ticket#774

This merges and replaces ConfigLog.ts.
We do not want to connect implicitly to the default ws address, instead set it directly or via Kilt.config to set the address of our desired node. This address can be changed any time.
LogLevel can still be set by ENV.
This non-strongly typed approach uses generic set and get to manipulate and access configOpts configuration object.
No specific key suggestion anymore since configOpts is union type with generic keyed object.
Throws error on unset generic property key when accessing or on retrieval of unset address.

## Checklist:


tjwelde:
I think from the API it would be best, if I could call `Kilt.config({...})`.

docs/getting-started.ts
```diff
@@ -34,7 +36,8 @@ async function main(): Promise<void> {
   })
 
   /* To store the CTYPE on the blockchain, you have to call: */
-  Kilt.connect(DEFAULT_WS_ADDRESS)
+  Kilt.config({ host: NODE_URL })
+  Kilt.connect(ConfigService.getNodeAddress())
```
rflechtner, line 40: isn't the point of the config service that the endpoint would be fetched implicitly ? I'd move `ConfigService.getNodeAddress()` into `Kilt.connect()`

src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -16,7 +17,7 @@ import { wannabeFaucet } from './utils'
 
 let blockchain: IBlockchainApi | undefined
 beforeAll(async () => {
-  blockchain = await getCached(DEFAULT_WS_ADDRESS)
+  blockchain = await getCached(getNodeAddress())
```
rflechtner, line 20: same here, I'd expect this to be done implicitly - and error if the config wasn't set

src/kilt/Kilt.ts
```diff
@@ -22,7 +24,16 @@ export async function disconnect(host: string): Promise<void> {
...
 }
 
+export function config(configuration: {
+  host?: string
+  logLevel?: LogLevel
+}): void {
+  const { host, logLevel } = configuration
+  setConfiguration(host, logLevel)
+}
+
```
rflechtner, line 34: what do you think of making this more versatile for the future by allowing arbitrary fields? We could store a `Map` of config options and retrieve them with a general `get()`. Getting the default host would thus look like `Kilt.config.get('host')`, setting it could be something like `Kilt.config.set({host: 'ws://....'})`
We could still maintain an interface that tells users which fields we currently use, but that way we can much easier extend it. You could set it up as a type generic that takes a type which defines known config options.

rflechtner: additionally you could define something like `getOrError` which throws a general error (like `CONFIG_NOT_SET`) when users try to rely on a config option but forgot to set it.

src/config/ConfigService.ts
```diff
@@ -14,24 +14,20 @@ import {
...
-  )
-)
-// Create a named loggerfactory and pass in the options and export the factory.
-// Named is since version 0.2.+ (it's recommended for future usage)
-// eslint-disable-next-line import/prefer-default-export
-export const factory = LFService.createNamedLoggerFactory(
-  'LoggerFactory',
-  options
-)
+const { DEFAULT_WS_ADDRESS } = process.env
```
rflechtner, line 19: I remember Timo saying that we want to remove endpoint configuration by environment variables. I'd keep this option in integration tests, but this could be solved by calling `Kilt.config()` in each integration test file (or in a file they all import, like `utils`

rflechtner (changes requested):
I'd prefer a config service that does not rely on a fixed set of fields to be able to extend easily in the future. What would you say, @tjwelde ?
I also remember we talked about config via environment variables, what's your opinion on this now?

LeonFLK:
Now works much more under the hood.
Use Kilt.config() to pass partial configOpts (address and logLevel).
If no host address is set, getting the address will result in thrown error.
getCached and similar all use configuration.host get for retrieval of set address.

src/kilt/Kilt.ts
```diff
@@ -9,20 +9,32 @@
...
+): Promise<void> {
   const cached = await getCached(host)
   cached.api.disconnect()
   clearCache()
 }
 
+export function config(configs: Partial<configOpts>): void {
+  const { address, logLevel } = configs
+  if (address) configuration.host = address
+  if (logLevel) configuration.logging = logLevel
```
rflechtner, line 33: This part illustrates that we'll commit to unnecessary maintenance effort if we hard-code properties into the config service and then also start translating their names. Do we want to add a line here for every new config option?

rflechtner (commented):
As I had mentioned when we talked about it, I am not a fan of hard-coding specific configuration options into the config service. I predict this will be a nightmare for maintaining our codebase later. Also it doesn't work well with the project to split the SDK into smaller sub-packages. Are we sure we want to go down that road ?

To make for some more constructive feedback, I had imagined something like this:
```typescript
let configuration: configOpts = {'logLevel': DEFAULT_DEBUG_LEVEL}

export function get<K extends keyof configOpts>(configOpt: K): configOpts[K] {
  if (typeof configuration[configOpt] === 'undefined') 
  throw new Error(`GENERIC NOT CONFIGURED ERROR FOR ${configOpt}`)
    return configuration[configOpt]
}

export function set(opts: Partial<configOpts>): void {
  if(typeof opts['logLevel'] !== 'undefined') {
    opts.logLevel = modifyLogLevel(opts.logLevel)
  }
  configuration = {...configuration, ...opts}
}
```
which could be exported as `ConfigService` or `Kilt.config` and would allow users to set options with (for example) `Kilt.config.set({defaultNodeAddress: 'value'})` and get values with `Kilt.config.get('defaultNodeAddress')`.

packages/core/src/config/ConfigService.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    it('returns logLevel property', () => {
+      ConfigService.set({ logLevel: LogLevel.Info })
+      expect(ConfigService.get('logLevel')).toEqual(LogLevel.Info)
+    })
+    it('gets custom configuration prop', () => {
+      ConfigService.set({ testProp: 'testing' })
+      expect(ConfigService.get('testProp')).toEqual('testing')
+    })
+    it('throws error if target prop does not exist or is undefined', () => {
+      expect(ConfigService.get('testinError')).toThrow(
```
tjwelde, line 73: small typo
```suggestion
      expect(ConfigService.get('testingError')).toThrow(
```

tjwelde (commented):
Just two nitpicks, otherwise I think it looks pretty good.
What do you think @rflechtner

packages/core/src/config/ConfigService.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * Config is used to configure logging.
```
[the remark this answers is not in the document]
LeonFLK: fixed, mentioned ws address and custom keys

rflechtner:
> Just two nitpicks, otherwise I think it looks pretty good.
> What do you think @rflechtner

Looks good & works for me!

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -31,7 +29,7 @@ export const CUSTOM_TYPES: RegistryTypes = {
 }
 
 export async function buildConnection(
-  host: string = DEFAULT_WS_ADDRESS
+  host: string | undefined = get('address')
```
rflechtner, line 32: What is the use case for calling `buildConnection(undefined)`?
If there is none
```suggestion
  host: string = get('address')
```
should do the trick.

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -42,7 +40,7 @@ export async function buildConnection(
 }
 
 export async function getCached(
-  host: string = DEFAULT_WS_ADDRESS
+  host: string | undefined = get('address')
```
rflechtner, line 43: same here

src/__integrationtests__/utils.ts
```diff
@@ -12,6 +12,7 @@ import Identity from '../identity/Identity'
 export const MIN_TRANSACTION = new BN(100_000_000)
 export const ENDOWMENT = MIN_TRANSACTION.mul(new BN(100))
 
+export const WS_ADDRESS = 'ws://127.0.0.1:9944'
```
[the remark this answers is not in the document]
rflechtner: @LeonFLK is there a decision regarding this issue?

src/config/ConfigService.ts
```diff
@@ -0,0 +1,89 @@
...
+  LoggerFactoryOptions,
+  LogGroupRule,
+  LogLevel,
+  getLogControl,
+  LogGroupControlSettings,
+} from 'typescript-logging'
+import { ERROR_WS_ADDRESS_NOT_SET } from '../errorhandling/SDKErrors'
+
+const DEFAULT_DEBUG_LEVEL =
+  process.env.DEBUG && process.env.DEBUG === 'true'
```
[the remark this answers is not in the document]
rflechtner: what about this?

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
LeonFLK: Ohh yes this is an artifact from before, is removed

rflechtner: just thinking: maybe it's easier to read if we import `get` as `ConfigService.get` ?

LeonFLK: For me this wouldn't change comprehension, since we already import get from ConfigService, this is unnecessary IMO.

rflechtner:
Sure, but it saves you the scrolling. I think that could help quite a bit to make it obvious what’s happening.

 

--

LeonFLK:
I mean one can always ctrl+click to get to the definition, and that specific file is quite small, so not much scrolling going on.

Just think if we were to clarify each import like that and hold it to the same standard it's going to be messy.

What do you think, make an exception for this file clarifying the import, or skip?

src/config/ConfigService.ts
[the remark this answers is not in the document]
LeonFLK: Yes as we've talked, setting debug in env is good-practice

rflechtner:
yeah importing the ConfigService as a whole would we the solution. But it's not a must, leave it if you don't think it adds much

tjwelde (approved):
Uff, finally, right? ;)
LGTM 🎉

PR #316, opened 2020-12-04
dev experience: error code jest matcher

rflechtner:
## no ticket
While working on https://github.com/KILTprotocol/sdk-js/pull/306 I realised that tests for our error code based error types can be quite painful, because the default jest matchers only compare error messages.
I learned about extending the default jest matchers with custom ones and gave it a go. Is that something we would want to use in our tests?

## How to test:
Use the matcher in unit tests as I demoed in one file.

## Checklist:


rflechtner:
have a look @LeonFLK as we talked about the issue

LeonFLK (approved):
LGTM

PR #317, opened 2020-12-04
feat: limit unneccessary gh test runs & builds

rflechtner:
## no ticket, just an improvement
I cleaned up the GitHub workflows a bit and limited builds to be triggered only on meaningful changes 

## How to test:
When we merge this, I'd expect no build to be triggered

## Checklist:


.github/workflows/tests.yml
```diff
@@ -7,19 +7,24 @@ on:
...
 jobs:
-
   check_skip:
     runs-on: ubuntu-latest
     if: "! startsWith(github.event.head_commit.message, '[CI Skip]') && github.repository == 'kiltprotocol/sdk-js'"
     steps:
-    - name: Not Skipped
-      run: echo "Not Skipped"
+      - name: Not Skipped
+        run: echo "Not Skipped"
```
rflechtner, line 28: @tjwelde can this go? this was just a test right?

.github/workflows/devpackpublish.yml
```diff
@@ -3,7 +3,13 @@ on:
...
     branches:
       - develop
-      - tw-851-git-push
+    paths:
+      - '**/*.ts'
+      - 'package.json'
+      - '!**/*.spec.ts'
+      - '!**/__integrationtests__'
+      - '!**/__mocks__'
+      - '!docs/**'
```
rflechtner, line 12: will only build if we change `.ts` files or the `package.json`, but not if the files are `.spec.ts` or in `__integrationtests__`, `__mocks__` or `docs`

rflechtner: shall I add anything else?

.github/workflows/tests.yml
```diff
@@ -7,19 +7,24 @@ on:
       - master
     tags:
       - '*'
+    paths-ignore:
+      - '**/*.md'
+      - 'LICENSE'
+      - 'Dockerfile'
+      - 'docker-compose.yml'
+      - 'dc.build.node.yml'
```
rflechtner, line 15: will not test if only markdown type files or other non-relevant files are changed

.github/workflows/devpackpublish.yml
tjwelde: No, I think it is all.

.github/workflows/tests.yml
tjwelde: It is necessary, when we don't want to repeat the `if` line in it in every job

tjwelde (approved):
LGTM

.github/workflows/tests.yml
rflechtner: 👍

PR #318, opened 2020-12-09
fix: undefined assignment on double spread

LeonFLK:
## fixes KILTProtocol/ticket#859

@rflechtner pinpointed an issue with our double spread logic to apply defaults, it might assign properties that are set to undefined.

## Checklist:


rflechtner (approved):
LGTM

rflechtner:
Users will not be able to set an undefined rejectOn, but that should be fine; they can set one that never triggers rejection

LeonFLK:
> Users will not be able to set an undefined rejectOn, but that should be fine; they can set one that never triggers rejection

I just kept your defaults and logic intact, you introduced this :)

PR #319, opened 2020-12-16
feat: convert to monorepo

tjwelde:
## fixes KILTProtocol/ticket#835
This PR converts the repo into a monorepo.

## How to test:
- Lint
- Style checks
- Unit and Integration tests

## Checklist:


rflechtner:
how about we add a template somewhere for new packages, which has the right folder structure and essential config files (`package.json`, tsconfig, etc) ?

tjwelde:
I think having the two packages is enough for now

rflechtner (approved):
LGTM!

PR #321, opened 2021-01-07
chore: compatablility with KILT parachain

weichweich:
## fixes KILTProtocol/ticket#886

Companion PR to https://github.com/KILTprotocol/kilt-parachain/pull/5

 This depends on @rflechtner changes in #322. 

## Todo:

- [ ] integration tests for a parachain?

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/Balance.spec.ts
+++ b/packages/core/src/__integrationtests__/Balance.spec.ts
@@ -27,7 +27,7 @@ import {
 
 let blockchain: IBlockchainApi | undefined
 beforeAll(async () => {
-  blockchain = await getCached(DEFAULT_WS_ADDRESS)
+  blockchain = await getCached(DEFAULT_WS_ADDRESS, true)
 })
 
 describe('when there is a dev chain with a faucet', () => {
--- a/packages/core/src/__integrationtests__/Balance.spec.ts
+++ b/packages/core/src/__integrationtests__/Balance.spec.ts
@@ -80,7 +80,7 @@ describe('when there is a dev chain with a faucet', () => {
     expect(balanceBefore.sub(balanceAfter).gt(MIN_TRANSACTION)).toBeTruthy()
     expect(balanceIdent.toNumber()).toBe(MIN_TRANSACTION.toNumber())
     expect(funny).toBeCalled()
-  }, 30_000)
+  }, 60_000)
 })
 
 describe('When there are haves and have-nots', () => {
@@ -140,7 +140,7 @@ describe('When there are haves and have-nots', () => {
     ])
     expect(zeroBalance.toString()).toEqual('0')
     expect(newBalance.lt(RichieBalance))
-  }, 30_000)
+  }, 60_000)
 
   it('should be able to make a new transaction once the last is ready', async () => {
     const listener = jest.fn()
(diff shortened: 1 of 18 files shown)


weichweich:
Contains changes from https://github.com/KILTprotocol/sdk-js/pull/322

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -9,43 +9,38 @@
...
 
 import { ApiPromise, WsProvider } from '@polkadot/api'
-import { RegistryTypes } from '@polkadot/types/types'
 import Blockchain from '../blockchain/Blockchain'
+import MASHNET_TYPES from './types/mashnet'
+import PARACHAIN_TYPES from './types/parachain'
 
 export const DEFAULT_WS_ADDRESS =
-  process.env.DEFAULT_WS_ADDRESS || 'ws://127.0.0.1:9944'
+  process.env.DEFAULT_WS_ADDRESS || 'ws://127.0.0.1:9977'
```
tjwelde, line 17: why this change?

packages/core/src/identity/Identity.ts
```diff
@@ -123,7 +123,7 @@ export default class Identity {
    * ```
    */
   public static buildFromSeed(seed: Uint8Array): Identity {
-    const keyring = new Keyring({ type: 'ed25519' })
+    const keyring = new Keyring({ type: 'sr25519' })
```
tjwelde, line 126: That is a bigger change, we should maybe do separately, or is this needed for parachains?

packages/core/src/__integrationtests__/Balance.spec.ts
```diff
@@ -80,7 +80,7 @@ describe('when there is a dev chain with a faucet', () => {
     expect(balanceBefore.sub(balanceAfter).gt(MIN_TRANSACTION)).toBeTruthy()
     expect(balanceIdent.toNumber()).toBe(MIN_TRANSACTION.toNumber())
     expect(funny).toBeCalled()
-  }, 30_000)
+  }, 60_000)
```
tjwelde, line 83: Why more?

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
weichweich: Parachains have two ws ports. 9944 is for the relay chain, 9977 is for the parachain.

We connect to the parachain...

packages/core/src/identity/Identity.ts
weichweich: This is only needed to have balances on the parachain since the endowed accounts are sr25519 accounts. Would do that in this PR either, but this was needed to test with the parachain.

packages/core/src/__integrationtests__/Balance.spec.ts
weichweich: Tests will take more time with the parachain. Blocktime is 12s not 6s. Also the blocktime is not constant. It takes longer sometimes.

weichweich:
Parachain types might be included in the latest polkadot-js. The separate types might not be needed anymore.

tjwelde:
#322 is now merged, so this PR can just concentrate on the compatibility to the kilt parachain

weichweich:
this can be closed. The most important change was the update the rest is out dated and needs to be checked again...

PR #322, opened 2021-01-08
feat: remove anonymous credentials from core & actors-api package

rflechtner:
## fixes KILTProtocol/ticket#840

Removes portablegabi features from core package including the actors features.
Also puts actors api in a separate package as it was easier to do this in one go.
This also moves the examples, which now live in the top level package and import the sdk-js package (which includes both core and actors).

## How to test:
Unit & integration tests plus examples. Because of incompatibilities downstream of the current SDK to the newest node, both integration tests and examples are not running atm. 

## Checklist:


packages/actors_api/src/credential/Credential.ts
```diff
@@ -11,15 +11,18 @@
  * @preferred
  */
```
tjwelde, line 13: This could all live in `AttestedClaim`, or `AttestedClaim` in core could be renamed to `Credential` and contain the methods here.
We should also keep the concept of making a Presentation.
Let's have a follow-up ticket here.

packages/actors_api/src/actors/Verifier.ts
```diff
@@ -185,96 +168,32 @@ async function verifyPublicPresentation(
...
-    }
-  } else {
-    throw ERROR_MESSAGE_TYPE(
-      message.body.type,
-      MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_CLASSIC,
-      MessageBodyType.SUBMIT_CLAIMS_FOR_CTYPES_PE
-    )
-  }
-  return { verified: false, claims: [] }
+  throw SDKErrors.ERROR_MESSAGE_TYPE(
```
tjwelde, line 195: Nitpick: you could negate the conditional and move the error up for better readability

docs/example.ts
```diff
@@ -1,27 +1,14 @@
...
-  CType,
-  CTypeUtils,
-  ICType,
-  Identity,
-  IRevocationHandle,
-  PublicAttesterIdentity,
-} from '../src'
-import { BlockchainUtils } from '../src/blockchain'
-import constants from '../src/test/constants'
+import * as Kilt from '@kiltprotocol/sdk-js'
```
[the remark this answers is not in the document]
rflechtner: I'm having issues with type annotations here. When using `Kilt.some.type` on any type annotation, it tells me there is no namespace "Kilt". The only solution is to import the types separately. Is that better?

rflechtner:
@tjwelde : I applied requested changes. I also realised there were still some functions that were async just because of the changes portablegabi had introduced. So I changed that.

rflechtner:
this is ready now :)

packages/actors_api/src/actors/Claimer.ts
```diff
@@ -121,19 +121,20 @@ export async function requestAttestation(
...
- * @param identity The Claimer's [[Identity]] which owns the [[AttestedClaim]].
+ * @param _identity The Claimer's [[Identity]] which owns the [[AttestedClaim]].
  * @param message The session object corresponding to the [[ISubmitAttestationForClaim]].
  * @param session The ClaimerAttestationSession which corresponds to the message and [[AttestedClaim]].
  * @returns A signed and valid [[Credential]].
  */
-export async function buildCredential(
-  identity: Identity,
+export function buildCredential(
+  _identity: Identity,
```
tjwelde, line 134: Should we just do it now?

docs/example.ts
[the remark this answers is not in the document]
tjwelde: Yes, I like it!

tjwelde (approved):
LGTM

PR #323, opened 2021-01-13
ci: caching in gh builds

rflechtner:
## fixes The Global Climate Crisis
Avoid unnecessary builds of the sdk to save the environment 🌳 🌍 🌳 
Builds the sdk once for each node version, zips the result, and uploads it as artefact. Each successive step (test, lint, integration test, run examples, etc.) just pulls the appropriate build, unzips, and continues from there. 

## How to test:
Just look at the checks it already ran.

## Checklist:


.github/workflows/tests.yml
```diff
@@ -30,6 +30,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    env:
+      CI: true
```
tjwelde, line 34: this should be set by github actions already automatically

tjwelde (commented):
I like it

.github/workflows/tests.yml
```diff
@@ -42,26 +45,78 @@ jobs:
...
       matrix:
         image: ['latest-develop', 'latest-rc']
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/download-artifact@v2
+        with:
+          name: build-12.x
+      - name: unzip
+        run: unzip build.zip -d .
```
tjwelde, line 119: these steps could maybe be abstracted away. A custom action maybe?

.github/workflows/tests.yml
rflechtner: ok, will remove it then. I had just moved it from the steps level to the job level.

.github/workflows/tests.yml
rflechtner: there are zip/unzip actions out there, just felt like its overkill to replace a single command

rflechtner: could just help with porting to different os

rflechtner: we could use this https://github.com/TonyBogdanov/zip ; it uses the exact same tooling but launches an alpine linux container to do so :D

tjwelde: no it's fine. as we talked directly, let's not overoptimize it

tjwelde (approved):

PR #324, opened 2021-01-20
tests: recoverable error scenarios

LeonFLK:
## fixes KILTProtocol/ticket#845, fixes Kiltprotocol/ticket#1062

Tries to trigger specific recoverable errors by sending colluding transactions.
Adds one reroute wrapper to match rpc-core returned errors to known SDKErrors.
Added tx errors to SDKErrors, wrapper matches to these.

## How to test:
start local chain
yarn test:integration:run "./packages/core/src/__integrationtests__/Blockchain.spec.ts" 

## Checklist:


packages/core/src/blockchain/Blockchain.ts
```diff
@@ -76,7 +77,10 @@ export default class Blockchain implements IBlockchainApi {
 
   public constructor(api: ApiPromise) {
     this.api = api
-    this.portablegabi = new gabi.Blockchain('portablegabi', this.api as any)
+    this.portablegabi = new gabi.Blockchain(
+      'portablegabi',
+      this.api as ApiPromise & IPortablegabiApi<PgabiModName>
+    )
```
rflechtner, line 83: this is not going to be around for long, hehe

LeonFLK: Ohh yes true, this is only a typing fix.

rflechtner (commented):
I had a few questions, should we talk about this in person later?

packages/core/src/blockchain/Blockchain.utils.ts
```diff
@@ -90,36 +99,74 @@ export function parseSubscriptionOptions(
...
+export async function submitSignedTxRaw(
   tx: SubmittableExtrinsic,
   opts: SubscriptionPromiseOptions
 ): Promise<SubmittableResult> {
   log.info(`Submitting ${tx.method}`)
   const { promise, subscription } = makeSubscriptionPromise(opts)
+  const catcher = async (reason: Error): Promise<never> => {
+    return Promise.reject(reason)
+  }
+  const unsubscribe = await tx.send(subscription).catch(catcher)
```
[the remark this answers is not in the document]
LeonFLK: Can be removed, as this is not necessary to differentiate between relevant errors.

packages/core/src/blockchain/Blockchain.utils.ts
```diff
@@ -27,15 +31,20 @@ export type SubscriptionPromiseOptions = TerminationOptions<SubmittableResult>
...
+export const IS_RELEVANT_ERROR: ErrorEvaluator = (err: Error | SDKError) => {
+  const outdated = err.message.includes(ERROR_TRANSACTION_OUTDATED().message)
+
+  const priority = err.message.includes(ERROR_TRANSACTION_PRIORITY().message)
+
+  const usurped = err.message.includes(ERROR_TRANSACTION_USURPED().message)
+
+  const duplicate = err.message.includes(ERROR_TRANSACTION_DUPLICATE().message)
+
+  return outdated || usurped || priority || duplicate
```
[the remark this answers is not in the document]
LeonFLK: I'll write a typeguard to differentiate SDKErros from normal Errors, and conditionally match these to the relevant errors by the errorCode.

packages/core/src/blockchain/Blockchain.utils.ts
```diff
@@ -90,36 +99,74 @@ export function parseSubscriptionOptions(
...
+        return Promise.reject(ERROR_TRANSACTION_PRIORITY())
+        break
+      case reason.message.includes(TxDuplicate):
+        return Promise.reject(ERROR_TRANSACTION_DUPLICATE())
+        break
+      default:
+        return Promise.reject(reason)
+    }
+  })
+}
```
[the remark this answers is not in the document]
LeonFLK: SubscriptionPromise maker needs to be refactored, as it doesn't cover all the cases that it seems to do.
IS_ERROR sadly relies entirely on the return of a SubmittableResult and it's statuses.

packages/core/src/blockchain/Blockchain.utils.ts
```diff
@@ -7,8 +7,14 @@
...
 import {
+  ErrorCode,
+  ERROR_TRANSACTION_DUPLICATE,
+  ERROR_TRANSACTION_OUTDATED,
+  ERROR_TRANSACTION_PRIORITY,
   ERROR_TRANSACTION_RECOVERABLE,
   ERROR_TRANSACTION_USURPED,
+  isSDKError,
+  SDKError,
 } from '../errorhandling/SDKErrors'
```
rflechtner, line 18: can you import everything as SDKErrors ? It's how I currently export the SDKErrors from the utils package. If we don't want to expose every single error type directly from the utils package, that's how this would have to be handled.

packages/core/src/blockchain/Blockchain.utils.ts
```diff
@@ -90,36 +98,72 @@ export function parseSubscriptionOptions(
...
 
-  const result = await promise
-    .catch(async (reason: Error) => {
-      if (reason.message === ERROR_TRANSACTION_USURPED().message) {
-        return Promise.reject(ERROR_TRANSACTION_RECOVERABLE())
-      }
-      return Promise.reject(reason)
-    })
-    .finally(() => unsubscribe())
+  const result = await promise.finally(() => unsubscribe())
```
rflechtner, line 110: you can just return here
```suggestion
  return promise.finally(() => unsubscribe())
```

rflechtner: love how it has gotten so much cleaner though

packages/core/src/blockchain/Blockchain.utils.ts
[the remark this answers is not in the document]
rflechtner: It is as I feared. The SubscriptionPromise only handles status updates during the transaction lifetime. If `tx.send()` fails to submit the transaction (connection fails, transaction is erroneous and directly rejected) it is never updated and  tx.send just rejects. I agree we need to handle this how you do it. It's tricky stuff!

rflechtner (approved):
LGTM, merely nitpicks

packages/core/src/blockchain/Blockchain.utils.ts
LeonFLK: Yes good suggestion!

packages/core/src/blockchain/Blockchain.utils.ts
LeonFLK: Ohh yes of course, finally don't have to have this weird thing.

PR #326, opened 2021-01-27
chore: split types from core package

LeonFLK:
## fixes KILTProtocol/ticket#843

Removed default exports in favor of named ones, we should discuss if we want to completely ditch the default export, since named ones are best-practice.

## Checklist:


rflechtner (approved):
Love it!

PR #327, opened 2021-01-28
feat: utils package

rflechtner:
## fixes KILTProtocol/ticket#837
Moves general stateless functions into a utils package.
Also added documentation for Crypto utils and added tests as the coverage was too low.

N.B.: I simplified code on most decoder functions as we had a check on there that has become redundant.

## How to test:
Tests, as usual

## Checklist:


rflechtner:
Seems like we need to figure out where to put development tools that should not be published, but are used during development (mocking tools, jest add-ons). This could either be a directory somewhere or a package which we don't publish and import as a dev dependency.

rflechtner:
Albi broke the sdk integration again but it tests successfully against yesterday's dev node.

weichweich:
> Albi broke the sdk integration again but it tests successfully against yesterday's dev node.

just making sure that my PR gets merged first! 😝

rflechtner:
> Albi broke the sdk integration again but it tests successfully against yesterday's dev node.

that's fixed !

weichweich:
> > Albi broke the sdk integration again but it tests successfully against yesterday's dev node.
> 
> that's fixed !

and it wasn't even my fault! 😮

packages/utils/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
...
+
+   "include": [
+    "src/**/*.ts", "src/**/*.js"
+  ],
+
+   "exclude": [
+    "coverage",
+    "**/*.spec.ts",
+  ]
+}
```
LeonFLK, line 16:
```suggestion
}

```

EOF newline

LeonFLK (commented):
LGTM

Nitpick: add EOF newline.
Other than that, do you think the `SubscriptionPromise.ts` should live inside the utils package as well?

rflechtner:
> LGTM
> 
> Nitpick: add EOF newline.
> Other than that, do you think the `SubscriptionPromise.ts` should live inside the utils package as well?

Timo & I thought it makes more sense on the BlockchainUtils. But it is something we can consider, it's still not clear to me where it makes most sense

LeonFLK:
> > LGTM
> > Nitpick: add EOF newline.
> > Other than that, do you think the `SubscriptionPromise.ts` should live inside the utils package as well?
> 
> Timo & I thought it makes more sense on the BlockchainUtils. But it is something we can consider, it's still not clear to me where it makes most sense

Yes that's what I thought, either inside utils or blockchain helper/utils.

LeonFLK (approved):
LGTM



PromiseSubscription goes into Blockchain Helpers

PR #328, opened 2021-01-29
chore: bump deps, remove lib compiler path

LeonFLK:
## fixes KILTProtocol/ticket#1048

Bumps polkadot deps to next minor release

## Checklist:


packages/vc-export/package.json
```diff
@@ -26,11 +26,12 @@
   },
   "dependencies": {
     "@kiltprotocol/core": "workspace:*",
+    "@kiltprotocol/sdk-js": "workspace:*",
```
tjwelde, line 29: This doesn't seem necessary
```suggestion
```

tjwelde (changes requested):

PR #329, opened 2021-02-03
feat: split config from core

LeonFLK:
## fixes KILTProtocol/ticket#869

Adds it as peerDep to the appropriate packages
Since SDKErrors are not split yet to utils, the Error had to be replaced
Renamed `factory` after request to `LogFactory`

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:


packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -41,7 +41,7 @@ export async function buildConnection(
 }
 
 export async function getCached(
-  host: string = get('address')
+  host: string = ConfigService.get('address')
```
rflechtner, line 44: 😍

packages/config/src/ConfigService.ts
```diff
@@ -52,7 +51,7 @@ let configuration: configOpts = {
 
 function checkAddress(): void {
   if (!configuration.address) {
-    throw ERROR_WS_ADDRESS_NOT_SET()
+    throw Error('Node address to connect to not configured!')
```
rflechtner, line 54: I suggest we merge this after the utils, then you can import the error from there

rflechtner (changes requested):
LGTM but I suggest we wait with this until the utils are merged so we can use the appropriate SDK error here.

rflechtner (approved):
LGTM!

PR #330, opened 2021-02-03
Update for delegation&attestation refactor

weichweich:
fixes KILTprotocol/ticket/issues/1063


packages/core/src/delegation/DelegationDecoder.ts
```diff
@@ -11,39 +11,41 @@
...
-    if (json instanceof Array) {
-      return {
-        cTypeHash: json[0],
-        account: json[1],
-        revoked: json[2],
-      }
+  assertCodecIsType(encoded, ['Option<DelegationRoot>'])
+  if (encoded instanceof Option && encoded.isSome) {
+    const delegationRoot = encoded.unwrap()
+    // TODO: check that root is none
```
weichweich, line 44:
```suggestion
  if (encoded.isSome) {
    const delegationRoot = encoded.unwrap()
```

packages/actors_api/src/actors/Attester.ts
```diff
@@ -92,12 +93,46 @@ export async function issueAttestation(
...
   revocationHandle: IRevocationHandle
 ): Promise<void> {
-  await Attestation.revoke(revocationHandle.claimHash, attester).then((tx) =>
-    BlockchainUtils.submitTxWithReSign(tx, attester)
-  )
+  const attestation = await Attestation.query(revocationHandle.claimHash)
+
+  if (attestation === null) {
+    throw SDKErrors.ERROR_NOT_FOUND('Attestation not on chain')
+  }
```
rflechtner, line 106: mmh that means we were submitting a transaction before with an extrinsic that's doomed to fail? Guess this is more efficient now :D

packages/actors_api/src/actors/Attester.ts
```diff
@@ -92,12 +93,46 @@ export async function issueAttestation(
...
+  // count the number of steps we have to go up the delegation tree for calculating the transaction weight
+  let delegationTreeTraversalSteps = 0
+
+  // if the attester is not the owner, we need to check the delegatoin tree
```
rflechtner, line 110:
```suggestion
  // if the attester is not the owner, we need to check the delegation tree
```

packages/core/src/delegation/DelegationNode.ts
```diff
@@ -129,6 +129,41 @@ export default class DelegationNode extends DelegationBaseNode
...
+
+    // while there is a parent and we have a node but not the right node ...
+    while (
+      delegationNode !== null &&
+      typeof delegationNode.parentId !== 'undefined' &&
+      delegationNode?.account !== address
+    ) {
+      delegationTreeTraversalSteps += 1
+
+      // this 'await in loop' is unavoidable because we need the node for the next await
```
rflechtner, line 156: I guess recursively attaching `.then()` clauses is the only other option and that may hit the call stack limit at some point

rflechtner: (recursion instead of a `while`)

rflechtner (commented):
LGTM so far but I see your still working on those tests :D

packages/core/src/delegation/DelegationNode.ts
weichweich: I wouldn't make the code worse only to avoid a lint error. 😅 I couldn't come up with a solution that wouldn't make the code worse.

rflechtner (approved):
LGTM

PR #297, opened 2020-09-07
feat: Message compression & decompression

Dudleyneedham:
## fixes KILTProtocol/ticket#667

Creating a message utility to compress and decompress most messages. 
Testing each compression and decompression.
Additionally, had to add all the types along with checks within the actors to see if the message has been compressed etc.

## How to test:

Created a test suite to check all the compression and decompression.

## Checklist:

- [x] I have verified that the code works
- [x] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [x] I have [left the code in a better state](https://deviq.com/boy-scout-rule/)
- [x] I have documented the changes (where applicable)

PR #331, opened 2021-02-03
feat: Message compression & decompression

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/667

New PR for ticket. Original PR and history found here. https://github.com/KILTprotocol/sdk-js/pull/297

Creating a message utility to compress and decompress most messages.
Testing each compression and decompression.
Additionally, had to add all the types along with checks within the actors to see if the message has been compressed etc.

## How to test:

Created a test suite to check all the compression and decompression.

## Checklist:


packages/core/src/claim/Claim.ts
```diff
@@ -151,8 +156,16 @@ export default class Claim implements IClaim {
    * @returns A new [[Claim]] object.
    */
 
-  public static decompress(compressedClaim: CompressedClaim): Claim {
-    const decompressedClaim = ClaimUtils.decompress(compressedClaim)
+  public static decompress(
+    compressedClaim: CompressedClaim
+  ): Claim | PartialClaim {
```
rflechtner, line 161: For typechecking the `Claim` type has no effect, but it's nice as a hint for the consumer

rflechtner: you could work with overloads as in ClaimUtils

packages/core/src/claim/Claim.ts
```diff
@@ -151,8 +156,16 @@ export default class Claim implements IClaim {
    * @returns A new [[Claim]] object.
    */
 
-  public static decompress(compressedClaim: CompressedClaim): Claim {
-    const decompressedClaim = ClaimUtils.decompress(compressedClaim)
+  public static decompress(
+    compressedClaim: CompressedClaim
```
rflechtner, line 160: to be nitpicky, the CompressedPartialClaim is missing on the signature

packages/core/src/claim/Claim.ts
rflechtner: just thinking, a function that either returns an instantiated class instance or just some general object feels a little tricky to use. What do we need it for?

rflechtner (commented):
just a few nitpicks & questions :)

docs/getting-started.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-import Kilt from '@kiltprotocol/sdk-js'
```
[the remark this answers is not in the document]
Dudleyneedham: It is an `enum` and doesn't import.

packages/core/src/claim/Claim.ts
Dudleyneedham: Maybe, separating the two would be an option.

Dudleyneedham: Another option would be not creating a claim. Either returning a full claim or partial claim. Then the user can handle as they need.

Dudleyneedham: I would prefer the second option.

packages/core/src/claim/Claim.ts
Dudleyneedham: What do you mean the signature?

rflechtner: could have written input type, but what I meant was the [function signature](https://developer.mozilla.org/en-US/docs/Glossary/Signature/Function)

packages/core/src/claim/Claim.ts
rflechtner: that would be aliasing the decompress function in the claim.utils, right? Because that's what that one does. I would also prefer that option.

rflechtner (changes requested):
I suggest you just call `ClaimUtils.decompress()` in the static method on the `Claim` class and copy the full signature (including overloads). This should be all you need to do what you set out to do with this method.

docs/getting-started.ts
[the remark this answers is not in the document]
rflechtner: In that case, maybe we should make sure we export this properly from core?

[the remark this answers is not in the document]
Dudleyneedham: Yes, I guess we can have it as its own export. 

I don't like how it looks neither.

packages/core/src/claim/Claim.ts
Dudleyneedham: Ok, good to know the specific name. 

Thanks, I have updated it.

packages/core/src/claim/Claim.ts
Dudleyneedham: Yes, I have updated this as well.

rflechtner (approved):
LGTM! took the liberty of fixing the docstrings (copied over from the decompression function on the utils to be a little more complete) and the import on the getting started.

PR #336, opened 2021-02-10
ci: force all prerelease packages to have the same version

rflechtner:
## no ticket
I noticed our prereleases on npm may differ in version number. That's no biggie, but we could force them to be equal within a prerelease with this proposed change. This way there is no confusion which prereleased subpackage goes with which other.

## How to test:
You can try running the commands I added locally. Use bash.

## Checklist:


tjwelde (approved):
LGTM

PR #340, opened 2021-02-15
docs: Actors readme

rflechtner:
## fixes KILTProtocol/ticket#1058
Adds a readme to the actors api package.
Also fixes a docstring in it.

## How to test:
Pull up the readme on your screen. Put your cat on your keyboard. If it hasn't produced an attested credential within 5 minutes, it's garbage.
Your cat, not the instructions. Those are great.

## Checklist:


packages/actors_api/README.md
```diff
@@ -0,0 +1,47 @@
...
+- Claimer
+- Attester
+- Verifier
+
+It is meant to make relevant functionality easily accessible and discoverable.
+
+## Claimer
+
+This section contains functions to request an attestation and to send it to a Verifier (presentation).
+`requestAttestation` allows you to request an attestation for a claim from an Attester, which returns a message object that can sent off to the Attester and a session object to keep track of the details of the request.
```
tjwelde, line 21: To make this a new paragraph in the resulting html, you have to either add 2 spaces to the end of the previous line for a single carriage return or a new line for 2 carriage returns.

tjwelde (changes requested):
Really like it. Just a small typing issue

packages/actors_api/README.md
rflechtner: pro tip: newline after each sentence. Is ignored, as you say, but makes diffs so much cleaner and more useful.
Trust me, I wrote my thesis in markdown :D

rflechtner: but I agree that paragraphs would be nice here

tjwelde (approved):
LGTM

PR #342, opened 2021-02-16
fix: add missing did type

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1086

Type IDidDocumentSigned was used within the Demo-client but wasn't accessible as not exported.
Added a version fix to the utils folder not being updated.

## How to test:

n/a

## Checklist:


wischli (approved):
LGTM

PR #345, opened 2021-02-18
feat: use Sr25519 & other signature key types

rflechtner:
## fixes KILTProtocol/ticket#977
This enables use of different types of signing keys (was `ed25519`) so that users can now use the full range of key types supported by substrate. Also switched the default key type to `sr25519`.

Note that using `sr25519` means that cryptographic functions implemented in Rust are called, which are supplied via a WASM included in **polkadot.js**. This requires calling their `cryptoWaitReady()` and awaiting the resulting Promise. To this end, we now have a `Kilt.init()` function which does exactly that plus allows passing parameters that will be handed to `Kilt.config()`. It can thus become the general setup step you always call after importing the sdk.

Key type is also dynamically written to the proof when exporting a credential to VC. For lack of standards, this uses a number of experimental/proposed denominations which are not in wide use at this point.

Tests have been minimally refactored to work with the current changes. To keep effort low, wherever I ran into problems stemming from key type (e.g. a signature or address was included in plain text), I just switched the respective identities back to `ed25519`.

I also fixed and updated the `getting-started.ts` and `example.ts`, and also made sure `getting-started.md` is up to date with  its `.ts` equivalent.

## How to test:
unit & integration test, examples.

## Checklist:


tjwelde:
So this also takes care of https://github.com/KILTprotocol/ticket/issues/1107 and https://github.com/KILTprotocol/ticket/issues/1126 ?

rflechtner:
> So this also takes care of [KILTprotocol/ticket#1107](https://github.com/KILTprotocol/ticket/issues/1107) and [KILTprotocol/ticket#1126](https://github.com/KILTprotocol/ticket/issues/1126) ?

It makes sure that getting-started.ts works correctly and makes sense and that getting-started.md contains comparable code. It also makes sure example.ts works and makes sense. Haven't checked for quality of the two other than that.

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -56,6 +57,7 @@ export async function buildConnection(
     provider,
     types: CUSTOM_TYPES,
   })
+  await cryptoWaitReady()
```
tjwelde, line 60: Why do we need it in `buildConnection`?

packages/core/src/identity/Identity.spec.ts
```diff
@@ -65,7 +75,9 @@ describe('Identity', () => {
   })
 
   it('should have different keys for signing and boxing', () => {
-    const alice = Identity.buildFromMnemonic(Identity.generateMnemonic())
+    const alice = Identity.buildFromMnemonic(Identity.generateMnemonic(), {
+      signingKeyPairType: 'ed25519',
+    })
     expect(Crypto.coToUInt8(alice.signPublicKeyAsHex)).not.toEqual(
       alice.boxKeyPair.publicKey
     )
```
tjwelde, line 83: Should we maybe add some SR25519 tests?

tjwelde (commented):
Just some questions

tjwelde:
Since you are also updating the gettins-started guide here: Please do not use any actors here.

tjwelde:
Also, this has to work with browsers. please test with a new create-react-app app

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
rflechtner: we don't, but I though it's a nice feature. Shall I remove?

rflechtner: or move it to Kilt.connect() maybe?

packages/core/src/identity/Identity.spec.ts
rflechtner: I didn't really touch tests but maybe as the minimum requirement we may want to test signing and encryption with all three key types.
Everything else I'd say will be done as part of the bigger outstanding test refactor.

tjwelde: yes, just the basics here

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
tjwelde: If you want to add this convenience, I think `Kilt.connect()` could be a good place, because when you have an application, which disconnects the chain, whenever it goes into background mode, you will call Kilt.connect to reconnect. On the other hand, an application might just want to leverage the implicit connections...

Since it is only connected to the Identities, adding it to blockchain connection related functions feels like the wrong place, though. I guess I would remove it here, but add a disclaimer in the documentation, that you have to call it, whenever you want to do something with the identities.

Can we not just call it, whenever we do identity functions?

packages/core/src/identity/Identity.ts
```diff
@@ -37,6 +42,14 @@ type BoxPublicKey =
 export default class Identity {
   private static ADDITIONAL_ENTROPY_FOR_HASHING = new Uint8Array([1, 2, 3])
 
+  public static async waitReady(): Promise<boolean> {
+    return cryptoWaitReady()
+  }
+
+  public static get ready(): boolean {
```
tjwelde, line 49: Please add jsdoc

packages/core/src/identity/Identity.ts
```diff
@@ -37,6 +42,14 @@ type BoxPublicKey =
 export default class Identity {
   private static ADDITIONAL_ENTROPY_FOR_HASHING = new Uint8Array([1, 2, 3])
 
+  public static async waitReady(): Promise<boolean> {
```
tjwelde, line 45: Please add jsdoc

tjwelde (approved):

packages/core/src/blockchainApiConnection/BlockchainApiConnection.ts
rflechtner: removed it, as discussed

PR #348, opened 2021-02-24
refactor: on-chain DID type

wischli:
## fixes KILTProtocol/ticket#1116
Applies changes to DID type on the mashnet-node proposed in https://github.com/KILTprotocol/mashnet-node/pull/118 as a result of mashnet-node CI failure due to clippy.

## How to test:

```
docker pull kiltprotocol/mashnet-node:develop
docker run -p 9944:9944 kiltprotocol/mashnet-node:develop --dev --ws-port 9944 --ws-external --rpc-external
yarn test:integration
```

## Checklist:

### DIFF
--- a/packages/core/src/did/Did.utils.ts
+++ b/packages/core/src/did/Did.utils.ts
@@ -30,7 +30,7 @@ export function decodeDid(
   identifier: string,
   encoded: Option<IEncodedDidRecord>
 ): IDid | null {
-  DecoderUtils.assertCodecIsType(encoded, ['Option<Did>'])
+  DecoderUtils.assertCodecIsType(encoded, ['Option<DidRecord>'])
   if (encoded.isSome) {
     const did = encoded.unwrap()
     return {
--- a/packages/core/src/did/Did.utils.ts
+++ b/packages/core/src/did/Did.utils.ts
@@ -34,11 +34,14 @@ export function decodeDid(
   DecoderUtils.assertCodecIsType(encoded, ['Option<DidRecord>'])
   if (encoded.isSome) {
     const did = encoded.unwrap()
+    const documentStore = did.docRef.isSome
+      ? hexToString(did.docRef.toHex())
+      : undefined
     return {
       identifier,
       publicSigningKey: did.signKey.toString(),
       publicBoxKey: did.boxKey.toString(),
-      documentStore: hexToString(did.docRef.unwrapOrDefault().toHex()),
+      documentStore,
     }
   }
 
--- a/packages/core/src/did/Did.utils.ts
+++ b/packages/core/src/did/Did.utils.ts
@@ -7,6 +7,7 @@ import { Option, Struct, u8, Vec } from '@polkadot/types'
 import { IPublicIdentity } from '@kiltprotocol/types'
 import { Crypto, DecoderUtils, SDKErrors } from '@kiltprotocol/utils'
 import { Hash } from '@polkadot/types/interfaces'
+import { hexToString } from '@polkadot/util'
 import Identity from '../identity/Identity'
 import {
   CONTEXT,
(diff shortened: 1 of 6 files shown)


wischli:
CI will fail until the corresponding [mashnet-node PR](https://github.com/KILTprotocol/mashnet-node/pull/118) is merged.

wischli:
Have to wait for new `mashnet-node:develop` image to be published before expecting success in integration tests..

packages/core/src/did/Did.utils.ts
```diff
@@ -143,7 +141,7 @@ export function verifyDidDocumentSignature(
   if (identifier !== id) {
     throw SDKErrors.ERROR_DID_IDENTIFIER_MISMATCH(identifier, id)
   }
-  const unsignedDidDocument = { ...didDocument }
+  const unsignedDidDocument = { ...didDocument, signature: undefined }
```
tjwelde, line 144: Why is that necessary, when we remove the signature right after?

wischli: The linter complained about the deletion.

wischli: I am happy to remove this as this is independent of the PR.

rflechtner: proposal: instead of deletion below, why don't we do 
```typescript
const { signature, ...unsignedDidDocument } = didDocument
```

wischli: Not sure whether this is fine because then the `signature` key still exists.

rflechtner: it does not in unsignedDidDocument, but of course we then have a variable in scope with that name - not sure how to improve this (the linter will complain)

wischli: Okay so I will change it to this, which won't have a linter complain:
```ts
  const { signature, ...unsignedDidDocument } = didDocument
  return Crypto.verify(
    Crypto.hashObjectAsStr(unsignedDidDocument),
    signature,
    getAddressFromIdentifier(identifier)
  )
```

wischli: Fixed in 692eca0

rflechtner: oooh I love it! didn't see we need the signature anyway, that's perfect.

packages/core/src/did/Did.utils.ts
```diff
@@ -21,25 +20,24 @@ import {
...
+  if (encoded.isSome) {
+    const did = encoded.unwrap()
     return {
       identifier,
-      publicSigningKey: decoded[0],
-      publicBoxKey: decoded[1],
-      documentStore,
+      publicSigningKey: did.signKey.toString(),
+      publicBoxKey: did.boxKey.toString(),
+      documentStore: did.docRef.unwrapOrDefault().toHuman()?.toString(),
```
rflechtner, line 40: doesn't this just build an string presentation of an array? why do we no longer use `hexToString`

wischli: Good point, fixed in f494360

rflechtner: I think to get to the target of `string | undefined` we even need to do something like 
```typescript
did.docRef.isSome ? hexToString(did.docRef.toHex()) : undefined
```
(sidetone: seems like `toHex()` handles a `none` option directly by returning `0x`)

wischli: Fixed in a8496ed. I was wondering why we don't to any real integration tests with DIDs or was I not looking thoroughly enough? I wanted to test my and your proposal but did not want to invest time into setting up a test for this.

rflechtner (approved):
LGTM

PR #349, opened 2021-02-26
fix: delegation revocation

rflechtner:
## fixes KILTProtocol/ticket#1121
Delegation revocation now requires max child revocations and max upwards search depth as parameters of the transaction. This adds these parameters to the delegation `revoke` functions and adds a solution to count child nodes on chain.

Also amends integration and unit tests as this case was not covered and therefore this issue was not immediately detected.

As `findParent` was not performing as expected when the revoking identity was owner of the root node, this was also refactored. To this purpose I found it easiest to move this function to the `DelegationBaseNode` class, which means it is available on both the root and (child) delegation node classes. I also renamed that function to `isDelegating` to match the function name on the node side, which implements similar logic.

## How to test:
Run integration tests, test in demo-client.

## Checklist:


packages/core/src/delegation/DelegationNode.ts
```diff
@@ -193,8 +158,21 @@ export default class DelegationNode extends DelegationBaseNode
...
    */
   public async revoke(identity: Identity): Promise<SubmittableExtrinsic> {
-    log.debug(`:: revoke(${this.id})`)
-    return revoke(this.id, identity)
+    const { steps, node } = await this.isDelegating(identity.address)
+    if (!node) {
+      throw SDKErrors.ERROR_UNAUTHORIZED(
+        `Identity with address ${identity.address} is not among the delegators and may not revoke this node`
+      )
+    }
```
rflechtner, line 166: this check is optional, but given that we are required to traverse the whole tree which means retrieving the information whether we can revoke or not, it seems dumb to spam the network with a transaction that is doomed to fail.

packages/core/src/delegation/DelegationNode.ts
```diff
@@ -193,8 +158,21 @@ export default class DelegationNode extends DelegationBaseNode
...
+    const childCount = await this.subtreeNodeCount()
+    // must revoke all children and self
+    const revocationCount = childCount + 1
+    // node side implementation requires steps + 1
+    const depth = steps + 1
```
rflechtner, line 171: I'd like to change this node-side, but let's do that separately.

packages/core/src/delegation/Delegation.ts
```diff
@@ -104,4 +104,50 @@ export default abstract class DelegationBaseNode
...
    */
   public abstract revoke(identity: Identity): Promise<SubmittableExtrinsic>
+
+  /**
+   * Checks on chain whether a given identity is delegating to the current node.
+   *
+   * @param address The address of the identity.
+   * @returns An object containing a `node` owned by the identity if it is delegating, plus the number of `steps` traversed. `steps` is 0 if the address is owner of the current node.
+   */
+  public async isDelegating(
```
tjwelde, line 114: Just a nitpick: the `is` in `isDelegating`, suggests a boolean return value, but it is a complex object.
I don't really have a much better name right now, though. Something with `find`? Like `findDelegatingNode`?

tjwelde (commented):
I like it. Just a nitpick I want to talk about

packages/core/src/delegation/Delegation.ts
rflechtner: mhhh yeah I meant to give it the same name as the function on the node, but of course the return type is odd for that. It was findParent before, so findDelegatingNode would work nicely

rflechtner: or let's go super descriptive: `findAncestorOwnedBy(address)`

tjwelde: uuuhhhh, I love it!

tjwelde (approved):
LGTM

PR #350, opened 2021-03-02
chore: update JSDoc

Diiaablo95:
- Update Typedoc dependency to ^0.19
- Adjust some doc errors

## fixes KILTProtocol/ticket#1103
I updated the documentation to repair broken links and inconsistencies with the current SDK status. Nevertheless, as we also discussed, I think the documentation still needs a large reworking, as different parts of the SDK have very different documentation styles.

## Checklist:


tjwelde:
While I was looking at this I noticed, that now with the SDK split-up, the structure of the API reference doesn't make much sense anymore.
In the past we have used `typedoc-plugin-external-module-name` to aggregate files in folders into separate modules. But even if we use it again, we might have to clarify what lives in what package.

tjwelde:
As discussed:
- Remove typedoc related files and commands from all packages
- Make sure the corresponding github workflow uses the command from root
- Exclude lib folder
- Fix paths showing up in documenation
Thanks ☺️

Diiaablo95:
Ok I guess I fixed the errors of the CI. @tjwelde or perhaps @rflechtner would you mind taking a look? Thanks!

packages/chain-helpers/src/blockchain/Blockchain.spec.ts
```diff
@@ -1,10 +1,8 @@
 /**
- * @packageDocumentation
  * @group unit/blockchain
- * @ignore
  */
 
-/* eslint-disable dot-notation */
+/* eslint-disable */
```
tjwelde, line 5: this will disable ALL eslint checks. Is that wanted here?

packages/core/src/__integrationtests__/Attestation.spec.ts
```diff
@@ -1,9 +1,3 @@
-/**
- * @packageDocumentation
- * @group integration/attestation
```
tjwelde, line 3: We have to add the `@group` tags to the integration tests, so that they are triggered correctly via the `yarn run test:integration:run` command

tjwelde (changes requested):
Looks very clean again, just small things

tjwelde (approved):
LGTM

PR #351, opened 2021-03-03
feat: export other packages as named exports in sdk-js

tjwelde:
## No-Ticket
While updating the sdk into the demo-client, it came up, that you can't import stuff like `BlockchainUtils` from the `sdk-js` package as a named import.

## Checklist:


Dudleyneedham (approved):
LGTM

PR #354, opened 2021-03-04
chore: fixes to be compliant with Substrate 3.0.0

wischli:
## fixes SDK incompatibilities with KILTProtocol/ticket#812

**This PR should not be merged before we updated the mashnet-node image. See [below](https://github.com/KILTprotocol/sdk-js/pull/354#issuecomment-791310000) for explanation!**

* Minor fixes which were required to make CI tests run with the runtime-upgraded testnetwork
* **Updated Polkadot dependencies + BN** to latest version because we require some of their changes with Substrate 3.0.0
* Related to changes made to the mashnet node in
   * https://github.com/KILTprotocol/mashnet-node/pull/122
   * https://github.com/KILTprotocol/mashnet-node/pull/119
  

## How to test:
Very hard to test. You would have to run the updated testnet locally and do more changes to the SDK:
* Change `Development` chain name to `KILT Testnet`
* Change identities to use TestNet authority hexes
* Change `IS_FINALIZED` to `IS_IN_BLOCK` for CI tests 

## Checklist:


packages/core/src/balance/Balance.chain.ts
```diff
@@ -40,7 +40,11 @@ export async function getBalance(
   accountAddress: IPublicIdentity['address']
 ): Promise<BN> {
   const blockchain = await BlockchainApiConnection.getConnectionOrConnect()
-  return (await blockchain.api.query.system.account(accountAddress)).data.free
```
wischli, line 43: I wonder why this hasn't be an issue before as `...free` is of type `Codec`.

packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
```diff
@@ -107,10 +107,10 @@ export const ExtrinsicErrors = {
  * PalletIndex reflects the numerical index of a pallet assigned in the chain's metadata.
  */
 export enum PalletIndex {
-  CType = 11,
```
wischli, line 110: We had to change the pallet indices to be compliant with the current testnet. One should not change the order of pallets (e.g. indices for the pallets) in `construct_runtime`.

packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -24,8 +26,12 @@ export const CUSTOM_TYPES: RegistryTypes = {
   BlockNumber: 'u64',
   Index: 'u64',
   RefCount: 'u32',
-
-  ErrorCode: 'u16',
+  AccountInfo: {
```
wischli, line 29: Not sure whether this is required, probably have to remove this again. This is the current `AccountInfo` type from Polkadot.

packages/core/src/balance/Balance.chain.ts
weichweich: This is probably there because you added „balance: u128“? That overwrites the internal balance type or something. Just a guess...

wischli: I tested this thoroughly and I couldn't make it work without calling `toString()` or `toBn()` on this value, independent of setting `Balance: 'u128'` or `AccountInfo` in the custom types.

I would have loved to use `.toBn()` but for some reason the mock didn't know this function even after updating all Polkadot dependencies to the latest type, and setting the correct type for `system.accountInfo`.

packages/utils/package.json
```diff
@@ -27,10 +27,10 @@
   },
   "dependencies": {
     "@kiltprotocol/types": "workspace:*",
-    "@polkadot/keyring": "^5.4.4",
-    "@polkadot/types": "^3.6.4",
-    "@polkadot/util": "^5.4.4",
-    "@polkadot/util-crypto": "^5.4.4",
+    "@polkadot/keyring": "^5.9.2",
```
wischli, line 30: We needed to update for the current version of `system.accountInfo` which used to be used `AccountInfo` and now its `AccountInfoWithProviders`.

packages/chain-helpers/src/blockchainApiConnection/__mocks__/BlockchainApiConnection.ts
```diff
@@ -254,8 +258,11 @@ const __mocked_api: any = {
       // default return value decodes to BN(0)
       // default return value decodes to AccountInfo with all entries holding BN(0)
       account: jest.fn(
-        async (address: IPublicIdentity['address'], cb): Promise<AccountInfo> =>
-          TYPE_REGISTRY.createType('AccountInfo')
+        async (
+          address: IPublicIdentity['address'],
+          cb
+        ): Promise<AccountInfoWithProviders> =>
+          TYPE_REGISTRY.createType('AccountInfoWithProviders')
```
wischli, line 265: As explained below, `AccountInfo` is the type used in Substrate 2.0 but with Substrate 3.0, we have to use `AccountInfoWithProviders`.

packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -24,8 +26,7 @@ export const CUSTOM_TYPES: RegistryTypes = {
   BlockNumber: 'u64',
   Index: 'u64',
   RefCount: 'u32',
-
-  ErrorCode: 'u16',
+  AccountInfo: 'AccountInfoWithProviders',
```
wischli, line 29: As explained below, `AccountInfo` is the type used in Substrate 2.0 but with Substrate 3.0, we have to use `AccountInfoWithProviders`.

wischli:
The CI pipeline is expected to fail because the current `mashnet-node:develop` image still has the incorrect pallet order which doesn't match the one on the Testnet. This will be fixed after we updated the develop image.

packages/core/src/balance/Balance.chain.ts
weichweich: Ah ok. Was just an idea. 😄

wischli:
@tjwelde As you can see, the integration tests now run through. Thus, this should be part of the upcoming release.

PR #358, opened 2021-03-11
docs: Dev vs Master mixups

weichweich:
# Removes Links to Develop from Master branch

Some links are hard coded to point to the master or develop branch. This can be very confusing.

Looks like GitHub can resolve paths in the local repository. Removing the domain and branch info prevents jumps between develop and master branch.

We might need to check how this works, when the README is displayed in the API doc or on NPM.

## How to test

* ho to the start page
* select this branch
* click on links to other documentation

## Checklist:

### DIFF
--- a/packages/sdk-js/README.md
+++ b/packages/sdk-js/README.md
@@ -18,6 +18,8 @@ Regarding the privacy enhancement, please have a look at our [lightning talk for
 
 ## Documentation
 
+* The Links found here are related to the `master` branch. Be aware that you might see this on the `develop` branch which is the current unstable development version. *
+
 [KILT](https://kilt.io) documentation is provided in several guides and demos.
 
 - [KILT workshop](https://github.com/KILTprotocol/kilt-workshop-101) 👈 Start here to get familiar with the basics
--- a/packages/sdk-js/README.md
+++ b/packages/sdk-js/README.md
@@ -18,7 +18,7 @@ Regarding the privacy enhancement, please have a look at our [lightning talk for
 
 ## Documentation
 
-*You are currently viewing the develop branch documentation! This is the latest development and might break. Have a look at the [master branch](https://github.com/KILTprotocol/sdk-js/tree/master) for the released version*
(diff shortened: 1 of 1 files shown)


weichweich:
@Diiaablo95 I'm not sure about the warning. It's now right above the section where all the links point to the Master branch. 😂

weichweich:
I added a warning that doesn't need to be removed when this will merge into master. But I'm still not really satisfied.

packages/sdk-js/README.md
```diff
@@ -18,6 +18,8 @@ Regarding the privacy enhancement, please have a look at our [lightning talk for
 
 ## Documentation
 
+* The Links found here are related to the `master` branch aka the current released version. * Links outside this section point to the version you are currently viewing.
```
Diiaablo95, line 21: I would rather use the > tag and optionally make the sentence bold, even thought that might be too much.

Diiaablo95, line 21: current -> latest?

Diiaablo95: I would write something like "to avoid confusion between the latest released SDK version and the `develop` default branch, the links in the list that follow point to the `master` branch, which always contains the latest official release of the SDK.

Diiaablo95 (changes requested):
I would just change a little bit the working of the warning, and display it with the > tag, which is usually used for warnings.

packages/sdk-js/README.md
weichweich: Ah didn't hat a look at the preview. Is should have been bold. :D That's why there is a `*` in the middle of the sentence. 😁

packages/sdk-js/README.md
```diff
@@ -18,6 +18,8 @@ Regarding the privacy enhancement, please have a look at our [lightning talk for
 
 ## Documentation
 
+> To avoid confusion between the latest released SDK version and the `develop` default branch, the links in the list **point to the master branch**, which always contains the latest official release of the SDK.
```
Diiaablo95, line 21: - list below or following list
- I would also make bold the whole sentence "the links in the list point to the master branch"

Diiaablo95 (commented):
I added one more suggestion. It is personal, so I won't mark it as a required change.

Diiaablo95 (approved):

PR #359, opened 2021-03-11
fix listenToBalanceChanges example

arty-name:
The current example for `listenToBalanceChanges` doesn’t take into account that the method is `async` and returns a promise. I’ve just added an `await`.

## Checklist:


weichweich:
Thanks for the PR! 🙂

PR #360, opened 2021-03-11
Add Conventional Commit CI

weichweich:
## Checklist:


weichweich:
Don't like it. Looks like there is a yarn error or it errors without properly reporting the issue. Either way nothing we want to use. 😞

PR #365, opened 2021-04-01
feat: jsonld-signatures suite to verify Kilt VCs in vc-js 

rflechtner:
## fixes KILTProtocol/ticket#777

Introduces `jsonld-signatures` `suite`s that allow verifying (and in part issuing/creating) the `proof`s that secure a KILT credential in its Verifiable Credential form using the packages `vc-js` and `jsonld-signatures`.

## How to test:
See unit tests and the package readme for examples.
- [ ] code in the readme was run and works as expected

## Checklist:


packages/vc-export/src/vc-js/suites/KiltIntegritySuite.ts
```diff
@@ -0,0 +1,122 @@
...
+      )
+    ) {
+      throw new Error(
+        "'claimHashes' and 'nonces' are required properties on 'existingProof'"
+      )
+    }
+    this.existingProof = existingProof
+  }
+
+  public async createProof(options: {
```
tjwelde, line 43: Please add a caveat in a jsdoc comment, that describes, that the vc id already has to be generated and is the same as our root hash.

tjwelde: Or, that we need an `existingProof` for it to generate a new one.

tjwelde: In general, I would just not offer creating proofs here, since it is not very feasible in the current state of the KILT project.
Let's keep the code in another branch for the future, when we are deciding to go deeper with vc.js.

tjwelde: (same with attested `createProof` and signature `createProof`)

tjwelde (changes requested):

packages/vc-export/src/vc-js/examples/example-vc.json
```diff
@@ -0,0 +1,56 @@
+{
```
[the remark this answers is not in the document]
rflechtner: this gets exported if you pass the ctype object to the exporter and is currently irrelevant to the suites because vc-js doesn't process it afaik

packages/vc-export/src/exportToVerifiableCredential.ts
```diff
@@ -25,6 +25,28 @@ import {
   VerifiableCredential,
 } from './types'
 
+const KILT_CREDENTIAL_URI_PREFIX = 'kilt:cred:'
+
+export function fromCredentialURI(credentialId: string): string {
```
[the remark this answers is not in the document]
tjwelde: @rflechtner this function still uses URI in the naming

packages/vc-export/src/vc-js/examples/example-vc.json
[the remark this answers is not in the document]
tjwelde: ok, great

tjwelde (approved):
Besides the naming change missing, you are good to go!

PR #367, opened 2021-04-07
feat: bundle sdk-js and publish in /dist

LeonFLK:
## fixes KILTProtocol/ticket#1163

Adds webpack configuration that builds sdk-js bundle in the packages /dist folder as minimized and non-minimized.
Publish with dist folder.

## How to test:
Add to html as script `  <script src="sdk-js.umd.js"></script>`

and use it via `window.sdk`

the script entry name is sdk for now, so to use the script call via window.sdk .

use @kiltprotocol/sdk-js and it's exposed functionalities.

## Checklist:

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -20,6 +20,17 @@ yarn link "@kiltprotocol/sdk-js"
 Go into your project folder and execute that second command.
 The SDK is now symlinked in your projects `node_modules` folder.
 
+## Embed bundle in HTML
+
+We include UMD bundles in our release and prerelease NPM publishes.
+They can be used to easily embed our complete SDK.
+
+```html
+<script src="https://unpkg.com/@kiltprotocol/sdk-js@/dist/sdk-js.min.umd.js"></script>
+```
+
+You can find the library on `window.sdk`, and use it completely dependency free.
+
 ## Build to see changes
 
 Note that **before you see your changes from the SDK, you have to build it**, by executing a `build`:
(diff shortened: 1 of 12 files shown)


tjwelde:
I think it is not completely clear, how to use this new bundle.
I think we should 
a) automatically put it on a CDN, as is specified in the ticket and 
b) think about having another package for the bundled version, or
c) include the minified version in our sdk-js package, but document it in the README

LeonFLK:
Totally agree, I will create a ticket to update the README, to show case how we can access the npm cdn and use the SPA.

unpkg and npmcdn automatically provide for npm published packages.

Once merged to develop we can access the bundle via  `{npmcdn.com/unpkg.com)/:package@:version/:file`
example: https://npmcdn.com/@kiltprotocol/sdk-js@0.20.0/lib/index.js

The bundles are named : sdk-js.min.umd.js and sdk-js.umd.js, do we want to stick to the sdk-js term?

tjwelde:
I don't really have a strong opinion in the naming of the bundle

LeonFLK:
We could name the bundle kilt, that would be my suggestion.

As an integrator I'm sure I'd be glad if the embedded bundles are easily distinguished.

Other than that, is the PR approved?

rflechtner:
> We could name the bundle kilt, that would be my suggestion.
> 
> As an integrator I'm sure I'd be glad if the embedded bundles are easily distinguished.
> 
> Other than that, is the PR approved?

I imagine it could also add to the confusion, especially if we decide to bundle other packages as well (core?). Maybe sdk-js is best, because then the reference is clear

LeonFLK:
Okay so for the naming of the files we stick to the implemented name.
As long as we do not release SPA bundles for different packages we want to access the bundle via the window.kilt object.

docs/development-setup.md
```diff
@@ -20,6 +20,17 @@ yarn link "@kiltprotocol/sdk-js"
 Go into your project folder and execute that second command.
 The SDK is now symlinked in your projects `node_modules` folder.
 
+## Embed bundle in HTML
```
tjwelde, line 23: Development setup is only for people working on the SDK, not _with_.

Please move these instructions into the README.md

tjwelde (changes requested):

tjwelde (approved):
LGTM

PR #370, opened 2021-04-12
feat: Request claims required Properties and approved attesters

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1199

The request-claims-for-ctype message currently only allows specifying acceptable ctypes.

To allow more use-cases, expand on it to be able to specify per ctype:

- [x] required attributes
- [x] accepted attesters

## How to test:

Added methods to verify the required properties. 

## Checklist:


packages/messaging/src/Message.utils.ts
```diff
@@ -15,12 +15,41 @@ import {
...
+ * @throws [[ERROR_CTYPE_HASH_NOT_PROVIDED]] when the properties do not match the provide [[CType]].
+ *
+ * @returns Returns the properties back.
+ */
+
+export function verifyRequiredCTypeProperties(
+  requiredProperties: string[],
+  cType: ICType
+): boolean {
+  if (!cType.hash) throw SDKErrors.ERROR_CTYPE_HASH_NOT_PROVIDED()
```
LeonFLK, line 39: Shouldn't we use our isICType typeguard instead of just checking the hash?

LeonFLK (commented):
Small nitpick, use isICType type guard instead of just checking existence of hash.
Or is this not feasible because the CType is partial, and thus not passing the type guard?

packages/messaging/src/Message.utils.ts
Dudleyneedham: I agree, I have updated it.

LeonFLK (approved):
LGTM

LeonFLK:
Integration test are expected to fail, as they are broken momentarily.

PR #371, opened 2021-04-13
refactor: encrypted message body naming

LeonFLK:
## fixes KILTProtocol/ticket#1193

Changes the property key of the encrypted message body from `message` to `ciphertext` for clarity.
https://en.wikipedia.org/wiki/Ciphertext

## Checklist:


rflechtner (approved):
LGTM

PR #372, opened 2021-04-16
feat: tx workflow revamp

LeonFLK:
## fixes KILTProtocol/ticket#1184

Kilt Object functions that prepare Transactions for the chain do not sign the SubmittableExtrinsic anymore, this has to be done by hand or with our all purpose `Blockchain.signAndSubmitTx`, with which unsigned SubmittableExtrinsic can be dispatched with additional optional resolve/reject criteria and the posibility to apply nonce collision recovery logic with the `reSign` flag.

Updated tx workflow: [Sign and Submit Workflow](https://hackmd.io/mnMZBi-wQm6bgflXkCKHbw?view)
## Checklist:

### DIFF
--- a/packages/chain-helpers/src/blockchain/Blockchain.utils.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.utils.ts
@@ -7,14 +7,12 @@
 import { SDKErrors } from '@kiltprotocol/utils'
 import { ConfigService } from '@kiltprotocol/config'
 import type {
-  IIdentity,
   ISubmittableResult,
   SubmittableExtrinsic,
   SubscriptionPromise,
 } from '@kiltprotocol/types'
 import { ErrorHandler, ExtrinsicError, ExtrinsicErrors } from '../errorhandling'
 import { makeSubscriptionPromise } from './SubscriptionPromise'
-import { getConnectionOrConnect } from '../blockchainApiConnection/BlockchainApiConnection'
 
 const log = ConfigService.LoggingFactory.getLogger('Blockchain')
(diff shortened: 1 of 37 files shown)


packages/chain-helpers/src/blockchain/Blockchain.utils.ts
```diff
@@ -142,30 +135,13 @@ async function submitSignedTxErrorMatched(
  */
 export async function submitSignedTx(
```
rflechtner, line 136: feels like this may only be relevant for our higher level functions, not for external callers? does it make sense to not export it?

packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -8,20 +8,17 @@ import { DecoderUtils } from '@kiltprotocol/utils'
...
 const log = ConfigService.LoggingFactory.getLogger('Attestation')
 
 /**
  * @param attestation
- * @param identity
  * @internal
  */
 export async function store(
-  attestation: IAttestation,
-  identity: Identity
```
rflechtner, line 24: soon this will require an identity again (a did then), but for now removing the parameter seems to be the only way

rflechtner (approved):
LGTM ! use my comments as you see fit

packages/chain-helpers/src/blockchain/Blockchain.utils.ts
LeonFLK: We call this directly from the Blockchain class when !flag, we need to export it.

PR #373, opened 2021-04-20
ci: tests on mashnet-staging

tjwelde:
## fixes KILTProtocol/ticket#1244
This PR runs integration tests / getting-started and example against the mashnet-staging network as well as the devnet

## How to test:
- Tests should run
- Devnet tests may fail

## Checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,6 +56,10 @@ jobs:
 
     needs: check_skip
 
+    strategy:
+      matrix:
+        image: ['latest-develop', 'latest-rc']
+
     steps:
       - uses: actions/checkout@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -80,7 +84,7 @@ jobs:
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           ECR_REPOSITORY: kilt/prototype-chain
-          IMAGE_TAG: latest-develop
+          IMAGE_TAG: ${{ matrix.image }}
         run: |
           yarn install --immutable
           yarn build
```
rflechtner, line 90: building twice feels completely unnecessary, but as far as I can see there is no matrix strategy for a single step only - so I guess there is no way around it. But it's another argument to finally figure out a way to build once and use the wherever required (running example, integration testing, maybe even unit testing ?)

rflechtner (approved):
LGTM. Introduces even more unnecessary builds, but this is not the place to fix that.

PR #374, opened 2021-04-20
fix: different versions of BN from Polkadot causes issues

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1211

Downgraded to version 4.11.9 to match polkadots current implementation of BN.

No changes to SDK

When updating the polkadot-js version we should take care of matching the appropriate bn.js version.

## How to test:

Run through all the tests

## Checklist:


LeonFLK (approved):
LGTM

LeonFLK:
latest-develop test are expected to fail.

PR #376, opened 2021-04-22
fix: outdated docs, unrequired identity

LeonFLK:
## fixes KILTProtocol/ticket#1249 fixes KILTProtocol/ticket#1248

Updated Docs and functions to account for removed parameter.
`Delegation.revoke` now takes only the necessary address.
More consistent tx workflow documentation and guidance.

## Checklist:

### DIFF
--- a/packages/chain-helpers/src/blockchain/Blockchain.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.ts
@@ -1,7 +1,7 @@
 /**
  * Blockchain bridges that connects the SDK and the KILT Blockchain.
  *
- * Communicates with the chain via WebSockets and can [[listenToBlocks]]. It exposes the [[submitTx]] function that performs a transaction.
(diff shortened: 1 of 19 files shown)


packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * Blockchain bridges that connects the SDK and the KILT Blockchain.
  *
- * Communicates with the chain via WebSockets and can [[listenToBlocks]]. It exposes the [[submitTx]] function that performs a transaction.
+ * Communicates with the chain via WebSockets and can [[listenToBlocks]]. It exposes the [[signTx]] function that performs the necessary tx signing.
```
rflechtner, line 4: I'd say submission is the more relevant use case? Why should we only mention signing here?

LeonFLK: The Blockchain Class doesn't really expose any submission function, the api is only used for signing.
BlockchainUtils handles dispatching.

PR #377, opened 2021-04-22
feat: adding delegation creation from method

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1118

Updating delegation to have a from method to build a delegation/node/rootNode 
Updating constructor to include `revoked` 
Adding error handling of delegation/node/rootNode 

## How to test:

Run delegation unit and integration tests.

## Checklist:


packages/core/src/delegation/Delegation.ts
```diff
@@ -36,13 +37,17 @@ export default abstract class DelegationBaseNode
...
   public constructor(
     id: IDelegationBaseNode['id'],
-    account: IDelegationBaseNode['account']
+    account: IDelegationBaseNode['account'],
+    revoked: IDelegationBaseNode['revoked']
   ) {
+    errorCheck(id, account, revoked)
     this.account = account
     this.id = id
+    this.revoked = revoked
```
rflechtner, line 50: idea: let `errorCheck` accept an `IDelegationBaseNode` as its only argument, then call `errorCheck(this)` after assignment. Think that would work? Makes more sense in my head to be able to check an `IDelegationBaseNode` object.
```suggestion
    this.account = account
    this.id = id
    this.revoked = revoked
    errorCheck(this)
```

packages/core/src/delegation/Delegation.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  } else if (!isHex(id)) {
+    throw SDKErrors.ERROR_DELEGATION_ID_TYPE()
+  }
+
+  if (!account) {
+    throw SDKErrors.ERROR_OWNER_NOT_PROVIDED
+  } else DataUtils.validateAddress(account, 'delegationNode owner')
+
+  if (typeof revoked !== 'boolean') {
+    throw SDKErrors.ERROR_NOT_FOUND('not boolean')
```
rflechtner, line 21: wouldn't this just be a `TypeError` ?

packages/core/src/delegation/Delegation.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+import { SDKErrors, DataUtils } from '@kiltprotocol/utils'
+import { isHex } from '@polkadot/util'
+
+export default function errorCheck(
+  id: IDelegationBaseNode['id'],
+  account: IDelegationBaseNode['account'],
+  revoked: IDelegationBaseNode['revoked']
+): void {
+  if (!id || typeof id !== 'string') {
+    throw SDKErrors.ERROR_NODE_QUERY(id)
```
rflechtner, line 11: seems like an odd choice to signal a missing id ?

packages/core/src/delegation/DelegationNode.ts
```diff
@@ -48,18 +61,18 @@ export default class DelegationNode extends DelegationBaseNode
...
-    super(id, account)
-    this.permissions = permissions
-    this.rootId = rootId
-    this.parentId = parentId
+  public constructor(delegationNodeInput: IDelegationNode) {
+    super(
+      delegationNodeInput.id,
+      delegationNodeInput.account,
+      delegationNodeInput.revoked
+    )
```
rflechtner, line 71: 100 % on board with changing the constructor interface the way you did, but shouldn't we also change it on the base node as well?

packages/core/src/delegation/Delegation.ts
rflechtner: update: if we would also be changing the base node constructor signature to work similar to the two child classes, we may just call the `errorCheck` on the input object

packages/core/src/delegation/DelegationRootNode.ts
```diff
@@ -19,11 +19,25 @@ import DelegationBaseNode from './Delegation'
...
+   *
+   * @param delegationRootNodeInput IDelegationBaseNode to instantiate the new delegation node from.
+   *
+   * @returns An instantiated DelegationRootNode.
+   */
+  public static async fromDelegationRootNode(
+    delegationRootNodeInput: IDelegationRootNode
+  ): Promise<DelegationRootNode> {
+    return new DelegationRootNode(delegationRootNodeInput)
+  }
```
rflechtner, line 39: why is this async - and what do we need it for?

rflechtner (commented):
Had some questions and comments; the main one being: why don't we change the `DelegationBaseNode`s constructor signature to work like the child classes?

packages/core/src/delegation/Delegation.utils.ts
Dudleyneedham: Yes, I have updated to be in line with what is happening.

packages/core/src/delegation/Delegation.utils.ts
Dudleyneedham: Do you mean something like this?

`new TypeError('Delegation revoked type is incorrect')`

packages/core/src/delegation/DelegationRootNode.ts
Dudleyneedham: Originally, I was going to have the constructor check if there is an `id` and if not generate one.

The `fromDelegationRootNode` was for pre-constructed delegations would take the id as a para, which would have checked the if the `id` was valid etc. 

With the purposed changes to construct the `id` in the mode, I removed it. 

I would say that this can be removed.

Dudleyneedham: I also removed the `fromDelegationNode` method from the `DelegationNode` class.

packages/core/src/delegation/Delegation.ts
Dudleyneedham: Not sure I understand what you mean with the second back. I have updated to accept `this`.

packages/core/src/delegation/DelegationNode.ts
Dudleyneedham: Not sure what you mean, since `super(...)` should call the error handling from the base node.

packages/core/src/delegation/Delegation.utils.ts
rflechtner: Yeah or `new TypeError('revoked is expected to be a boolean')`

packages/core/src/delegation/DelegationNode.ts
rflechtner: That is to say that I don't see a reason why we wouldn't change the `DelegationBaseNode`'s constructor to accept only a single argument as well, which is an object containing an `id`, `account` and `revoked` property.

rflechtner (approved):
LGTM

PR #378, opened 2021-04-23
feat: optional tip for signing tx

LeonFLK:
## fixes KILTProtocol/ticket#1225

Adds the ability to provide AnyNumber as tip for the validator when signing the tx.

Apart from tip, Polkadot-JS can take multiple different SignerOptions:

```
export interface SignerOptions {
    blockHash: Uint8Array | string;
    era?: IExtrinsicEra | number;
    nonce: AnyNumber | Codec;
    signer?: Signer;
    tip?: AnyNumber;
}
```
Besides the implemented optional passable options, are any other possible options of interest for us?

## Checklist:

- [ x] I have verified that the code works

### DIFF
--- a/docs/example.ts
+++ b/docs/example.ts
@@ -69,7 +69,7 @@ async function setup(): Promise<{
   const ctype = new Kilt.CType(rawCtype)
 
   // Store ctype on blockchain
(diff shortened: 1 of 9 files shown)


rflechtner (approved):
LGTM

PR #382, opened 2021-04-30
Did pallet essentials

rflechtner:
## Fixes: 
- fixes KILTProtocol/ticket#990
- fixes KILTProtocol/ticket#1176 
- fixes KILTProtocol/ticket#1177
- fixes KILTProtocol/ticket#1178

As detailed in ticket#990, this PR introduces a new package which so far only contains low-level code to create, update and delete DID records on the new DID pallet.

## How to test:
For examples, see the spec file included with the package.

## Checklist:


packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
```diff
@@ -15,40 +15,178 @@ import Blockchain from '../blockchain/Blockchain'
 let instance: Promise<Blockchain> | null
 
 export const CUSTOM_TYPES: RegistryTypes = {
+  // Runtime types
```
Diiaablo95, line 18: Can't we use the types from the [type-definition](https://github.com/KILTprotocol/type-definitions/blob/master/src/index.ts) repo here?

rflechtner: that's actually already implemented downstream, so will be merged (yeeeey, conflicts)

rflechtner: or upstream ? I dunno, it's in develop

packages/core/src/delegation/Delegation.chain.ts
```diff
@@ -24,7 +24,7 @@ export async function getAttestationHashes(
 ): Promise<string[]> {
   const blockchain = await BlockchainApiConnection.getConnectionOrConnect()
   const encodedHashes = await blockchain.api.query.attestation.delegatedAttestations<
-    Vec<H256>
+    Option<Vec<H256>>
```
Diiaablo95, line 27: I think we should replace all `H256` with `Hash`, at this point, right? At least for all things related to attestations.

rflechtner: Can do, it's nothing but an alias anyway

packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
Diiaablo95: Then perhaps you should also merge from `develop`?

packages/kilt-did/src/Did.chain.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export async function queryByDID(
+  did: IDid['identifier']
+): Promise<IDidRecord | null> {
+  // we will have to extract the id part from the did string
+  const didId = getIdentifierFromDid(did)
+  return queryById(didId)
+}
+
+export async function createCreateTx(
```
Diiaablo95, line 119: I would perhaps replace the first `create` with `generate`, so to avoid the `createCreate` naming.

Diiaablo95: Same would apply to the other functions as well.

packages/kilt-did/src/Did.chain.ts
```diff
@@ -0,0 +1,161 @@
...
+  deleteOptions: IDeleteOptions,
+  authenticationKeypair: ISigningKeyPair
+): Promise<SubmittableExtrinsic> {
+  const blockchain = await BlockchainApiConnection.getConnectionOrConnect()
+  const encoded = encodeDidDelete(blockchain.api.registry, deleteOptions)
+  const { payload, signature } = signCodec(encoded, authenticationKeypair)
+  return blockchain.api.tx.did.submitDidDeleteOperation(payload, signature)
+}
+
+export async function createAuthenticationTx(
```
Diiaablo95, line 150: Also this one, I would rename to something like `generateDidAuthenticatedTx`.

packages/kilt-did/src/Did.integration.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+    await expect(
+      BlockchainUtils.signAndSubmitTx(tx, alice, {
+        resolveOn: BlockchainUtils.IS_IN_BLOCK,
+      })
+    ).resolves.not.toThrow()
+
+    await expect(ctype.verifyStored()).resolves.toEqual(true)
+  }, 20_000)
+
+  it('no longer authorizes ctype creation after did deactivation', async () => {
```
Diiaablo95, line 189: I would call it DID deletion, not deactivation. And I would capitalize DID everywhere.

packages/kilt-did/src/Did.utils.ts
```diff
@@ -0,0 +1,160 @@
...
+    return false
+  })
+  if (!matched)
+    throw new Error(
+      'only endpoint urls starting with http/https, ftp, and ipfs are accepted'
+    )
+  return typedUrl as UrlEnum
+}
+
+export function encodeDidCreate(
```
Diiaablo95, line 74: I would rename it to `encodeDidCreationOperation`, same for the other ones.

packages/kilt-did/src/Did.utils.ts
```diff
@@ -0,0 +1,160 @@
...
+): DidDeletionOperation {
+  return new (registry.getOrThrow<DidDeletionOperation>(
+    'DidDeletionOperation'
+  ))(registry, {
+    did: didIdentifier,
+    txCounter,
+  })
+}
+
+export function encodeCallAuthentication(
```
Diiaablo95, line 149: Same renaming here.

packages/kilt-did/src/types.chain.ts
```diff
@@ -0,0 +1,132 @@
...
+  payload: Vec<u8>
+}
+
+export interface Url extends Enum {
+  isHttp: boolean
+  isFtp: boolean
+  isIpfs: boolean
+  asHttp: UrlEncoding
+  asFtp: UrlEncoding
+  asIpfs: UrlEncoding
```
Diiaablo95, line 55: Don't we have `type` and `value` here as well?

Diiaablo95: I don't see it either in other enums after, so perhaps it is just not required everywhere 😄

packages/kilt-did/src/types.chain.ts
```diff
@@ -0,0 +1,132 @@
...
+  asSr25519: Vec<u8>
+}
+
+export interface DidAuthorizedCallOperation extends Struct {
+  did: DidIdentifier
+  txCounter: u64
+  call: DidCallable
+}
+
+type DidCallable = Call
```
Diiaablo95, line 132: Maybe let's keep all the type definitions together, and the top of the file sounds like a good place to do that.

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+import { u64 } from '@polkadot/types'
+import { AnyNumber } from '@polkadot/types/types'
+import { KeyId } from './types.chain'
+
+export type KeypairType = string
+
+export interface IPublicKey {
+  publicKey: Uint8Array
+  type: KeypairType
+  // do we need the identifier?
```
Diiaablo95, line 11: If we want to delete a public key, we need its ID, so I guess the identifier is needed here.

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export type IKeyPair = IPublicKey
+
+export type ISigningKey = IPublicKey
+export type IEncryptionKey = IPublicKey
```
Diiaablo95, line 17: If we use `IEncryptionKey` and `IEncryptionKeyPair`, then I think it would make more sense to rename `IEncryptionKey` to `IEncryptionPublicKey`. For `ISigningKey`, the right terminology would be `IVerificationKey` for the public key, and `ISigningKey` for the private one.

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface IEncryptionKeyPair extends IEncryptionKey {
+  decrypt: (
+    message: Uint8Array,
+    additionalData: Record<string, unknown>
+  ) => Uint8Array | null
+}
+
+// TODO: should this use the same keys as the chain type?
+export interface KeySet {
```
Diiaablo95, line 31: Since this does not include the past attestation keys (now simple verification keys), perhaps it would be better to call it `ActiveKeySet`.

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  authentication: ISigningKeyPair
+  encryption: IEncryptionKeyPair
+  attestation?: ISigningKeyPair
+  delegation?: ISigningKeyPair
+}
+
+export interface PublicKeySet {
```
Diiaablo95, line 38: Same, something like `ActivePublicKeySet`.

packages/kilt-did/src/types.ts
Diiaablo95: I guess also to perform lookup from chain we need the ID. I don't know exactly where this `IPublicKey` would be used? With other parties? Or just internally by the DID owner?

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface KeyDetails extends TypedPublicKey {
+  id: string
+  includedAt: number
+}
+
+export interface IDidRecord {
+  did: IIdentity['address']
+  authenticationKey: KeyDetails
+  keyAgreementKeys: KeyDetails[]
```
Diiaablo95, line 58: Can this syntax also refer to sets? Or does it only represent an array? In the former case, we might want to change it with a set?

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface UrlEncoding {
+  payload: string
+}
+
+export type UrlEnum =
+  | { Http: UrlEncoding }
+  | { Ftp: UrlEncoding }
+  | { Ipfs: UrlEncoding }
+
+export interface DidSigned<PayloadType> {
```
Diiaablo95, line 80: Why not `DidSignedPayload`?

packages/kilt-did/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  payload: PayloadType
+  signature: SignatureEnum
+}
+
+export interface ICreateOptions {
```
Diiaablo95, line 85: I would rename it to `IDidCreationOptions`, for instance. Same for the others below.

Diiaablo95:
I left some comments, mostly about naming to make it easier to use the new package. I will go over once more once the conflicts have been resolved and the integration tests are running again. Good job! 💪

packages/kilt-did/src/types.chain.ts
rflechtner: that's coming from extending `Enum`, it inherits the properties defined there

packages/kilt-did/src/types.ts
rflechtner: just realised I'm not actually using this any more. So I may just remove this definition.

packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
rflechtner: I would have done it before merge, but I can do it now if you want to test

packages/kilt-did/src/types.ts
Diiaablo95: Ok, if not used we can remove it. Otherwise I think it would be better to rename it.

packages/kilt-did/src/types.chain.ts
rflechtner: I'm not sure what this refers to ? But I should probably comment on the longer term plans for these types files. The `types.ts` file is supposed to be moved to the types package (I could also do that now), while the `types.chain.ts` is tightly linked to the types-definitions repo and may live there in the future, as this makes it easier to maintain typescript interfaces along with the substrate types they represent. Additionally, there is some (currently broken) tooling to convert the substrate types into interfaces.

packages/chain-helpers/src/blockchainApiConnection/BlockchainApiConnection.ts
Diiaablo95: I think it is better just to avoid duplicating work when something in the types changes, right?

packages/kilt-did/src/types.ts
rflechtner: Could also be named `KeyRolesMap` or something similar, as it maps roles/functions/verification relationships to keys? 
Honestly, I am still not super happy with the way keys are managed here. Not only can we have multiple encryption keys, which this does not represent. It also doesn't really represent the fact very well that on chain we only have a vector of keys, with references to them identifying the roles these keys are meant to be used in.
So if you have any ideas for alternatives, let me know.

rflechtner: I guess one alternative would be to manage keys as an Array, mirroring how they are stored on chain, and annotate them with their respective roles by adding an additional "role" property where applicable: 
```js
[ 
  { id: '0x...', publicKeyHex: '0x...', type: 'x25519', role: 'KeyAgreement', includedAt: 110},
  { id: '0x...', publicKeyHex: '0x...', type: 'sr25519', role: 'Authentication', includedAt: 85},
  { id: '0x...', publicKeyHex: '0x...', type: 'sr25519', includedAt: 85}
]
```

packages/kilt-did/src/types.ts
rflechtner: this reads nicer when used: 
`DidSigned<IDidCreationOperation>`

packages/kilt-did/src/types.chain.ts
Diiaablo95: I just mean that we should group together all the `type` definitions within the same file, where possible 😄

packages/kilt-did/src/types.ts
Diiaablo95: I don't think storing keys in an array would be the most efficient solution. I think it is probably more clear in which direction to go after yesterday's discussion, and my point was more about the naming of this interface than anything else really.

packages/kilt-did/src/types.ts
Diiaablo95: Simply because we might offer O(1) retrieval of a specific key, instead of always iterating over the array looking for that very one key.

packages/kilt-did/src/types.ts
Diiaablo95: Mhh okay, agree.

packages/kilt-did/src/Did.utils.ts
```diff
@@ -0,0 +1,184 @@
...
+  if (typeof keypair === 'object') {
+    const { publicKey, type } = keypair as any
+    return publicKey instanceof Uint8Array && typeof type === 'string'
+  }
+  return false
+}
+
+export function encodeEndpointUrl(url: string): UrlEnum {
+  const typedUrl: Record<string, UrlEncoding> = {}
+  const matched = Array.from(['http', 'ftp', 'ipfs']).some((type) => {
```
Diiaablo95, line 61: I guess here we might want to perform some more serious checks, like using existing regexes to parse those addresses.

Diiaablo95:
Just a couple of minor comments more, otherwise I would say it is time to merge this PR. 💪

packages/kilt-did/src/types.ts
rflechtner: Sets in js are fairly useless for this purpose I think. There is no indexing for sets (but there is for arrays), sets are not ordered - all they offer is certainty that each value is only in there once
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

packages/kilt-did/src/types.chain.ts
rflechtner: oh, yeah, we can do that. doesn't make a difference to me

packages/kilt-did/src/Did.utils.ts
rflechtner: I disagree tbh. Reimplementing the string validation we're doing on the node feels like we'd be setting ourselves up for another maintenance nightmare where we risk that the node and sdk implementations are out of sync. And consumers of the sdk are incentivised to run their own checks by the transaction fees they waste every time they mess it up

rflechtner: the checks that are run as part of this util are also only there because the url string needs to be transformed to something like this:
```
{
  http: {
    payload: 'https://example.com'
  }
}
``` 
Could we not just have `newEndpointUrl` be a string on `DidCreationOperation` and `DidUpdateOperation` ?

packages/kilt-did/src/types.ts
rflechtner: I checked, the id is added in the inheriting interface `KeyDetails` together with the `includedAt` block number, the rationale being that the id is not really known before submitting to chain (but can be computed with tooling I added later).

packages/kilt-did/src/types.ts
rflechtner: I called the interface `PublicKeyRoleAssignment`, hope that works for you

Diiaablo95 (approved):

PR #387, opened 2021-05-06
chore: don't fail workflow on dev error

tjwelde:
## NO-TICKET

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -103,6 +103,7 @@ jobs:
     strategy:
       matrix:
         image: ['latest-rc', 'latest-develop']
+        can_fail: [false]
         include:
         - image: 'latest-develop'
           can_fail: true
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -102,7 +102,13 @@ jobs:
 
     strategy:
       matrix:
-        image: ['latest-develop', 'latest-rc']
+        image: ['latest-rc']
+        required: ['required']
```
rflechtner, line 106: I believe now that we have `continue-on-error: ${{ matrix.required == 'optional' }}`, we wouldn't even need this line, but let's not over-engineer

rflechtner (approved):
i LOOOOOVE it ❤️ 💯

.github/workflows/tests.yml
rflechtner: plus it makes the output nicer in the dashboard!

PR #393, opened 2021-05-12
feat: use types repo

rflechtner:
## fixes KILTProtocol/ticket#1307

Imports the types repo in the sdk so we don't have to copy-paste type definitions and maintain them in two different places.
Also exports the types registry used by the API which allows you to create SCALE (chain) types without connecting to the blockchain.

Currently does not provide a mechanism to choose with which types you want to use the sdk (wasn't possible before, not sure we need it). This means that the sdk would likely only ever work with the latest chain version. 

## How to test:

Run against latest develop chain. Will still fail most likely because integration is not fixed, but this shouldn't fail due to types issues.

## Checklist:

### DIFF
--- a/packages/chain-helpers/package.json
+++ b/packages/chain-helpers/package.json
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@kiltprotocol/config": "workspace:*",
+    "@kiltprotocol/type-definitions": "0.1.3",
     "@kiltprotocol/types": "workspace:*",
     "@kiltprotocol/utils": "workspace:*",
     "@polkadot/api": "^3.11.1",
--- a/packages/chain-helpers/package.json
+++ b/packages/chain-helpers/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@kiltprotocol/config": "workspace:*",
-    "@kiltprotocol/type-definitions": "0.1.3",
+    "@kiltprotocol/type-definitions": "0.1.4",
     "@kiltprotocol/types": "workspace:*",
     "@kiltprotocol/utils": "workspace:*",
     "@polkadot/api": "^3.11.1",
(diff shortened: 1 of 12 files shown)


packages/chain-helpers/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@kiltprotocol/config": "workspace:*",
+    "@kiltprotocol/type-definitions": "0.1.3",
```
Dudleyneedham, line 29:
```suggestion
    "@kiltprotocol/type-definitions": "^0.1.3",
```

Would stop us from needing to creating PRs to update it

yarn.lock
```diff
@@ -818,6 +818,7 @@ __metadata:
   resolution: "@kiltprotocol/chain-helpers@workspace:packages/chain-helpers"
   dependencies:
     "@kiltprotocol/config": "workspace:*"
+    "@kiltprotocol/type-definitions": 0.1.3
```
Dudleyneedham, line 821:
```suggestion
    "@kiltprotocol/type-definitions": ^0.1.3
```

Dudleyneedham: Same as above

Dudleyneedham (changes requested):

packages/chain-helpers/package.json
rflechtner: oh that's intended, type definitions break to easily with additional work required on the sdk. We'd have to be really careful with semversioning if we're not pinning

Dudleyneedham (approved):
LGTM

rflechtner:
using types8 (mashnet node runtime version 8) for now; this is expected to work flawlessly with the latest release

packages/chain-helpers/src/blockchainApiConnection/TypeRegistry.ts
```diff
@@ -0,0 +1,8 @@
+import { types8 as KILT_TYPES } from '@kiltprotocol/type-definitions'
+import { TypeRegistry } from '@polkadot/types'
```
Dudleyneedham, line 2:
```suggestion
import type { TypeRegistry } from '@polkadot/types'
```

Dudleyneedham (changes requested):
One minor change regarding `type`

packages/chain-helpers/src/blockchainApiConnection/TypeRegistry.ts
rflechtner: actually this is imported by value, not by type. You see below that we instantiate TypeRegistry

Dudleyneedham (approved):
LGTM

PR #394, opened 2021-05-14
Presentations

rflechtner:
## originally implemented as part of KILTProtocol/ticket#1216
now stale; shall be kept for future implementations of presentations

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:


rflechtner:
while I like the presentation class the way it is now, for me the question is still open whether we should send around `Presentation`s in our messaging, or should just add the challenge as a property on the `SubmitClaimsForCType` and `RequestClaimsForCTypes` message types, as we sign the messages anyway

PR #395, opened 2021-05-17
refactor: Removing encryption from message constructor

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1268

Separate message objects
Message constructor creates a message object. Removed encryption and changed sender from `Identity` to `IPublicIdentity`
Encrypt and decrypt are methods for the message object. Takes sender `Identity` and receiver `PublicIdentity`

## How to test:

`yarn test`

## Checklist:

### DIFF
--- a/docs/example.ts
+++ b/docs/example.ts
@@ -153,7 +153,10 @@ async function doAttestation(
   )
 
   // The message can be encrypted as follows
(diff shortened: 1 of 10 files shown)


LeonFLK (approved):
LGTM

PR #396, opened 2021-06-01
feat: fromFemto, toFemto

LeonFLK:
## fixes KILTProtocol/ticket#1241
This adds toFemtoKilt and fromFemtoKilt functions, akin to the ethjs-unit conversion functions, to the `balance.utils`.
`toFemtoKilt` supports input of `'mill' | 'bill' | 'tril'` , while the output of `fromFemtoKilt` is purely metric, as this uses polkadot's `formatBalance`.

Adds `metricPrefix` string literal union type.

`balanceNumberToString` converts input `BalanceNumber` to string representation of given number.

```javascript
toFemtoKilt(
  input: BalanceNumber,
  unit: metricPrefix = 'KILT'
): BN
```

toFemtoKilt derives input base and uses convertToTxUnit to convert to femto Kilt.

```javascript
fromFemtoKilt(
  input: BalanceNumber,
  options?: BalanceOptions
): string
```
fromFemtoKilt uses `formatKiltBalance` (using polkadots formatBalance) to first format and convert input, the number part of the resulting string is then internationalized using `Intl.NumberFormat`.

## Checklist:

### DIFF
--- a/packages/core/src/balance/Balance.utils.ts
+++ b/packages/core/src/balance/Balance.utils.ts
@@ -40,6 +40,14 @@ export const Prefixes = new Map<MetricPrefix, number>([
   ['yotta', 24],
 ])
 
+/**
+ * Uses the polkadot.js balance formatter, to convert given BN to a human readable prefixed number.
+ *
+ * @param amount BN to format.
+ * @param additionalOptions Optional formatting settings, these are defaulted to KILT specific settings.
+ * @returns String representation of the given BN with prefix and unit ('KILT' as default).
+
+ */
 export function formatKiltBalance(
   amount: BN,
   additionalOptions?: BalanceOptions
@@ -130,18 +138,24 @@ export function toFemtoKilt(
  * it's output can therefore be formatted via the polkadot formatting options.
  *
  * @param input [[BalanceNumber]] to convert from Femto Kilt.
+ * @param decimals [[number]] Set the minimum decimal places in the formatted localized output, default is 4.
  * @param options [[BalanceOptions]] for internationalization and formatting.
  * @returns String representation of the given [[BalanceNumber]] with unit und metric prefix.
  */
 export function fromFemtoKilt(
   input: BalanceNumber,
+  decimals = 4,
   options?: BalanceOptions
 ): string {
   const inputBN = new BN(balanceNumberToString(input))
   const formatted = formatKiltBalance(inputBN, options)
   const [number, ...rest] = formatted.split(' ')
+  const negative = number.startsWith('-')
   const localeNumber = new Intl.NumberFormat(options?.locale, {
-    minimumFractionDigits: 3,
-  }).format(Number(number))
-  return `${localeNumber} ${rest.join(' ')}`
+    minimumFractionDigits: 4,
+  }).format(Number(negative ? number.slice(1) : number))
+  return `${negative ? '-' : ''}${localeNumber.slice(
+    0,
+    localeNumber.length - 4 + decimals
+  )} ${rest.join(' ')}`
 }
--- a/packages/core/src/balance/Balance.utils.ts
+++ b/packages/core/src/balance/Balance.utils.ts
@@ -11,11 +11,15 @@
  */
 
 import { BN, formatBalance } from '@polkadot/util'
-import type { BalanceNumber, BalanceOptions } from '@kiltprotocol/types'
+import type {
+  BalanceNumber,
+  BalanceOptions,
+  metricPrefix,
+} from '@kiltprotocol/types'
 
 export const KILT_COIN = new BN(1)
 
-export const prefixes = new Map<string, number>([
+export const prefixes = new Map<metricPrefix, number>([
   ['femto', -15],
   ['pico', -12],
   ['nano', -9],
@@ -53,6 +57,15 @@ export function formatKiltBalance(
 export function convertToTxUnit(balance: BN, power: number): BN {
   return new BN(balance).mul(new BN(10).pow(new BN(15 + power)))
 }
+/**
+ * Safely converts the given [[BalanceNumber]] to a string, using the supplied methods,
+ * or it given a string checks for valid number representation.
+ *
+ * @param input [[BalanceNumber]] to convert.
+ * @returns String representation of the given [[BalanceNumber]].
+ * @throws On invalid number representation if given a string.
+ * @throws On malformed input.
+ */
 export function balanceNumberToString(input: BalanceNumber): string {
   if (typeof input === 'string') {
     if (!input.match(/^-?\d*\.?\d+$/)) {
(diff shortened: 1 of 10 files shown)


Dudleyneedham (dismissed):
Some minor comments

packages/core/src/balance/Balance.utils.ts
```diff
@@ -32,17 +57,98 @@ export function formatKiltBalance(
...
+        (input instanceof BigInt && input.toLocaleString)))
+  ) {
+    return input.toString()
+  }
+  throw new Error('could not convert to String')
 }
+/**
+ * Converts the given [[BalanceNumber]] to the femto Kilt equivalent.
+ *
+ * @param input [[BalanceNumber]] to convert.
```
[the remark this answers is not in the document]
LeonFLK: Yes.

packages/core/src/balance/Balance.utils.ts
```diff
@@ -32,17 +57,98 @@ export function formatKiltBalance(
...
+ * @param unit Metric prefix of the given [[BalanceNumber]].
+ * @returns Exact BN representation in femtoKilt, to use in tx and calculations.
```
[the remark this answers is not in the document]
LeonFLK: Yes.

packages/core/src/balance/Balance.utils.ts
```diff
@@ -32,17 +57,98 @@ export function formatKiltBalance(
...
+ * @throws Unknown metricPrefix, or if the input has too many decimal places for it's unit.
+ */
+export function toFemtoKilt(
+  input: BalanceNumber,
+  unit: metricPrefix = 'kilt'
```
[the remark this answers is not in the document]
LeonFLK: Yes.

packages/core/src/balance/Balance.utils.ts
```diff
@@ -32,17 +57,98 @@ export function formatKiltBalance(
...
+): string {
+  const inputBN = new BN(balanceNumberToString(input))
+  const formatted = formatKiltBalance(inputBN, options)
+  const [number, ...rest] = formatted.split(' ')
+  const localeNumber = new Intl.NumberFormat(options?.locale, {
+    minimumFractionDigits: 3,
+  }).format(Number(number))
+  return `${localeNumber} ${rest.join(' ')}`
+}
 export const TRANSACTION_FEE = convertToTxUnit(new BN(125), -9)
```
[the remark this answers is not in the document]
LeonFLK: Yes.

packages/core/src/balance/Balance.utils.ts
```diff
@@ -33,16 +58,99 @@ export function convertToTxUnit(balance: BN, power: number): BN {
...
+    if (!input.match(/^-?\d*\.?\d+$/)) {
+      throw new Error('not a string representation of number')
+    }
+    return input
+  }
+  if (
+    typeof input === 'number' ||
+    (typeof input === 'object' &&
+      ((input instanceof BN && input.toString) ||
+        (input instanceof BigInt && input.toLocaleString)))
```
tjwelde, line 83: Needs to be `typeof` to check for `bigint`

packages/core/src/balance/Balance.utils.ts
```diff
@@ -33,16 +58,99 @@ export function convertToTxUnit(balance: BN, power: number): BN {
...
+ * This function uses the polkadot formatBalance function,
+ * it's output can therefore be formatted via the polkadot formatting options.
+ *
+ * @param input [[BalanceNumber]] to convert from Femto Kilt.
+ * @param options [[BalanceOptions]] for internationalization and formatting.
+ * @returns String representation of the given [[BalanceNumber]] with unit und metric prefix.
+ */
+export function fromFemtoKilt(
+  input: BalanceNumber,
+  options?: BalanceOptions
```
tjwelde, line 138: Please add the an option to set the minimum fraction digits

tjwelde (changes requested):

packages/core/src/balance/Balance.utils.ts
LeonFLK: now checks for typeof input === 'bigint' && input.toString

packages/core/src/balance/Balance.utils.ts
LeonFLK: sadly this does not work as polkadotjs formatBalance always returns only up to 6 relevant digits, and they ceil so maximum of 3 digits are before the decimal point.

https://github.com/polkadot-js/common/blob/77289d0ec02bcb4714099593a8c3712df5f0606b/packages/util/src/format/formatBalance.ts#L68

packages/core/src/balance/Balance.utils.ts
```diff
@@ -11,16 +11,41 @@
...
+import type {
+  BalanceNumber,
+  BalanceOptions,
+  MetricPrefix,
+} from '@kiltprotocol/types'
 
 export const KILT_COIN = new BN(1)
 
+export const Prefixes = new Map<MetricPrefix, number>([
+  ['femto', -15],
```
Diiaablo95, line 23: Is there a way to avoid repeating the same names twice here and in `types/Balance.ts`?

packages/core/src/balance/Balance.utils.ts
```diff
@@ -33,16 +58,99 @@ export function convertToTxUnit(balance: BN, power: number): BN {
...
+ * or it given a string checks for valid number representation.
+ *
+ * @param input [[BalanceNumber]] to convert.
+ * @returns String representation of the given [[BalanceNumber]].
+ * @throws On invalid number representation if given a string.
+ * @throws On malformed input.
+ */
+export function balanceNumberToString(input: BalanceNumber): string {
+  if (typeof input === 'string') {
+    if (!input.match(/^-?\d*\.?\d+$/)) {
```
Diiaablo95, line 74: I guess this regex accepts also a number like `-.1`. Is this allowed? Or should also the expression before the dot be `\d+`?

LeonFLK: Yes you are correct, this accepts -.2 as -0.2. This behavior is insprired (copied) from the web3 ethjs.

Diiaablo95: I haven't seen any test cases about that tho, so it might be good to add one for positive and one for negative at least 😊

packages/core/src/balance/Balance.utils.ts
LeonFLK: I would love to have that, but I think this wouldn't work with the compile/run time type existence in TS/JS.

If you have a suggestion, I'm more than happy to adopt that!

Diiaablo95: Yeah I guess the only solution would be dynamic lookup from the entries of this map for anything that should be `MetricPrefix`.  I think Raphael is the wizard of those things, so I won't even try here 😂

packages/core/src/balance/Balance.utils.ts
LeonFLK: I implemented negative and positive tests for that, except for the case of it being actually -.2.

So thanks for bringing that up!

packages/core/src/balance/Balance.utils.ts
LeonFLK: That'd be awesome, but maybe a lil itsybitsy tiny bit overkill for a measly 20ish elements. If your type list is 100s long, then that would be a must 😂

Diiaablo95 (commented):
My review is probably not 100% authoritative, so I'll limit myself to positively comment this PR 🤓

LeonFLK:
`fromFemtoKilt` now returns up to 4 decimal places, adjusted with optional argument.

The return value also is not rounded up anymore, since that might give a false impression of how much spending power there is.

packages/core/src/balance/Balance.utils.ts
```diff
@@ -33,8 +66,97 @@ export function convertToTxUnit(balance: BN, power: number): BN {
...
+  options?: BalanceOptions
+): string {
+  const inputBN = new BN(balanceNumberToString(input))
+  const formatted = formatKiltBalance(inputBN, options)
+  const [number, ...rest] = formatted.split(' ')
+  const negative = number.startsWith('-')
+  const localeNumber = new Intl.NumberFormat(options?.locale, {
+    minimumFractionDigits: 4,
+  }).format(Number(negative ? number.slice(1) : number))
+  return `${negative ? '-' : ''}${localeNumber.slice(
```
tjwelde, line 158: `slice` can take a negative number as second parameter, than you would not have to use `localeNumber.length`

tjwelde (approved):
Little nitpick, but  LGTM

packages/core/src/balance/Balance.utils.ts
LeonFLK: Not really possible here, as in the case of decimals equaling 4, calling slice(0,0) returns empty string ''

I could of course concat a space to be able to offset the second parameter to max -1

PR #397, opened 2021-06-02
chore: update polkadot dependencies

rflechtner:
## fixes KILTProtocol/ticket#1283

Updates dependencies from the `@polkadot` scope to version ^4 which was required for the next v0.9.0 Polkadot/Kusama chain upgrades where parachains are used according to their [release notes](https://github.com/polkadot-js/api/releases/tag/v4.9.1).
This required updating typescript to version ^4 as well, and bumps node requirements to >= 14. Github actions are now run against node v14, additionally running unit tests against node v16 (soon to be LTS).  
With the update we also saw the ErrorHandler implementation breaking, which required some changes there. 

## How to test:
Integration tests and unit tests should cover a lot of ground; feel free to run additional tests if you can come up with something.

## Checklist:


packages/chain-helpers/src/errorhandling/ErrorHandler.spec.ts
```diff
@@ -24,20 +23,17 @@ describe('ErrorHandler', () => {
         },
       },
     }
-    const submittableResult: ISubmittableResult = {
-      // @ts-ignore
-      events: [evtRecord],
-    }
+    const submittableResult = {
+      events: ([evtRecord] as unknown) as EventRecord[],
```
tjwelde, line 27: a lot of type assertions here. Is there a way to make it work without enforcing the types?

packages/core/src/requestforattestation/RequestForAttestation.ts
```diff
@@ -206,7 +206,7 @@ export default class RequestForAttestation implements IRequestForAttestation {
    * ```
    */
   public removeClaimOwner(): void {
-    delete this.claim.owner
+    delete (this.claim as Partial<IClaim>).owner
```
tjwelde, line 209: kinda hacky (though I know that the deletion doesn't fit with the defined type of `this.claim`). What about the `@ts` comments you did in the test files. Is there one, where we can say: "just let it be"?

jest.config.js
```diff
@@ -1,10 +1,11 @@
 module.exports = {
-  preset: 'ts-jest',
+  preset: 'ts-jest/presets/js-with-ts',
```
tjwelde, line 2: why is this necessary?

tjwelde (commented):
Looks good overall, just some questions and nitpicks

packages/chain-helpers/src/errorhandling/ErrorHandler.spec.ts
rflechtner: `@ts-ignore` as we had it before :D

rflechtner: this is testing, right, so we're using a mocked EventRecord. Could potentially construct the real thing using the TypeRegistry, but is it worth the trouble?

packages/core/src/requestforattestation/RequestForAttestation.ts
rflechtner: it was `@ts-ignore` before; if you like that better, I can revert. We have a linter rule against its usage though, so we need to escape that as well.

jest.config.js
rflechtner: this transpilesjs files, including polkadot dependencies, to CommonJS syntax. This is because jest can't really handle esm yet

packages/core/src/requestforattestation/RequestForAttestation.ts
tjwelde: I would rather want to have a warning for it.
So if the linting is configured to show it as an error, let's reconfigure it to list it as warning.
Because it is actually not the correct way to do it and we will need the reminder.

packages/chain-helpers/src/errorhandling/ErrorHandler.spec.ts
tjwelde: oh yeah, sure. You are right. Not worth the trouble

tjwelde (approved):
LGTM

PR #398, opened 2021-06-04
feat: createPresentation on AttestedClaim

rflechtner:
## fixes KILTProtocol/ticket#1216
Removes `Credential` from the actors package and instead moves the `createPresentation` method to the `AttestedClaim` in core.
Deprecates removal of the claim owner - this produces unverifiable credentials because the owner public key is required for signature verification.

## How to test:
unit tests have also been moved

## Checklist:


tjwelde (approved):
LGTM

PR #401, opened 2021-06-07
chore: use types 10

rflechtner:
## fixes KILTProtocol/ticket#1375
Updates the type definitions

## How to test:
Should allow connecting to the node even if tests still fail (at least one integration test passing)

## Checklist:


tjwelde (approved):
LGTM

PR #402, opened 2021-06-28
docs: add license to all source files and packages

rflechtner:
## fixes KILTProtocol/ticket#1432
This PR introduces two changes to how we handle the licenses included with the SDK code; it does not change the licensing conditions themselves.

1. This makes sure that a copy of the LICENSE file in the root of the repository is included with all packages published to npm. It does so by copying the LICENSE file to each sub-package as a pre-publish step. 
2. It additionally introduces a linter rule which checks for a licensing disclaimer at the top of each source file. You can automatically add this disclaimer to files where it is missing using the command `yarn lint:fix`.  Only `.ts` files are considered.

## How to test:
You can try running the linting after removing or altering a license header.

## Checklist:

### DIFF
--- a/.eslintrc-jsdoc.json
+++ b/.eslintrc-jsdoc.json
@@ -4,6 +4,7 @@
     "prettier/prettier": ["warn", {
       "semi": true
     }],
-    "@typescript-eslint/no-unused-vars": "off"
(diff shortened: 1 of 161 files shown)


tjwelde (approved):
LGTM

Issue #403, opened 2021-07-13
vc-export: missing import of types


jrhender:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@kiltprotocol/vc-export@0.22.0` for the project I'm working on.

The `vc-export` `index.js` seems to export `types` but `types` isn't imported in order to be exported. I received the following error at runtime:
```
Cannot start the application. ReferenceError: types is not defined
    at Object.<anonymous> (/Users/john/ew/flex-backend/node_modules/@kiltprotocol/vc-export/lib/index.js:13:5)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
```

If it would be helpful to have reproduction example, let me know.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@kiltprotocol/vc-export/lib/index.js b/node_modules/@kiltprotocol/vc-export/lib/index.js
index 58239eb..5cb4eb2 100644
--- a/node_modules/@kiltprotocol/vc-export/lib/index.js
+++ b/node_modules/@kiltprotocol/vc-export/lib/index.js
@@ -3,6 +3,7 @@ Object.defineProperty(exports, "__esModule", { value: true });
 const tslib_1 = require("tslib");
 const verificationUtils_1 = tslib_1.__importDefault(require("./verificationUtils"));
 const presentationUtils_1 = tslib_1.__importDefault(require("./presentationUtils"));
+const types = tslib_1.__importDefault(require("./types"));
 const exportToVerifiableCredential_1 = require("./exportToVerifiableCredential");
 const vcjsSuites = tslib_1.__importStar(require("./vc-js"));
 exports.default = {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

rflechtner:
In the source, types is imported but this is done via a type import, so it is stripped from the resulting `index.js`. I think our mistake was to then export it as part of the default export. We'll look into it.

rflechtner:
closed by #404

PR #404, opened 2021-07-21
fix: vc-export documentation & exports

rflechtner:
## fixes https://github.com/KILTprotocol/sdk-js/issues/403
- Improves documentation
- Adds possibility to create presentations with all attributes
- fixes type export
- fixes kilt VC context

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,5 +60,5 @@
     "typedoc-plugin-external-module-name": "^4.0.6",
     "typescript": "4.0"
   },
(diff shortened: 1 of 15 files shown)


rflechtner:
importing types with `import type { types } from '@kiltprotocol/vc-export'` works just fine, I guess we can leave it like that?

tjwelde (approved):
LGTM

PR #407, opened 2021-08-19
feat: re-add actors

rflechtner:
## companion to https://github.com/KILTprotocol/sdk-js/pull/405
This reverts the removal of the actors package which happened in https://github.com/KILTprotocol/sdk-js/pull/405 to simplify did integration.
The actors package will have to be largely refactored, which shall happen on this branch. This PR prepares this.

## How to test:
tbd.

## Checklist:


tjwelde:
I think we have moved on too much. Can we close this?

rflechtner:
Plus we are going more low-level rather than high-level, so yeah, will probably never be revived. Closing.

PR #408, opened 2021-09-02
feat: light (offchain) DIDs

Diiaablo95:
## fixes KILTProtocol/ticket#1531
This PR adds support for light (or off-chain) DIDs. The new class of DIDs are still compliant with the same `IDidDetails` interface, so they are completely transparent to the users. A light DID is created by specifying an authentication key (which can be `ed25519`, `sr25519`, or `ecdsa`), and optionally an encryption key and some service endpoints. The latter two, if present, are CBOR serialised and Base64 encoded and are added as part of the DID string itself.

This PR also changes the structure of DIDs in general, as support for versioning has been added. An example of a full DID is the following: `did:kilt:v1:4r1WkS3t8rbCb11H8t3tJvGVCynwDXSUBiuGB6sLRHzCLCjs`, while a light DID that only specifies an authentication key would have the following structure: `did:kilt:light:v1:004r1WkS3t8rbCb11H8t3tJvGVCynwDXSUBiuGB6sLRHzCLCjs`. The first two `00` in the identifier represent the type of the authentication key that has been encoded to be the DID identifier.

### EDIT AFTER PR REVIEW

DID version 1 skip the `v1` bit in the URI, so a full DID v1 is `did:kilt:4r1WkS3t8rbCb11H8t3tJvGVCynwDXSUBiuGB6sLRHzCLCjs` while a light DID v1 is `did:kilt:light:004r1WkS3t8rbCb11H8t3tJvGVCynwDXSUBiuGB6sLRHzCLCjs`.

## How to test:

Run `yarn test`. All unit tests and integration tests are passing (for Node 14). I did not add much coverage to the existing test cases as I think that is out of the scope for this PR. Next PR could include updating the getting-started guides and examples with the latest version (that uses DIDs everywhere).

## Checklist:


Diiaablo95:
@tjwelde I think this PR could be reviewed now. The getting started and examples are clearly still failing (probably the scope of another PR). Tests on Node 14 are failing because of the lack of extensive coverage. For Node 16 it seems to be the very same problem that has been going on for a while. Integration tests with master are obviously failing, while the ones with develop have mostly gone through, with some timeout errors for some balance-specific stuff which I do not think is to be taken into consideration.

packages/kilt-did/src/Did.utils.ts
```diff
@@ -367,3 +447,51 @@ export function writeDidfromIdentity(
...
+
+export async function signWithDid(
+  toSign: Uint8Array | string,
+  did: IDidDetails,
+  signer: KeystoreSigner,
+  whichKey: KeyRelationship | IDidKeyDetails['id']
+): Promise<{ keyId: string; alg: string; signature: Uint8Array }> {
+  let key: IDidKeyDetails | undefined
+  if (Object.values(KeyRelationship).includes(whichKey as KeyRelationship)) {
+    ;[key] = did.getKeys(KeyRelationship.authentication)
```
tjwelde, line 473: C: I would prefer
```suggestion
    key = did.getKeys(KeyRelationship.authentication)[0]
```

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      encryptionKey: encryptionDidKeyDetails,
+      services,
+    }
+
+    const encodedAdditionalDetails = serializeAndEncodeAdditionalLightDidDetails(
+      didCreationDetails
+    )!
+
+    const did = new LightDidDetails(didCreationDetails)
+    expect(did.did).toEqual(testDid.concat(':', encodedAdditionalDetails))
```
tjwelde, line 75: S: Since this shouldn't break in the future, I think we should test it against the final string instead of relying on the `serializeAndEncodeAdditionalLightDidDetails` function, especially since it is used in the constructor for `LightDidDetails`

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      authenticationKey: authenticationDidKeyDetails,
+      encryptionKey: encryptionDidKeyDetails,
+    }
+
+    const encodedAdditionalDetails = serializeAndEncodeAdditionalLightDidDetails(
+      didCreationDetails
+    )!
+
+    const did = new LightDidDetails(didCreationDetails)
+    expect(did.did).toEqual(testDid.concat(':', encodedAdditionalDetails))
```
tjwelde, line 94: S: Same here

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      authenticationKey: authenticationDidKeyDetails,
+      services,
+    }
+
+    const encodedAdditionalDetails = serializeAndEncodeAdditionalLightDidDetails(
+      didCreationDetails
+    )!
+
+    const did = new LightDidDetails(didCreationDetails)
+    expect(did.did).toEqual(testDid.concat(':', encodedAdditionalDetails))
```
tjwelde, line 121: S: Same here

packages/kilt-did/src/DidDetails/LightDidDetails.ts
```diff
@@ -0,0 +1,136 @@
...
+    let did = getKiltDidFromIdentifier(
+      id,
+      'light',
+      LightDidDetails.LIGHT_DID_VERSION
+    )
+    if (encodedDetails) {
+      did = did.concat(':', encodedDetails)
+    }
+
+    super(
```
tjwelde, line 62: really like the abstraction with the `DIDDetails` parent class

packages/kilt-did/src/DidDetails/LightDidDetails.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
+import { encode as cborEncode, decode as cborDecode } from 'cbor'
+import type { LightDidDetailsCreationOpts } from './LightDidDetails'
+
+const ENCRYPTION_KEY_MAP_KEY = 'encryptionKey'
+const SERVICES_MAP_KEY = 'services'
```
tjwelde, line 12: S: we might be able to save some bytes, by using shorter keys here

packages/kilt-did/src/DidResolver/DefaultResolver.ts
```diff
@@ -44,44 +63,142 @@ async function detailsFromIdentifier(
...
+        return details
+      }
+
+      // Otherwise, return either the key or the service endpoints referenced by the URI.
+      return details?.getKey(didUri) || details?.getService(didUri) || null
+    }
+    case 'light': {
+      // In light DIDs the key type encoding (first two chars) is part of the identifier.
+      // We are sure the URI follows the expected structure as it has been checked in `parseDidUrl`.
+      const keyTypeEncoding = identifier.substring(0, 2)
```
tjwelde, line 115: S: Since we have a function which builds a `FullDidDetails` object above, I think it would be good to have something similar here. Like a `buildLightDetailsFromIdentifier` function e.g.

tjwelde (changes requested):
Looks pretty good I must say ;)

S: I would leave the `v1` part out for the first version, just because it looks nicer ;)

packages/vc-export/src/exportToVerifiableCredential.spec.ts
```diff
@@ -1,447 +0,0 @@
-/**
```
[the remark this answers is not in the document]
Diiaablo95: I removed it because with the changed DID structure it was producing different values and I forgot to generate the new ones. But removing the version number from DID v1 makes them work again.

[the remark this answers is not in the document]
Diiaablo95: Fixed in https://github.com/KILTprotocol/sdk-js/pull/408/commits/18dfe0b57dd7f164ded8675d24f1da5eb94c23ac

packages/kilt-did/src/Did.utils.ts
Diiaablo95: Fixed in https://github.com/KILTprotocol/sdk-js/pull/408/commits/ed4d9c3448b4bfc329e87bdd5930b88f88287e64

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
Diiaablo95: Fixed in https://github.com/KILTprotocol/sdk-js/pull/408/commits/ed4d9c3448b4bfc329e87bdd5930b88f88287e64.

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
Diiaablo95: Same as above.

packages/kilt-did/src/DidDetails/LightDidDetails.spec.ts
Diiaablo95: Same as above.

packages/kilt-did/src/DidDetails/LightDidDetails.utils.ts
Diiaablo95: Fixed in https://github.com/KILTprotocol/sdk-js/pull/408/commits/ed4d9c3448b4bfc329e87bdd5930b88f88287e64.

packages/kilt-did/src/DidResolver/DefaultResolver.ts
Diiaablo95: Fixed in https://github.com/KILTprotocol/sdk-js/pull/408/commits/ed4d9c3448b4bfc329e87bdd5930b88f88287e64.

tjwelde (approved):
LGTM

PR #409, opened 2021-09-08
doc: add README for did package

Diiaablo95:
## fixes KILTProtocol/ticket#1560

Add README for DIDs + fix error when bundling the SDK for use in the web.


Diiaablo95:
@arty-name I cannot add you as reviewer, but I would say that a first final draft for the DID README is now ready.

LeonFLK:
LGTM for the bundle issues

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+    publicKey: authenticationKeyPublicDetails.publicKey,
+    type: authenticationKeyPublicDetails.alg,
+  },
+})
+// Will print `did:kilt:4sxSYXakw1ZXBymzT9t3Yw91mUaqKST5bFUEjGEpvkTuckar`.
+console.log(did)
+
+// Submit the DID creation tx to the KILT blockchain after signing it with the KILT account specified.
+await BlockchainUtils.signAndSubmitTx(submittable, aliceKiltAccount, {
+  resolveOn: BlockchainUtils.IS_IN_BLOCK,
```
arty-name, line 184: I’ve seen that much of the documentation uses `resolveOn: BlockchainUtils.IS_IN_BLOCK`, however the recommendation for the Sporran team was to use `resolveOn: BlockchainUtils.IS_FINALIZED`. Should we also recommend that to other developers?

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+})
+
+const fullDid = (await DefaultResolver.resolveDoc(did)) as FullDidDetails
```
arty-name, line 187: First, `FullDidDetails` is not imported in this example. However, my suggestion is to remove `as FullDidDetails` from here completely. I assume the reason for it was that resolveDoc returns `FullDidDetails | null`, and I want to challenge that as well. Do you think that "DID not found" is an error deserving a throw?

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+import {
+  DefaultResolver,
+  DemoKeystore,
+  DidUtils,
+  SigningAlgorithms
+} from '@kiltprotocol/did'
+import { KeyRelationship } from '@kiltprotocol/types'
+
+// Initialise connection to a KILT blockchain node.
+await Kilt.init({ address: 'ws://127.0.0.1:9944' })
```
arty-name, line 149: I guess people trying to write some code example would like a template that works for them. Do we expect they went beyond getting the package from NPM and actually started a local Kilt blockchain? If not, should the template rather reference one of the public endpoints?

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+const encryptionSeed = '0x987654321'
+
+// Ask the keystore to generate a new keypar to use for encryption.
+const encryptionKeyPublicDetails = await keystore.generateKeypair({
+  seed: encryptionSeed,
+  alg: EncryptionAlgorithms.NaclBox,
+})
+
+// Define the set of URLs pointing to the external services exposed by the DID subject.
+// All URLs must point to the same resource, which must match the provided hash.
```
arty-name, line 219: Does our code actually verify that the resource is the same and matches the hash? As I was trying to read the requirements, I was really confused where they come from? My guess is they are a small part of "how to make a valid DID" knowledge. Should we rather refer to some document on that topic, and skip this detail here? To me it looks unnecessary and confusing in this place.

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+
+// Ask the keystore to generate a new keypar to use for authentication.
+const authenticationKeyPublicDetails = await keystore.generateKeypair({
+  seed: authenticationSeed,
+  alg: SigningAlgorithms.Ed25519,
+})
+
+// Generate the DID-signed creation extrinsic.
+// The extrinsic is unsigned and contains the DID creation operation signed with the DID authentication key.
+const { submittable, did } = await DidUtils.writeDidfromPublicKeys(keystore, {
```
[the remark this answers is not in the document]
arty-name: I think this is not updated everywhere…

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+  keystore,
+  {
+    [KeyRelationship.authentication]: {
+      publicKey: authenticationKeyPublicDetails.publicKey,
+      type: authenticationKeyPublicDetails.alg,
+    },
+    [KeyRelationship.keyAgreement]: {
+      publicKey: encryptionKeyPublicDetails.publicKey,
+      // The chain-supported type is called differently than the type supported in the keystore
+      type: 'x25519',
```
arty-name, line 241: If we control both the chain and the store, could we align the names? Or at least add a synonym?

arty-name (commented):
I think a section on consuming the DIDs is also missing. Once I have a DID coming from somewhere, what do I do with it? Should I care whether it is light or full? How do I use its keys? Stating the answers explicitly would likely benefit a developer new to the topic.

packages/kilt-did/README.md
LeonFLK: IS_IN_BLOCK would be prone to consensus errors afaik (correct me if im wrong)
For the extension or other stuff where you want to ensure that the tx is finalized, use IS_FINALIZED
Tho ofc IS_FINALIZED takes 2 blocks longer

packages/kilt-did/README.md
```diff
@@ -3,6 +3,320 @@
...
+// Initialise connection to a KILT blockchain node.
+await Kilt.init({ address: 'ws://127.0.0.1:9944' })
+
+// Generate the KILT account that will submit the DID creation tx to the KILT blockchain.
+// It must have enough funds to pay for the tx execution fees.
+const aliceKiltAccount = new Keyring({
+  type: 'ed25519',
+  // KILT has registered the ss58 prefix 38
+  ss58Format: 38,
+}).createFromUri('//Alice')
```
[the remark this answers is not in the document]
Diiaablo95: Yes I know, but this is probably the scope of a different PR, where we might want to add a conveniency layer on top of this...

packages/kilt-did/README.md
Diiaablo95: I think this is made clear (or should be made clear) in the doc somewhere else. And I also think that for demo purposes such as a README for DIDs, using `IS_IN_BLOCK` would allow for faster testing.

packages/kilt-did/README.md
Diiaablo95: Thanks for pointing this out. Since `fullDid` is anyway not used anywhere else, it makes sense to remove `as FullDidDetails`. It is up to the user to decide what to do with a potentially `null` value.

packages/kilt-did/README.md
Diiaablo95: That part, relative to the services resolver, is not yet implemented in the SDK, and will be the next big chunk of work for our next sprint. Nevertheless, I agree that is probably too much information to put in here 👍

packages/kilt-did/README.md
Diiaablo95: As the keystore is developed and managed by developers, I am exposing just a utility function that can be called on the DemoKeystore to get the right key type. Otherwise semantically the keystore needs more information about a key than the chain does, so it is not possible to call them the same.

packages/kilt-did/README.md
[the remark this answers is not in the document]
arty-name: I agree, was just sharing my feelings

packages/kilt-did/README.md
arty-name: Good documentation does not require you to read the entirety of it to make a simple thing right. Currently, this example is a trap. I see the point about quicker testing, but it should rather be done the other way: correct code in the example, comment about speeding up somewhere. Or declaring above: `const resolveOn = process.env.NODE_ENV === 'production' ? BlockchainUtils.IS_FINALIZED : BlockchainUtils.IS_IN_BLOCK`

packages/kilt-did/README.md
arty-name: Do you think that `resolveDoc` should throw an error instead of returning null?

packages/kilt-did/README.md
arty-name: Since it would be irrelevant for the actual production code, I suggest either removing this comment or modifying it to make clear this is only about the demo keystore
```suggestion
      // The chain-supported type is called differently than the type supported in the demo keystore
      type: 'x25519',
```

packages/kilt-did/README.md
Diiaablo95: `resolveDoc` does return `Promise<IDidDetails | null>` already. And I think that is the correct way. Only in the example I was casting it because we are sure the DID exists as it was created with no issues. But I fixed that.

arty-name (approved):

packages/kilt-did/README.md
```diff
@@ -3,6 +3,339 @@
...
+
+// Ask the keystore to generate a new keypair to use for authentication with the generated seed.
+const authenticationKeyPublicDetails = await keystore.generateKeypair({
+  alg: SigningAlgorithms.Ed25519,
+  seed: authenticationSeed,
+})
+
+// Create a light DID from the generated authentication key.
+const lightDID = new LightDidDetails({
+  authenticationKey: {
```
tjwelde, line 59: C: This could maybe in the future just take the `authenticationKeyPublicDetails` and extract the information itself

packages/kilt-did/README.md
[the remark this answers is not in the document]
tjwelde: I also agree, there should be some layer on top, which makes this all a bit easier to use

packages/kilt-did/README.md
```diff
@@ -3,6 +3,339 @@
...
+
+// Ask the keystore to generate a new keypar to use for authentication.
+const authenticationKeyPublicDetails = await keystore.generateKeypair({
+  seed: authenticationSeed,
+  alg: SigningAlgorithms.Ed25519,
+})
+
+// Generate the DID-signed creation extrinsic.
+// The extrinsic is unsigned and contains the DID creation operation signed with the DID authentication key.
+const { extrinsic, did } = await DidUtils.writeDidFromPublicKeys(keystore, {
```
tjwelde, line 183: C: It is weird to me, that for the off-chain DID, we use `new LightDidDetails`, but for the on-chain DID, we write it directly to the chain via the DidUtils and resolve it to get the `FullDidDetails`.

packages/kilt-did/README.md
```diff
@@ -3,6 +3,339 @@
...
+)
+```
+
+## Deleting a full DID
+
+Once not needed anymore, it is recommended to remove the DID details from the KILT blockchain. The following snippet shows how to do it:
+
+```typescript
+// Create a DID deletion operation.
+const didDeletionExtrinsic = await DidChain.getDeleteDidExtrinsic()
```
tjwelde, line 321: C: these function names are hard to read. Might it be better to just have `deleteDidExtrinsic()`?

tjwelde: Same with the `getSetKeyExtrinsic()`

tjwelde (approved):
Mainly comments for future improvements, a question and a suggestion.

packages/kilt-did/README.md
LeonFLK: I have a bit about it in my tx submission doc ready for review here : https://github.com/KILTprotocol/ticket/issues/1191

This doc is for Majella and her documentation efforts.

packages/kilt-did/README.md
arty-name: Maybe even `DidChain.extrinsics.delete(), DidChain.extrinsics.setKey()`?

packages/kilt-did/README.md
```diff
@@ -3,6 +3,339 @@
...
+
+An example of a light KILT DID is the following:
+
+```
+did:kilt:light:014sxSYXakw1ZXBymzT9t3Yw91mUaqKST5bFUEjGEpvkTuckar
+```
+
+Beyond the standard prefix `did:kilt:`, the `light:` component indicates that this DID is a light DID, hence it can be resolved and utilized offline.
+
+Light DIDs optionally support the specification of an **encryption key** (of one of the supported key types) and **service endpoints**, which are serialised and encoded and added at the end of the DID, like the following:
```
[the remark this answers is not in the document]
Diiaablo95: Yes I guess this is a more future-proof way, right? We don't have to change it again when/if we will support other keys as well.

packages/kilt-did/README.md
```diff
@@ -3,6 +3,339 @@
...
+  LightDidDetails,
+  SigningAlgorithms,
+} from '@kiltprotocol/did'
+import type { IServiceDetails } from '@kiltprotocol/types'
+
+// Instantiate the demo keystore.
+const keystore = new DemoKeystore()
+
+// Generate seed for the authentication key/
+const authenticationSeed = '0x123456789'
```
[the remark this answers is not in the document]
Diiaablo95: Yes we also expose the UUID package from `@kiltprotocol/utils`. I will add a link to that.

packages/kilt-did/README.md
Diiaablo95: I agree. Or even just a KILT account, if we re-introduce that for blockchain addresses.

packages/kilt-did/README.md
Diiaablo95: Yes, for me as well. The thing is that the `FullDidDetails` constructor is really just a wrapper around chain data (e.g., the keys it takes in input contain info about the block at which the key was created). In the refactoring, we should probably make the constructor async (if that is allowed) and do both creation and writing on the blockchain at once. And keep this constructor for when we retrieve stuff from the blockchain.

PR #414, opened 2021-09-20
feat: support for DID migrations

Diiaablo95:
## fixes KILTProtocol/ticket#1554

The PR allows for a light DID to be upgraded to a full DID by writing on the KILT blockchain a DID with the same identifier as the light one. Furthermore, attested claim presentation verification has been modified to only accept signatures from a valid DID if a DID has been migrated since the claim was initially created and included in an attestation.

## Checklist:


packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -94,43 +93,4 @@ export class LightDidDetails extends DidDetails {
       this.keyRelationships.keyAgreement = [`${this.didUri}#encryption`]
     }
   }
-
```
Diiaablo95, line 97: Removed as I realized they are already exposed by the `DidDetails` abstract superclass.

packages/did/src/DidResolver/DefaultResolver.ts
```diff
@@ -215,7 +231,7 @@ export async function resolveKey(
         throw SDKErrors.ERROR_INVALID_DID_FORMAT(didUri)
       }
       // The fragment includes the '#' symbol which we do not need
-      return resolvedDetails.getKey(fragment.substring(1)) || null
+      return resolvedDetails.details.getKey(didUri) || null
```
arty-name, line 234: The comment above is not needed anymore

tjwelde (commented):
Some questions below

tjwelde (approved):
LGTM

PR #419, opened 2021-10-05
fix: disconnect fix

Dudleyneedham:
## fixes https://github.com/KILTprotocol/ticket/issues/1625

fix for the disconnect due to old WebSocket instances 

## Checklist:


Diiaablo95 (approved):
LGTM!

PR #423, opened 2021-10-20
feat: @cfworker instead of ajv schema validator

tjwelde:
## fixes KILTProtocol/ticket#1602
This PR replaces `ajv` with `@cfworker/json-schema`, because it doesn't use `eval` internally.

## How to test:
- run the tests

## Checklist:

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -26,7 +26,15 @@ module.exports = {
     '!**/kilt/*',
     '!**/blockchainApiConnection/*',
     '!**/types/**/*',
-    '!**/SDKErrors.ts'
+    '!**/SDKErrors.ts',
+    '!**/*.chain.ts',
+    // TODO: write tests for these files and remove here
+    '!did/src/index.types.ts',
+    '!did/src/Did.chain.ts',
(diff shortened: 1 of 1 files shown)


packages/core/src/quote/Quote.spec.ts
```diff
@@ -42,7 +42,7 @@ describe('Claim', () => {
   let attesterIdentity: IDidDetails
   let keystore: DemoKeystore
   let invalidCost: ICostBreakdown
-  let date: Date
```
LeonFLK, line 45: Why switch to string from date type now, does @cfworker work nicer with it?

packages/types/src/CType.ts
```diff
@@ -12,12 +12,21 @@ import type { IDidDetails } from './DidDetails'
...
  */
 
+export type InstanceType =
+  | 'array'
+  | 'boolean'
+  | 'integer'
+  | 'null'
+  | 'number'
+  | 'object'
+  | 'string'
```
LeonFLK, line 22: nice

PR #424, opened 2021-10-22
test: test the sdk bundle

LeonFLK:
## fixes KILTProtocol/ticket#1284

Lets see.......
playwright e2e testing the bundle in html.

Adds pipeline to test bundle integrity against develop and latest rc of the node.
If the integrations tests fail for a specific node version, this will fail as well.

This has to be manually updated when introducing breaking changes to the api or node.

## Checklist:


Diiaablo95 (approved):
LGTM! Integration tests are meant to be failing as they are fixed on `master` right now. I would wait to merge this PR until we merge back from `master` after we release 0.26.0

tests/bundle-test.js
```diff
@@ -87,17 +90,17 @@ async function runAll() {
...
-      type: 'X25519',
+      type: 'x25519',
     },
   }
   const testDid = Did.LightDidDetails.fromDetails(didCreationDetails)
   if (
-    testDid.did !==
-    `did:kilt:light:01${address}:z1Ac9CMtYCTRWjetJfJqJoV7FcPDD9nHPHDHry7t3KZmvYe1HQP1tgnBuoG3enuGaowpF8V88sCxytD7HjY6Doe`
+    testDid.uri !==
+    `did:kilt:light:01${address}:z1Ac9CMtYCTRWjetJfJqJoV7FcPDD9nHPHDHry7t3KZmvYe1HQP1tgnBuoG3enuGaowpF8V88sCxytDPDy6ZxhW`
```
tjwelde, line 103: This looks like a breaking change to me. The light did encoded stuff should not change after the last commits on develop, no? @ntn-x2 ?

tjwelde: After research it turns out, that this version is the correct and non-breaking one

PR #430, opened 2021-11-02
feat: add querying support for deposits and deleted DIDs

Diiaablo95:
# Add querying support for deposits and deleted DIDs

This PR adds:
- support to query the amount of KILTs needed to deposit before creating an attestation, a delegation, or a DID
- support to query all the deleted DID identifiers and also to check whether a given DID identifier has been deleted (or blacklisted).

For both features, integration tests have been added.

## Linked components

- Node commit with which this PR has been tested to work: https://github.com/KILTprotocol/mashnet-node/commit/2e937de64c2c0979d2f1b74b8faa2a038707828c
- Fixes https://github.com/KILTprotocol/ticket/issues/1657
- Fixes https://github.com/KILTprotocol/ticket/issues/1639

Blocking https://github.com/KILTprotocol/sdk-js/pull/431.

## Checklist:

### DIFF
--- a/packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
+++ b/packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
@@ -143,6 +143,7 @@ export const ExtrinsicErrors = {
 
 /**
  * @internal
+ * PalletIndex reflects the numerical index of a pallet assigned in the chain's metadata.
  */
 export enum PalletIndex {
   CType = 9,
--- a/packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
+++ b/packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
@@ -143,7 +143,6 @@ export const ExtrinsicErrors = {
 
 /**
  * @internal
- * PalletIndex reflects the numerical index of a pallet assigned in the chain's metadata.
  */
 export enum PalletIndex {
   CType = 9,
@@ -166,7 +165,6 @@ export interface IPalletToExtrinsicErrors {
 
 /**
  * @internal
- * This dictionary holds all [[ExtrinsicError]]s, divided by pallets.
  */
 export const PalletToExtrinsicErrors: IPalletToExtrinsicErrors = {
   [PalletIndex.CType]: {
@@ -212,12 +210,9 @@ export const PalletToExtrinsicErrors: IPalletToExtrinsicErrors = {
 
 /**
  * @internal
- * Maps a [[ModuleError]] to its corresponding [[ExtrinsicError]].
- *
  * @param p The parameter object.
  * @param p.index The index of the KILT pallet in the metadata.
  * @param p.error The index of the position of the pallet's error definition inside the chain code.
- *
  * @returns A new corresponding [[ExtrinsicError]].
  */
 export function errorForPallet({
--- a/packages/core/src/attestation/Attestation.chain.ts
+++ b/packages/core/src/attestation/Attestation.chain.ts
@@ -151,18 +151,17 @@ export async function reclaimDeposit(
   log.debug(
     () => `Claiming deposit for the attestation with claim hash ${claimHash}`
   )
-  const tx: SubmittableExtrinsic = blockchain.api.tx.attestation.reclaimDeposit(
-    claimHash
-  )
+  const tx: SubmittableExtrinsic =
+    blockchain.api.tx.attestation.reclaimDeposit(claimHash)
   return tx
 }
 
-async function queryRawDepositAmount(): Promise<U128> {
+async function queryDepositAmountEncoded(): Promise<U128> {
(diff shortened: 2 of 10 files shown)


packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -155,3 +156,13 @@ export async function reclaimDeposit(
   )
   return tx
 }
+
+async function queryRawDepositAmount(): Promise<U128> {
```
Dudleyneedham, line 160: Is there a use-case to `queryRawDepositAmount`, it seems most will use the `queryDepositAmount`?

packages/core/src/delegation/DelegationNode.chain.ts
```diff
@@ -179,3 +180,13 @@ export async function getAttestationHashes(
   >(id)
   return decodeDelegatedAttestations(encodedHashes)
 }
+
+async function queryRawDepositAmount(): Promise<U128> {
```
Dudleyneedham, line 184: Again the same as above.

packages/did/src/Did.chain.ts
```diff
@@ -55,6 +55,17 @@ export async function queryDidEncoded(
   return api.query.did.did<Option<IDidChainRecordCodec>>(didIdentifier)
 }
 
+// Query ALL deleted DIDs, which can get quite time consuming.
```
Dudleyneedham, line 58: `// Query ALL deleted DIDs this is time-consuming.`

packages/did/src/Did.chain.ts
```diff
@@ -88,6 +99,11 @@ export async function queryEndpointsCountsEncoded(
   return api.query.did.didEndpointsCount<u32>(didIdentifier)
 }
 
+async function queryDepositAmountEncoded(): Promise<U128> {
```
Dudleyneedham, line 102: Same as above. If it is needed, then this should be renamed to follow the convention of being `raw` amount.

Dudleyneedham (commented):
Minor suggestions to change naming

packages/core/src/attestation/Attestation.chain.ts
Diiaablo95: Not really, but I did it to maintain consistency with the other functions as well. The lowest level fetching directly from the blockchain, and a higher level decoding those values as needed.

packages/did/src/Did.chain.ts
Diiaablo95: You are right. I will rename everything as encoded.

packages/core/src/attestation/Attestation.chain.ts
Diiaablo95: Indeed, the function is not even exported, so not usable outside of this module.

packages/did/src/Did.chain.ts
Dudleyneedham: What you added now reads better.

packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
```diff
@@ -143,7 +143,6 @@ export const ExtrinsicErrors = {
 
 /**
  * @internal
- * PalletIndex reflects the numerical index of a pallet assigned in the chain's metadata.
```
Dudleyneedham, line 146: Please, add back. 👍

packages/did/src/Did.chain.ts
```diff
@@ -55,6 +55,31 @@ export async function queryDidEncoded(
...
+  // Query all the storage keys, and then only take the relevant property, i.e., the encoded DID identifier.
+  return api.query.did.didBlacklist
+    .keys<GenericAccountId[]>()
+    .then((entries) =>
+      entries.map(({ args: [encodedDidIdentifier] }) => encodedDidIdentifier)
+    )
+}
+
+// Returns the raw representation of the storage entry for the given DID identifier.
+export async function queryDidDeletionStatusEncoded(
```
Dudleyneedham, line 70: Does this need to be exported?

Dudleyneedham (changes requested):
Looking good, just a minor question.

packages/chain-helpers/src/errorhandling/ExtrinsicError.ts
Diiaablo95: https://github.com/KILTprotocol/sdk-js/pull/430/commits/f531e493e386bb09102141e7c3346b586885ae0c

packages/did/src/Did.chain.ts
Diiaablo95: https://github.com/KILTprotocol/sdk-js/pull/430/commits/b55b82314ca92e4bc2c590ddc2b0bad5f3c3dfe1

tjwelde (approved):
LGTM

Dudleyneedham (approved):
LGTM

PR #437, opened 2021-11-08
chore: update cbor to v8

tjwelde:
## fixes KILTProtocol/ticket#1677
This updates the CBOR dependency to the latest version, because v6 doesn't work in the Sporran extension.
Since it uses the global URL object and it is missing from the type-definitions (I opened a discussion here: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/57055), I had to add a typescript declaration file for it.

## How to test:
- Run tests
- Test in sporran

## Checklist:

### DIFF
--- a/packages/did/package.json
+++ b/packages/did/package.json
@@ -33,6 +33,6 @@
     "@polkadot/types": "6.4.2",
     "@polkadot/util": "7.5.1",
     "@polkadot/util-crypto": "7.5.1",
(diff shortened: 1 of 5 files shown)


Diiaablo95 (approved):
LGMT! As you see, unit tests are failing as apparently v8 and v6 differ in how the encoding is performed. We can just update the expected results to match that, and hopefully nobody has generated and used any light DID yet.

PR #438, opened 2021-11-09
feat: ajv alternative

tjwelde:
## fixes KILTProtocol/ticket#1602
Tried to use https://github.com/ExodusMovement/schemasafe and noticed too late, that they also use `Function()` to generate the validation code, which also doesn't working with CSP (Still here is a draft PR for it: https://github.com/KILTprotocol/sdk-js/pull/441).
So I had to copy https://github.com/cfworker/cfworker/tree/main/packages/json-schema into the codebase, since it doesn't use `eval` and `Function`. I couldn't add it as a dependency, because this codebase can't import a full es6 module.

## How to test:
- Tests should run
- UNSAFE_EVAL should not be necessary in a CSP environment.

## Checklist:

  - Unfortunately not


packages/core/src/quote/Quote.ts
```diff
@@ -41,26 +40,21 @@ import { QuoteSchema } from './QuoteSchema'
...
-        }
-      })
-    }
+  const validator = new JsonSchema.Validator(schema)
+  if (schema.$id !== QuoteSchema.$id) {
+    validator.addSchema(QuoteSchema)
+  }
+  const result = validator.validate(validate)
+  if (!result.valid && messages) {
+    result.errors.forEach((error) => {
```
Diiaablo95, line 53: Probably out of scope for this PR, but I am not sure what is the purpose of the input `messages` parameter, as, as far as I have seen, it is not used anywhere in the codebase and it is not tested in any test scenario

packages/vc-export/package.json
```diff
@@ -37,8 +37,7 @@
     "@polkadot/keyring": "7.5.1",
     "@polkadot/types": "6.4.2",
     "@polkadot/util": "7.5.1",
-    "@polkadot/util-crypto": "7.5.1",
-    "ajv": "^6.12.5",
+    "@polkadot/util-crypto": "^7.5.1",
```
Diiaablo95, line 40: Can we use the `^` here where the other @polkadot dependencies use an exact version?

Diiaablo95 (approved):
LGMT! With just a couple of small questions.

packages/vc-export/package.json
tjwelde: ah, that's a mistake. good catch

packages/core/src/quote/Quote.ts
tjwelde: Yeah, this is for finding out, what exactly went wrong while verifying an object.
Not sure what to do about it, since it could be useful for people, if e.g. they want to inform a user filling out a form, what did go wrong.
I guess tests should be at least added in the future.

PR #439, opened 2021-11-10
fix: resolve bug in resolver resolution logic

Diiaablo95:
# Fix a bug in the default DID resolver and improve the structure of the information returned

This PR fixes KILTProtocol/ticket#1692 and KILTProtocol/ticket#1712.

There was a logic in the default DID resolver that was caught when working on the DID Universal Resolver driver and was not caught by the unit tests.

The error was in the following snippet for the `resolveDoc` function of `DefaultResolver.ts`:

```typescript
case 'full': {
      const details = await queryFullDetailsFromIdentifier(identifier, version)
      return { details } as IDidResolvedDetails
    }
```
The line `return { details } as IDidResolvedDetails` would fail in case the details are `null`. This PR fixes that, and also slightly changes the resolution metadata from the following:

```typescript
export type IDidResolutionDocumentMetadata = {
  canonicalId: string
  deleted: boolean
}

export type IDidResolvedDetails = {
  details: IDidDetails
  metadata?: IDidResolutionDocumentMetadata
}
```

to the following:

```typescript
export type IDidResolutionDocumentMetadata = {
  canonicalId?: string
  deactivated: boolean
}

export type IDidResolvedDetails = {
  details?: IDidDetails
  metadata: IDidResolutionDocumentMetadata
}
```

While before the metadata was added only in case a light DID was migrated, while the details of the DID being resolved were **always** returned, now the resolver **always** returns the metadata with at least the deletion information (`true`/`false`), and optionally, if a migrated DID has not been deleted, with the `canonicalId` of the full DID. Furthermore, the DID details are only returned if the DID, be it light or full, has not been deleted.

So, for a deleted DID, the resolution process will simply return

```typescript
{
    metadata: {
        deactivated: true
    }
}
```

while for a migrated light DID that has not been deleted

```typescript
{
    details: <light_did_details>,
    metadata: {
        deactivated: false,
        canonicalId: <full_did_uri>
    }
}
```

This change makes it easier for the universal driver to return the correct info when a DID is deleted, and also makes DID signature verification (and consequently anything that uses those signatures, e.g., credential presentations) easier by simply checking if the `deactivated` flag is set to `true`, invalidating the signature altogher with no other complex logic as it was before.

Plus, this PR changes the DID parsing regex to not accept any Substrate address anymore, but only KILT addresses, meaning addresses that start with a leading `4`, and whose checksum is still correctly verified.

## How to test:

More unit tests have been added for the default resolver, and other tests were updated to reflect the changed structure. Run the unit tests for the default resolver to check for the tested cases.

## Checklist:


packages/core/src/ctype/CType.spec.ts
```diff
@@ -135,11 +135,10 @@ describe('CType', () => {
       SDKErrors.ERROR_OBJECT_MALFORMED()
     )
     expect(() => CType.fromCType(invalidAddressCtype))
-      .toThrowErrorMatchingInlineSnapshot(`
-      "Provided DID identifier address invalid 
-
-          Address: Dp6K4tpdZtY3rNqM2uorQmsS6d3woxtnWMHjtzGftHmDb41N"
-    `)
+      // eslint-disable-next-line prettier/prettier
```
Diiaablo95, line 138: Apparently using the prettier suggestions, the snapshots, which were identical, were failing to match for equality. So the `'` and the `"` have to be on the same line for it to succeed.

tjwelde (approved):
LGTM

PR #440, opened 2021-11-10
fix: keyring and type from self exposed

LeonFLK:
## fixes KILTProtocol/ticket#1694

Import Keyring and types from self exposed location, to avoid having redundant packages, reducing bundle size etc.
This could also be done with all the other polkadot stuff we use.

## Checklist:


Dudleyneedham (approved):
LGTM

PR #442, opened 2021-11-16
feat: getting started to remove outdated dev chain

Dudleyneedham:
## Checklist:


tjwelde:
yes, but after the last release, it has to be the spiritnet chain we point to.
Also there is the question, if we should advertise peregrine? Since that will be the only chain with a faucet...

tjwelde:
getting started was moved to docs

PR #446, opened 2021-12-12
chore: tsconfig.esm for ESM outputs

jacogr:
This adds a `tsconfig.esm.json` to allow for the output of ESM into the `lib/esm` folder.

The important thing to note is that this doesn't break anything. However, it is not complete

- the ESM outputs does not import from e.g. `errorhandler.js` which is required for Node 12 resolution. (more on this later)
- there are no export maps provided, i.e. this won't by default magically start using the ESM outputs (this is to the first point that "nothing breaks")
- there are also no `"module": "./lib/esm/index.js"` specifier in the `package.json` files - this can only be added after the filename extensions are resolved (otherwise browser bundles may work, but Node ESM won't)

On the ESM imports, Node requires the extra `.js` specifier and TSC (up to 4.5) does not add it. So options are -

- wait till tsc 4.6 and use `"moduleResolution": "node12"` in the `tsconfig.esm.json`
- everywhere where we import from a file, add the `.js` path manually (this could be error prone, but TS fully allows it - the issue is that along the way somebody might forget)
- add a script that check for local imports (files) and add the extension to the output (this is more-or-less what I do elsewhere, although I compile via babel and it has a plugin that does that)

## How to test:

1. Run `yarn build`
2. Inspect the `lib/esm` outputs

## Checklist:

### DIFF
--- a/packages/chain-helpers/tsconfig.esm.json
+++ b/packages/chain-helpers/tsconfig.esm.json
@@ -1,0 +1,15 @@
+{
+  "extends": "../../tsconfig.esm.json",
+
+  "compilerOptions": {
+    "outDir": "./lib/esm"
+  },
+
+  "include": [
+    "src/**/*.ts", "src/**/*.js"
+  ],
+  "exclude": [
+    "node_modules",
+    "**/*.spec.ts"
+  ]
+}
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
-    "build:ts": "tsc --declaration -p tsconfig.build.json"
+    "build:ts": "tsc --declaration -p tsconfig.build.json && tsc -p tsconfig.esm.json"
   },
   "repository": "github:kiltprotocol/sdk-js",
   "engines": {
--- a/tsconfig.esm.json
+++ b/tsconfig.esm.json
@@ -1,0 +1,7 @@
+{
+  "extends": "./tsconfig.build.json",
+  "compilerOptions": {
+    "module": "es6",
+    "moduleResolution": "node"
+  }
+}
--- a/packages/chain-helpers/package.json
+++ b/packages/chain-helpers/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
-    "build:ts": "tsc --declaration -p tsconfig.build.json"
+    "build:ts": "tsc --declaration -p tsconfig.build.json && tsc -p tsconfig.esm.json"
   },
   "repository": "github:kiltprotocol/sdk-js",
   "engines": {
--- a/packages/config/tsconfig.esm.json
+++ b/packages/config/tsconfig.esm.json
@@ -1,0 +1,15 @@
+{
+  "extends": "../../tsconfig.esm.json",
+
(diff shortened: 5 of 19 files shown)


packages/chain-helpers/tsconfig.esm.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.esm.json",
```
jacogr, line 2: Not 100% happy with this since it duplicates - so options were -

- don't add the root `tsconfig.esm.json`
- extend from the local and add the esm/nodeResolutions outputs to all

Neither approach is DRY, but I guess the approach specified here (which is not committed), may be DRY-er.

packages/config/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
-    "build:ts": "tsc --declaration -p tsconfig.build.json"
+    "build:ts": "tsc --declaration -p tsconfig.build.json && tsc -p tsconfig.esm.json"
```
jacogr, line 13: We don't re-output the declarations. In the export maps (if/when added) we will just map to the top-level.

tsconfig.esm.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.build.json",
+  "compilerOptions": {
+    "module": "es6",
+    "moduleResolution": "node"
```
jacogr, line 5: Keeping the `"moduleResolution": "node"` so it can be converted to `"moduleResolution": "node12"` with TSC 4.6

tjwelde:
Hey Jaco, thank you very much for your PR.
We are actually thinking about switching completely to ES Modules, since it is stable since node v14.17.0. (we would require this version in our engine definition)
Other than not supporting older node versions, do you see any issues with that approach?

jacogr:
It is just supporting users that and not fully on ESM that is an issue.

In a perfect world, I would be there already :) Sadly, the amount of different environments I have to deal with just makes it a non-starter atm. I honestly believe it is the right approach and the future. With a time-machine I would just a short while ahead and go full-in.

There are a number of vendors that have switched to ESM completely. Generally their approach is to bump to a major and tell non-CJS people to stay on older versions. Once again, it is not quite what I can do since things changes the whole time. So a rock and a hard-place.

I don't known your userbase, so don't know how they will be affected. It may be worth a shot by bumping, releasing ESM-only and if there is a huge support overhead, go for a dual-packaging approach.

jacogr:
Going to close this one. It was more meant as a POC. The base is good, but obviously needs "tweaks" to be "final". However, do feel free to re-use as-is if it adds value. (And if it does and can go as-is, happy to re-open as well.)

TL;DR Don't want to clutter the list with things that won't go, I'm sure you have enough to think about as it stands.

tjwelde:
Thanks for having made this PR with all the information. It was a good starting point.
I researched a lot more an am currently working on providing both commonjs and an esm in our packages.
I am going the route of transpiling to `lib` subdirectories, e.g. `lib/esm`and adding a package.json file there, which just contains `{ "type": "module" }` for the esm package.

tjwelde:
hey @jacogr could you have a look at https://github.com/KILTprotocol/sdk-js/pull/451 and tell me what you think?
It fully supports cjs and esm, but maybe with an unusual way of hinting on the module format. (what I talked about in my last comment here)

jacogr:
Will do - not tonight, but first thing on the morning.

PR #452, opened 2022-01-11
refactor: kill extrinsic errors, use metadata error matching instead

rflechtner:
## fixes KILTProtocol/ticket#1652

This makes use of module error matching based on chain metadata info.
Our internal extrinsic error matching is removed.
Minor changes necessary to get this done relate to the recovery by resign functionality, and the SubscriptionPromise maker. 

## How to test:
test & integration tests suffice 

## Checklist:


packages/chain-helpers/src/blockchain/Blockchain.utils.ts
```diff
@@ -116,48 +90,41 @@ export async function dispatchTx(
...
+    )
   }
+  if (
+    reason &&
+    typeof reason === 'object' &&
+    typeof reason.status === 'object'
+  ) {
+    const { status } = reason as ISubmittableResult
+    if (status.isUsurped) return true
+    // if (status.isRetracted) return true
```
tjwelde, line 123: Is this still needed?

packages/chain-helpers/src/errorhandling/ErrorHandler.ts
```diff
@@ -72,20 +57,16 @@ export function extrinsicSuccessful(
...
 
   if (errorEvent && errorEvent.isModule) {
     const moduleError = errorEvent.asModule
-
-    const index = moduleError.index.toNumber()
-    const error = moduleError.error.toNumber()
-    if (index >= 0 && error >= 0) {
-      return errorForPallet({ index, error })
+    try {
+      return moduleError.registry.findMetaError(moduleError)
```
tjwelde, line 66: oh nice. You don't even have to pass through the `api` object. Good find!

packages/chain-helpers/src/errorhandling/ErrorHandler.ts
```diff
@@ -72,20 +57,16 @@ export function extrinsicSuccessful(
...
+    } catch {
+      // handled below
```
tjwelde, line 68: Maybe different wording, in case more code comes in between in the future?
Like "handled with last return"?

tjwelde (changes requested):
That looks SO MUCH better!
Just some nitpicks

packages/chain-helpers/src/blockchain/Blockchain.utils.ts
rflechtner: no, this can go - retracted actually means the transaction goes back in the pool, so we would not want to resubmit. Unsure about `dropped` and `invalid` though

rflechtner:
while working on this I wrote this little gadget which I had included in the `ErrorHandler.ts` but not committed (yet):
```ts
export function throwExtrinsicError(extrinsicResult: ISubmittableResult): void {
  const { dispatchError } = extrinsicResult
  if (!dispatchError) return
  if (dispatchError.isModule) {
    const moduleError = dispatchError.registry.findMetaError(
      dispatchError.asModule
    )
    const { section, index, name, docs } = moduleError
    throw new ExtrinsicError(
      `Module Error ${section}.${name}(${index}): ${docs.join(' ')}`
    )
  }
  if (dispatchError.isToken || dispatchError.isArithmetic) {
    throw new ExtrinsicError(
      `${dispatchError.type} Error: ${(dispatchError.value as TokenError).type}`
    )
  }
  throw new ExtrinsicError(`${dispatchError.type} Error`)
}
```
It's not used or needed anywhere, do you think that's something users would want to use?

tjwelde:
Regarding https://github.com/KILTprotocol/sdk-js/pull/452#issuecomment-1011372939:
Not sure about the usefullness. I imagine people will either want to show a generic message, when something goes wrong, or want to look at errors more in depth instead of just parsing an error message, if they want to handle something specific. I would leave it out of this PR.

packages/chain-helpers/src/blockchain/Blockchain.utils.ts
tjwelde: I think it is ok to not regard them here.

tjwelde (approved):
LGTM

PR #457, opened 2022-01-17
style: functions start with lower case

weichweich:
## Related KILTProtocol/ticket#1696

I tried to find an eslint rule for that, but it doesn't exist.

## Checklist:


Dudleyneedham (approved):
LGTM

Diiaablo95:
@weichweich @Dudleyneedham I have already renamed the ctype function in https://github.com/KILTprotocol/sdk-js/pull/447 to `isCtypeOnChain`. So I would propose I add the json renaming in there and we close this one?

PR #458, opened 2022-01-19
refactor: use type definitions from metadata, remove dependencies on type-definitions repo

rflechtner:
## fixes KILTProtocol/ticket#1687
Bye bye static types repo, hello dynamic metadata type imports!
In previous versions of polkadot/api, we had to import our own type-definitions repository to inform polkadot-js's SCALE decoders about the makeup of custom composite types (`DelegationNodeInfo` etc.). This has changed; the required info is now part of the metadata that can be retrieved from any up to date kilt full node.

## What has changed?
Building an api / BlockchainApiConnection instance via the chain helpers no longer imports the static kilt types. Querying the metadata is done implicitly when a new api instance is created.

To approximate this behavior in unit tests and make sure we are not using outdated types, api instances created for testing can use a json dump of the metadata, which is included in a new private package meant exclusively for testing purposes. This package now also houses the blockchain mocks. The package includes a script to update the metadata dump. You can update to the latest version of the kilt spiritnet by running `yarn workspace @kiltprotocol/testing run updateMetadata` and commit changes, if desired.

N.B.: We used to mock a blockchain connection by mocking away the function that would create a new connection instance with a jest mock, which created pretty messy import paths: 
```ts
jest.mock(
   '@kiltprotocol/chain-helpers/lib/cjs/blockchainApiConnection/BlockchainApiConnection'
 )
```
This is not actually necessary since we allow setting a connection instance manually, allowing us to also set a mocked connection instance. This PR updates tests accordingly.

## How to test:
Make sure integration tests run properly.

## Checklist:


packages/did/src/Did.chain.ts
```diff
@@ -527,16 +527,16 @@ export async function generateDidAuthenticatedTx({
   blockNumber,
 }: AuthorizeCallInput & KeystoreSigningOptions): Promise<SubmittableExtrinsic> {
   const { api } = await BlockchainApiConnection.getConnectionOrConnect()
-  const signableCall =
-    new (api.registry.getOrThrow<IDidAuthorizedCallOperation>(
-      'DidDidDetailsDidAuthorizedCallOperation'
-    ))(api.registry, {
+  const signableCall = api.registry.createType<IDidAuthorizedCallOperation>(
+    api.tx.did.submitDidCall.meta.args[0].type.toString(),
```
tjwelde, line 531: sneaking this in? :D
Might be already included in the merged DID refactor?!

packages/testing/package.json
```diff
@@ -0,0 +1,44 @@
...
+      "require": "./lib/cjs/index.js"
+    }
+  },
+  "files": [
+    "lib/**/*"
+  ],
+  "scripts": {
+    "updateMetadata": "node ./scripts/fetchMetadata.js -o './src/mocks/metadata/spiritnet.json' -e 'https://spiritnet.api.onfinality.io/public-http'",
+    "clean": "rimraf ./lib",
+    "build": "yarn clean && yarn build:ts",
```
tjwelde, line 21: does this need to be built?

packages/testing/scripts/fetchMetadata.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const timeout = new Promise((_, reject) => {
+  setTimeout(() => {
+    exitCode = exitCode || 124
+    reject(new Error('timeout waiting for metadata fetch'))
+  }, 10000)
+})
+
+Promise.race([fetch(), timeout])
```
tjwelde, line 78: Ha! interesting way of doing a timeout

packages/did/src/Did.chain.ts
rflechtner: this is from the merge commit integrating the refactor. After Antonio's changes there and my changes here it was the last line to still use hard-coded type names for encoding (apart from testing-only code), so I decided to fix it up here for consistency.

packages/testing/package.json
rflechtner: Pretty sure it does, code that is imported from another package (i.e. testing package) would always import from the path that's in the package.json, right? And we probably can't point to ts code there. The only alternative would be to write js instead I assume.
If we do not build, tests currently fail with `Cannot find module '@kiltprotocol/testing' from '<test file>'`.

packages/testing/scripts/fetchMetadata.js
rflechtner: Is it so unusual? I thought that's how it's done in js. What would you do?

rflechtner: Another thing I'm realising now: do we need to cancel timeouts? I had the feeling that the script wouldn't exit until the timeout timer has elapsed. And jest has started indicating in it's "process has not exited 1s after end of tests" message that timeouts might be a cause.

tjwelde: It's just cool to let them `race`, that's why I found it interesintg. Normally you might have them set-up sequentially, because you might want to cleanup the fetch, when the timeout triggers. But since we exit in this case, that should be fine.

tjwelde (approved):
LGTM

PR #461, opened 2022-01-20
fix: delegation removal integration test

rflechtner:
## NO TICKET

This fixes an integration test concerning delegation node removal to conform to what I understand to be the expected behaviour as it is currently defined:

> Delegation nodes can only be removed by calling a removal extrinsic on a node signed by the the owner of that node (a DID). This also removes all children of this node; however, a parent may not selectively remove child nodes.

Note that this is different from revoking a delegation node, where different rules apply.

Nota bene: To my best knowledge, the test never reflected expected behaviour, so I am not sure how this ever worked. From what I gathered the expected behaviour was previously defined as above, but with the extension that the owner of a deposit associated with the delegation node (a substrate account I would assume) can also remove that node to reclaim their deposit. This test however had the owner of the parent node try and remove the child node, and oddly this succeeded on the previous release candidate of the mashnet node (which I assume to be the latest full release).
This would only make sense if the owner of the delegation node was also the owner of the deposit, but judging from the test itself that was not the case; the extrinsic submission was paid for by the faucet account, which was not associated with the delegator did.  

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/Delegation.spec.ts
+++ b/packages/core/src/__integrationtests__/Delegation.spec.ts
@@ -252,12 +252,12 @@ describe('revocation', () => {
     ).resolves.not.toThrow()
     await expect(delegationA.verify()).resolves.toBe(false)
 
-    // Test removal with deposit payer's account.
+    // Test removal with delegee's did.
     await expect(
       delegationA
         .remove()
         .then((tx) =>
-          delegator.authorizeExtrinsic(tx, signer, paymentAccount.address)
+          firstDelegee.authorizeExtrinsic(tx, signer, paymentAccount.address)
         )
         .then((tx) =>
           BlockchainUtils.signAndSubmitTx(tx, paymentAccount, {


rflechtner:
The only possible explanation I see is that the delegator was made deposit owner even if they did not pay it; could work with the descriptions in the original PR (https://github.com/KILTprotocol/mashnet-node/pull/258), which says
> Adds a deposit as a reserve to a delegation which will be paid by the user who delegates (invites another user)

and

> The delegator can remove the delegation at any point (it does not have to be revoked) which will unreserve the deposit

weichweich:
deposit owner is a completely different address type than deletgator or delegate. Also the deposit owner cannot remove a node by calling `remove`.

To sum up:
The old-state in the test was, the owner of the parent node tried to remove one of their children. This was failing.
The new-state in the test is, the owner of the child deletes his own node. This is passing.

The deposit paying and TX sending is all done with the same account and didn't change.

wischli:
> To sum up: The old-state in the test was, the owner of the parent node tried to remove one of their children. This was failing. The new-state in the test is, the owner of the child deletes his own node. This is passing.
> 
> The deposit paying and TX sending is all done with the same account and didn't change.

Yes.

Diiaablo95:
I also addressed the change here -> https://github.com/KILTprotocol/sdk-js/pull/447/commits/3b4cf91b7015764138f402d8e4fc7d4d47045b98. That is exactly due to the change described above.

rflechtner:
Antonios fix indeed looks cleaner, we should probably try to get that in asap instead. Do we still test that the node owner can remove elsewhere?

Diiaablo95:
Yes, I added a further test case to test that. I'll try to address the comments in the PR today and merge it by the end of today.

weichweich:
> I also addressed the change here -> [3b4cf91](https://github.com/KILTprotocol/sdk-js/commit/3b4cf91b7015764138f402d8e4fc7d4d47045b98) (part of the open PR #447). That is exactly due to the change described above.

what's not part of this PR! 😂

Diiaablo95:
Superseded by https://github.com/KILTprotocol/sdk-js/pull/463.

PR #463, opened 2022-01-24
fix: ensure message is owner in messages

Diiaablo95:
## fixes KILTProtocol/ticket#1772

Fixes a bug in the message verification logic which used to match for equality between sender and owner.
This PR introduces an utility function in `DidUtils` called `isSameSubject(didA, didB)` which returns `true` if both DIDs refer to the same subject (i.e., the two DIDs share the same identifier). Several unit tests have been added to the messaging unit tests.

### DIFF
--- a/packages/messaging/src/Message.spec.ts
+++ b/packages/messaging/src/Message.spec.ts
@@ -9,6 +9,8 @@
  * @group unit/messaging
  */
 
+import { hexToU8a } from '@polkadot/util'
+
 import type {
   ICredential,
   IEncryptedMessage,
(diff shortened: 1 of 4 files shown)


Dudleyneedham (approved):
LGTM

packages/messaging/src/Message.spec.ts
```diff
@@ -243,14 +326,375 @@ describe('Messaging', () => {
...
+        new Message(
+          submitAttestationBodyWithEncodedDetails,
+          lightDidAttester.did,
+          lightDidOwner.did
+        )
+      )
+    ).not.toThrow()
+
+    // Should not throw if the sender is the full DID version of the owner.
+
```
tjwelde, line 614: C: Remove empty line

tjwelde (approved):
LGTM

PR #464, opened 2022-01-24
refactor: submittable extrinsic renaming

Dudleyneedham:
## fixes  https://github.com/KILTprotocol/ticket/issues/1675

If a function is a getter for a `SubmittableExtrinsic` without doing more stuff inside, they should start with get and end with Tx.

## How to test:

yarn test

## Checklist:


packages/core/src/balance/Balance.chain.ts
```diff
@@ -124,15 +124,15 @@ export async function listenToBalanceChanges(
...
+ * sdk.Balance.getMakeTransferTx(address, amount, 0) //
  *   .then(tx => BlockchainUtils.signAndSendTx(tx, identity))
  *   .then(() => console.log('Successfully transferred ${amount.toNumber()} tokens'))
  *   .catch(err => {
  *     console.log('Transfer failed')
  *   })
  * ```
  */
-export async function makeTransfer(
+export async function getMakeTransferTx(
```
tjwelde, line 135: S: I think `getTransferTx` would be even better here.

tjwelde: There are more places, where it is still `getMakeTransferTx` please search through your whole codebase.

tjwelde (approved):
LGTM

PR #469, opened 2022-02-01
chore: fast build

tjwelde:
## no-ticket
This PR adds a new building option, which is much faster (4min vs 4sec).
It only transpiles to commonjs, does no type checks, but can be used for jest testing, either locally or in PRs.

## TODO
- [x] Use fast build in tests for PRs

## How to test:
- `yarn run build:fast`
- `yarn run test`

## Checklist:


tjwelde:
Couldn't get it to work. Does not produce testable code.

PR #474, opened 2022-02-07
dedupe dependencies in lockfile

rflechtner:
## NO TICKET

Even though yarn v1 does advertise that it takes care of deduplication of dependencies, it does not actually do that.
Yarn v2 has a `dedupe` command again to do that. It's worthwhile running this from time to time to keep install times low and minimise bundle size. This PR deduplicates and also sets up a warning on PR workflows in case deps may be deduplicated.

## How to test:

There is at least one workflow run in this PR that issued a warning. It's also a good idea to test with this new set of dependencies - but in the end the lock file does not matter for the end user anyway (except for bundle users).

## Checklist:


tjwelde (approved):
LGTM

PR #475, opened 2022-02-11
fix: jsonabc import

tjwelde:
This PR fixes an import problem, when using in a create-react-app project.

## How to test:
- use yalc to publish `@kiltprotocol/utils`
- use it via esm
- use it via cjs

## Checklist:


weichweich (approved):
LGTM

PR #478, opened 2022-02-21
fix: demo keystore error with Ecdsa keys

Diiaablo95:
## fixes [KILTProtocol/ticket#1834](https://github.com/KILTprotocol/ticket/issues/1834)

- Fix the error
- Adds minimal unit tests for the demo keystore
- Adds an additional integration test writing a full DID with an Ecdsa authentication key (to test signing chain stuff)

There is some repetition to compute an address from a public key and a type, but I did not find a better way to solve this, because the code must be present in both the demo keystore, which talks `SigningAlgorithms`, and the DID batcher, which talks `VerificationKeyType`.


packages/did/src/DemoKeystore/DemoKeystore.ts
```diff
@@ -84,7 +103,8 @@ export class DemoKeystore
       )
     const keyType = DemoKeystore.getKeypairTypeForAlg(alg)
     try {
-      const keypair = this.signingKeyring.getPair(publicKey)
+      const encodedAddress = encodeSigningPublicKeyToAddress(publicKey, alg)
+      const keypair = this.signingKeyring.getPair(encodedAddress)
```
rflechtner, line 107: Love your solution!
Although `getPairs().find(pair => u8aEq(pair.publicKey, publicKey)` could also have done the trick? (maybe additionally filtering for key type)
Alternatively you can set arbitrary metadata on a keyring pair, which if we put the alg and or public key here can help with filtering

rflechtner (approved):
LGTM

packages/did/src/DemoKeystore/DemoKeystore.ts
Diiaablo95: Yes, there could be alternative ways. This is also the way polkadotjs goes with, but instead of a function they have an object of closures to transform each public key to an address depending on the type. And this also provides constant time access to the right key, if present.

packages/did/src/DidBatcher/FullDidCreationBuilder.ts
```diff
@@ -127,9 +133,17 @@ export class FullDidCreationBuilder extends FullDidBuilder {
 
     this.consumed = true
 
+    const { publicKey, type } = this.authenticationKey
+
+    const pk =
+      type === VerificationKeyType.Ecdsa && publicKey.length > 32
```
tjwelde, line 139: C: since you are using it twice in this file and similarly also in the DemoKeyStore, could this maybe be a little util function?

Diiaablo95: https://github.com/KILTprotocol/sdk-js/pull/478/commits/8b62e3aad995ee0c1332052a536a18e3d037a0ce

The problem of using one function for both cases is that in the builder we deal with `DidVerificationKeyType`s, while in the demo keystore we deal with `SigningAlgorithms`, which can potentially be unrelated to DIDs. So we must declare two separate functions for both cases. I have declared a utility function for the DID case in this commit, so I guess all is good now 👍

PR #479, opened 2022-02-22
fix: reject subscription promise on disconnect

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/1835

To avoid tx status & event subscription to go stale or miss important updates (like finalisation), this makes sure that they are rejected on disconnect. Users would then have to go and check on chain if their transactions were successful by checking the desired effects. 

## How to test:
Using a test chain, make a transaction, submit it via the Blockchain helpers and interrupt the connection after receiving the `isReady` status update (e.g. by switching wifi off).

## Checklist:


tjwelde (approved):
ah interesting way of handling it. I like it.

PR #480, opened 2022-02-23
feat: add web3 name deposit querying

Diiaablo95:
## fixes [KILTProtocol/ticket#312](https://github.com/KILTprotocol/ticket/issues/1841)

- Add support for querying web3 claiming amount
- Add deposit integration tests to check that deposits are correctly taken/released


packages/did/src/Web3Names/Web3Names.chain.ts
```diff
@@ -134,3 +135,13 @@ export async function queryDidForWeb3Name(
...
 }
+
+/**
+ * Retrieves the deposit amount to claim a web3 name as currently stored in the runtime.
+ *
+ * @returns The deposit amount. The value is indicated in femto KILTs.
+ */
+export async function queryDepositAmount(): Promise<BN> {
+  const { api } = await BlockchainApiConnection.getConnectionOrConnect()
+  return new BN(api.consts.web3Names.deposit.toString())
```
tjwelde, line 146:
```suggestion
  return (api.consts.web3Names.deposit as u128).toBn()
```
In this case I think using `as` is justified, because we actually know better (in other query methods we use generics to specify this).
And if you do it like this, you don't have to convert to string and back (`u128` is basically `BN`  so not much is happening in `toBN` here.

tjwelde (changes requested):

packages/did/src/Web3Names/Web3Names.chain.ts
Diiaablo95: You are 100% right https://github.com/KILTprotocol/sdk-js/pull/480/commits/b16d2a81aabf26b3d09bd92b6b49bf35fb097faa

PR #481, opened 2022-02-28
test against updated dependencies

rflechtner:
## fixes KILTProtocol/ticket#1826

Will run unit & integration tests every Saturday night, using an install with no lock file. That means all dependencies are at their latest version allowed by the semver range.
Will run from the default branch (develop), but will checkout master (latest commit).

I fixed two type issues in an attempt to make the tests run on the current branch with the latest polkadot deps, but could revert that potentially. 
I'm torn especially about deleting the `global.d.ts` (https://github.com/KILTprotocol/sdk-js/pull/481/commits/06c536bc504aeee883ebc08522875a54c1881037) although it constituted a dirty fix for a specific version of `@types/node` that is lacking some global definitions. The community was torn on this one and removed and re-added it repeatedly, so that some versions have it, others don't. For those that have it, the global definition produces a type conflict. For those that don't have it, removing the global definition results in an error with a missing type in one of our dependencies.

## How to test:
Look at the older workflow runs, e.g. [this](https://github.com/KILTprotocol/sdk-js/runs/5376718315)

## Checklist:

### DIFF
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -47,21 +47,6 @@ jobs:
           name: build-${{ matrix.node-version }}
           path: build.zip
 
-  lint:
-    runs-on: ubuntu-latest
-    needs: build
-
-    steps:
-      - uses: actions/download-artifact@v2
-        with:
-          name: build-14.x
-      - name: unzip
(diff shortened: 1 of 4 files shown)


tjwelde (approved):
LGTM

PR #483, opened 2022-03-03
deps: update polkadot dependencies to latest

Diiaablo95:
Update polkadot dependencies to latest and fixes the consequent typescript error according to https://polkadot.js.org/docs/api/FAQ/#since-upgrading-to-the-7x-series-typescript-augmentation-is-missing.

Fixes https://github.com/KILTprotocol/ticket/issues/1870.

In short:

- adds a `import '@polkadot/api-augment'` to the `index.ts` of each package (there might be a better way, but I would suggest to investigate that as part of the next release cycle)
- updates the mocks to compile with the latest version
- updates integration tests to run against `latest-master` and not `latest-rc`

## How to test

Start a new project from scratch and import `@kiltprotocol/sdk-js@0.26.0-beta.2`, which matches the version in this PR.


tjwelde (approved):
LGTM

PR #493, opened 2022-03-07
ci: create issue on failing dependency tests

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/1826

This creates an issue on the sdk repository when the scheduled workflow run fails. Unfortunately I can't currently assign issues to our weekly board - the configuration allows it, but fails.

## How to test:
Look at run https://github.com/KILTprotocol/sdk-js/actions/runs/1947824073 and ticket https://github.com/KILTprotocol/sdk-js/issues/494

## Checklist:


.github/workflows/tests-polkadot-deps.yml
```diff
@@ -77,10 +86,10 @@ jobs:
 
     strategy:
       matrix:
-        image: ['latest-rc']
-        required: ['optional']
+        image: ['latest']
+        required: ['required']
         include:
-          - image: 'latest-develop'
+          - image: 'latest-rc'
```
Diiaablo95, line 92: `latest-rc` is now `latest-master`. I though I merged the change back into `develop`...

.github/workflows/tests-polkadot-deps.yml
```diff
@@ -77,10 +86,10 @@ jobs:
 
     strategy:
       matrix:
-        image: ['latest-rc']
-        required: ['optional']
+        image: ['latest']
```
tjwelde, line 89: latest-master required and latest-develop optional?

tjwelde: waiting for answer regarding node tagging

tjwelde (approved):
Love it

.github/workflows/tests-polkadot-deps.yml
tjwelde: so, the `latest` tag moves forward with every release, so I think this is fine

rflechtner: Ok I’ll merge then !

PR #495, opened 2022-03-09
chore: add release candidate workflow

Diiaablo95:
Fixes https://github.com/KILTprotocol/ticket/issues/1858 and fixes https://github.com/KILTprotocol/ticket/issues/1877.

On the side, it also deletes seemingly outdated dispatched upon dev releases, which are always failing anyway.

### DIFF
--- a/.github/workflows/devpackpublish.yml
+++ b/.github/workflows/devpackpublish.yml
@@ -1,4 +1,4 @@
-name: Publish prerelease package to npm
+name: Publish dev package to npm
 on:
   push:
     branches:
@@ -33,7 +33,7 @@ jobs:
       - name: Set all other packages to same version
         env:
           PACKAGE_VERSION: ${{ steps.package_version.outputs.package_version }}
-        run: yarn workspaces foreach -pt --no-private exec npm version --no-git-tag-version $PACKAGE_VERSION
+        run: yarn set:version $PACKAGE_VERSION
       - name: Build packages
         run: yarn run build
       - name: Bundle library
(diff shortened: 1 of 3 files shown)


.github/workflows/devpackpublish.yml
```diff
@@ -48,23 +48,4 @@ jobs:
           git config --local user.email "action@github.com"
           git add .
           git commit -m "[CI Skip] ci: publish prerelease" -n
-          git push
-  dispatch:
```
Diiaablo95, line 52: Is it ok to remove these outdated triggers?

tjwelde: yes

.github/workflows/devpackpublish.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Set all other packages to same version
         env:
           PACKAGE_VERSION: ${{ steps.package_version.outputs.package_version }}
-        run: yarn workspaces foreach -pt --no-private exec npm version --no-git-tag-version $PACKAGE_VERSION
```
tjwelde, line 36: That would also set the version for the root package, what has been done right before. Not sure if that is a problem, though.

tjwelde (approved):
LGTM

.github/workflows/devpackpublish.yml
Diiaablo95: The `set:version` does exactly the same as this line that is being replaced, including changing the root package version. Why would this be a problem?

rflechtner (approved):
Works for me!

PR #498, opened 2022-03-15
chore: remove symmetric crypto

rflechtner:
## fixes KILTProtocol/ticket#1837
Does what's in the name.

## How to test:
Not required 

## Checklist:


Diiaablo95 (approved):
LGMT! Although not sure why we are removing the symmetric crypto?

rflechtner:
> LGMT! Although not sure why we are removing the symmetric crypto?

Eh, I'm not making the rules here okey? ¯\_(ツ)_/¯

PR #499, opened 2022-03-15
fix: bump node types

rflechtner:
## fixes KILTProtocol/sdk-js#497

Bumps `@types/node` in our lock file to a version that includes global definitions for `URL` and removes these definitions from our repo. This is to avoid conflicts between declarations as typescript does not simply merge these. 
SDK consumers for whom this creates issues should either update their node types, make a global declaration in their own project or disable dependency type checks. 

## How to test:
N/A

## Checklist:

### DIFF
--- a/packages/did/src/global.d.ts
+++ b/packages/did/src/global.d.ts
@@ -1,7 +1,0 @@
-import type { URL as NodeURL } from 'url'
-
-declare global {
-  class URL extends NodeURL {}
-}
-
-export {}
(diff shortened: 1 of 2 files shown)


Diiaablo95 (approved):
This does look good to me, but would like for @tjwelde to double-check this.

tjwelde:
yeah, hope they keep it in this time.
Since this concerns mostly new projects, I think it is fine to merge it  like this. Projects with lock files will not be affected until they upgrade and upgrading already needs work to resolve conflicts.

PR #500, opened 2022-03-15
fix: dev release pipeline

Diiaablo95:
https://github.com/KILTprotocol/sdk-js/runs/5554031564?check_suite_focus=true was failing. I guess this would fix this issue, as suggested?


rflechtner (approved):
yeah I think that should work

PR #506, opened 2022-03-21
feat: check service urls

rflechtner:
## fixes KILTProtocol/ticket#1888
This performs runtime checks on service uris and service ids that assure they are valid URI and URI fragments, respectively.

## How to test:
Unit tests are included.

## Checklist:

### DIFF
--- a/packages/did/src/Did.chain.spec.ts
+++ b/packages/did/src/Did.chain.spec.ts
@@ -1,4 +1,11 @@
 /**
+ * Copyright 2018-2021 BOTLabs GmbH.
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
+/**
  * @group unit/did
  */
 
--- a/packages/did/src/Did.chain.spec.ts
+++ b/packages/did/src/Did.chain.spec.ts
@@ -1,0 +1,102 @@
+/**
+ * @group unit/did
+ */
+
+import {
+  Blockchain,
+  BlockchainApiConnection,
+} from '@kiltprotocol/chain-helpers'
+import { ApiMocks } from '@kiltprotocol/testing'
+import { getAddEndpointExtrinsic } from './Did.chain'
+
+let api: any
+
+beforeAll(() => {
+  api = ApiMocks.createAugmentedApi()
+  BlockchainApiConnection.setConnection(Promise.resolve(new Blockchain(api)))
+})
+
+describe('services validation', () => {
+  const validTestURIs = [
+    'data:image/gif;base64,R0lGODdhMAAwAPAAAAAAAP///ywAAAAAMAAwAAAC8IyPqcvt3wCcDkiLc7C0qwyGHhSWpjQu5yqmCY',
+    'data:text/plain;charset=iso-8859-7,%be%fg%be',
+    'http://2.example.org#frag2',
+    'http://www.example.org/foo.xml#xpointer(//Rube)',
+    'http://www.example.com/questions/3456/my-document',
+    'https://john.doe@www.example.com:123/forum/questions/?tag=networking&order=newest#top',
+    'ldap://[2001:db8::7]/c=GB?objectClass?one',
+    'mailto:John.Doe@example.com',
+    'news:comp.infosystems.www.servers.unix',
+    'tel:+1-816-555-1212',
+    'telnet://192.0.2.16:80/',
+    'urn:oasis:names:specification:docbook:dtd:xml:4.1.2',
+    'ctype:0x12345A',
+    'did:example:1234',
+  ]
+  const validTestIds = [
+    'frag2',
+    'My_Awesome_Service',
+    '@johndoe',
+    'http://example.com',
+    'abc.de',
+  ]
+
+  const unencodedTestUris = [
+    'http://www.example.org/Dürst',
+    'http://www.example.org/foo bar/qux<>?^`{|}',
+  ]
+  const invalidTestIds = ['¯\\_(ツ)_/¯', '#1234567890', '😇', 'μαλάκα', 'мир']
+  const malformedTestUris = [...invalidTestIds, 'www.com']
+
+  it.each([...validTestURIs, ...unencodedTestUris.map(encodeURI)])(
+    'allows adding services with valid URI "%s"',
+    async (uri) => {
+      await expect(
+        getAddEndpointExtrinsic({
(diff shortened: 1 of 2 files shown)


ntn-x2 (commented):
Love it! I would also add the checks to the `LightDidDetails` class, so that we avoid a situation in which an "invalid" light DID is used and then it can't be upgraded because of some characters that are not allowed.

rflechtner:
> Love it! I would also add the checks to the `LightDidDetails` class, so that we avoid a situation in which an "invalid" light DID is used and then it can't be upgraded because of some characters that are not allowed.

True, I didn't think about that. Will probably have to move functionality out of the chain file then.

rflechtner:
> > Love it! I would also add the checks to the `LightDidDetails` class, so that we avoid a situation in which an "invalid" light DID is used and then it can't be upgraded because of some characters that are not allowed.
> 
> True, I didn't think about that. Will probably have to move functionality out of the chain file then.

Mhh this seems to be a larger issue, given that the checks on lightDids we currently do are something else entirely, and we do not check for length, number of urls, etc (see [here](https://github.com/KILTprotocol/sdk-js/blob/cc641350de9fe735aedb0d06166ad0fe020edc62/packages/did/src/DidDetails/LightDidDetails.utils.ts#L120-L144)). Do you think we should start doing this?

ntn-x2:
Yes, unfortunately that is the case. I would say that it would be a good thing if we do that, so that we avoid problems when upgrading to a full DID, as invalid input will not be accepted, but the light DID might already be issued credentials.

rflechtner:
@tjwelde we're unsure about whether or not we should introduce constructor checks for services because it would mean that some LightDids created with older sdk versions stop working with newer ones. Technically, a Did where the service id is not a URI fragment - or where the URL/URI field is not a valid URI - has always been against the specs, but we have not been enforcing this so far. What do you think? Alternatively we could stick to enforcing these rules only on writing on chain, which means that anything that is already on there remains valid, and would not touch lightDids at all. In that case, users may then face errors when trying to upgrade their lightDids, of course.

tjwelde:
hmm, I think that lightDids using service endpoints are not used anywhere yet, so it is fine for me to introduce this breaking change now.

packages/did/src/Did.utils.ts
```diff
@@ -380,3 +380,21 @@ export function assembleKeyUri(
...
+    return url.href === s || encodeURI(decodeURI(s)) === s // make sure our URI has not been converted implictly by URL
+  } catch {
+    return false
+  }
+}
+
+const UriFragmentRegex = /^[a-zA-Z0-9._~%+,;=*()'&$!@:/?-]+$/
+export function isUriFragment(s: string): boolean {
+  try {
+    return UriFragmentRegex.test(s) && !!decodeURIComponent(s)
```
tjwelde, line 396: What does the second half of this condition do?

packages/did/src/Did.utils.ts
```diff
@@ -380,3 +380,21 @@ export function assembleKeyUri(
   }
   return `${did}#${keyId}`
 }
+
+export function isUri(s: string): boolean {
+  try {
+    const url = new URL(s) // this actually accepts any URI but throws if it can't be parsed
```
tjwelde, line 386: hmm, what are the pros of using `new URL` here instead of a regex?

rflechtner: that you don't have to define a regex that is so long and convoluted that nobody understands it. Try defining the regex for a valid URI

packages/did/src/Did.utils.ts
rflechtner: it tries to parse this as a URI encoded string, and would throw if anything does not make sense (like invalid percent encoding etc.)

packages/did/src/Did.utils.ts
rflechtner: (your regex would not just have to limit the character set but reproduce URI syntax)

rflechtner:
After discussing with @tjwelde we decided to have no length checks on the LightDids for now, only syntax checks. Instead we require users to opt-in if they want their service endpoints migrated to the FullDid

weichweich:
As discussed, please also fix https://github.com/KILTprotocol/ticket/issues/1948

packages/did/src/commonHelpers.ts
```diff
@@ -0,0 +1,184 @@
+/**
```
weichweich, line 1: IMHO things like `commonHelpers`, `util` should be avoided by all means. It's not telling you what is contained at all and the separation to other modules is not given at all.

packages/did/src/commonHelpers.ts
```diff
@@ -0,0 +1,184 @@
...
+const UriFragmentRegex = /^[a-zA-Z0-9._~%+,;=*()'&$!@:/?-]+$/
+export function isUriFragment(s: string): boolean {
+  try {
+    return UriFragmentRegex.test(s) && !!decodeURIComponent(s)
+  } catch {
+    return false
+  }
+}
+
+export function checkServiceEndpointSyntax(
```
weichweich, line 92: What is the reason for returning the error and not throwing it?
Instead of returning an error I would maybe combine the string to list the problems and return a single error.

packages/did/src/commonHelpers.ts
rflechtner: > IMHO things like `commonHelpers`, `util` should be avoided by all means. It's not telling you what is contained at all and the separation to other modules is not given at all.

Yeah I hate it too. The only reason this exists is because anything else provokes problems with import cycles. Did not find another way to solve this issue, so our options are duplicating code or not checking service endpoints on the builder class at all. Note that for the end user this is invisible, because it is only exported via the utils file. And @tjwelde is working on reorganising the whole codebase, so this would be temporary in any case.

packages/did/src/commonHelpers.ts
rflechtner: Also a possible solution. I thought this is something you would want to do on a higher level (like I did it on `packages/did/src/DidBatcher/FullDidBuilder.ts`).

ntn-x2:
Can be merged after pulling `develop` in.

packages/did/src/commonHelpers.ts
tjwelde: We have been reorganising the code a little bit. Got rid of the commonHelpers and just put the function responsible for the circular dependency into a `Did.signature.ts` file

PR #508, opened 2022-03-21
hotfix: cherry pick integration test fix for scheduled runs

rflechtner:
fixes https://github.com/KILTprotocol/sdk-js/issues/504

As far as I can see the only issue leading to failure in our weekly run with updated deps is that the integration tests are out of sync with the latest chain runtime. This should take care of the problem.

### DIFF
--- a/packages/core/src/__integrationtests__/Attestation.spec.ts
+++ b/packages/core/src/__integrationtests__/Attestation.spec.ts
@@ -11,7 +11,6 @@
 
 import type { ICredential, IClaim, KeyringPair } from '@kiltprotocol/types'
 import { DemoKeystore, FullDidDetails } from '@kiltprotocol/did'
-import { BN } from '@polkadot/util'
 import { Crypto } from '@kiltprotocol/utils'
 import { Attestation } from '../attestation/Attestation'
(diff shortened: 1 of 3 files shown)


ntn-x2 (approved):
LGTM! Unfortunately, the merging of https://github.com/KILTprotocol/mashnet-node/pull/316 into `develop` means that integration tests now are again broken, because something changed in the storage layout. It is definitely not to be fixed in this PR, but I guess we will keep having problems as long as the issue is not fixed, or as long as we can't safely ignore the result of testing against `latest-develop`.

rflechtner:
> LGTM! Unfortunately, the merging of [KILTprotocol/mashnet-node#316](https://github.com/KILTprotocol/mashnet-node/pull/316) into `develop` means that integration tests now are again broken, because something changed in the storage layout. It is definitely not to be fixed in this PR, but I guess we will keep having problems as long as the issue is not fixed, or as long as we can't safely ignore the result of testing against `latest-develop`.

Technically the tests against latest-develop are optional, so this may hit us only later when these changes are released - although the GitHub Actions status has behaved somewhat inconsistently in the past

PR #509, opened 2022-03-22
fix: add authentication key requirement for didLookup pallet calls

ntn-x2:
The DID logic did not include any DID key for the `didLookup` pallet, which accepts a DID origin with an authentication key.

## How to test

Use `"@kiltprotocol/sdk-js": "0.26.1-beta.1"` against a version of the node that has the fixed origin, e.g., https://github.com/KILTprotocol/mashnet-node/pull/343.


rflechtner (approved):
Good catch! LGTM

PR #514, opened 2022-03-29
feat: support for Did<->Account links

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/1921
Adds support for (un)linking accounts to Dids.

## How to test:

Run integration tests for the `AccountLinking` module.

## Checklist:

- [x] Fix the hardcoded `u64` in the tuple encoding so that it uses the **chain**-defined `BlockNumber` type
- [x] Fix code OR integration tests so that it is possible to link a generic Substrate account that is not encoded with the KILT network prefix
- [x] Replace the `Signer` interface with an equivalent closure


packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,243 @@
...
+// TODO: update with string pattern types once available
+type AccountAddress = IIdentity['address']
+type DidUri = FullDidDetails['did']
+
+interface ConnectionRecord extends Struct {
+  did: AccountId
+  deposit: Deposit
+}
+
+interface DepositJSON {
```
ntn-x2, line 37: We have this defined in `types/src/Deposit.ts`.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,243 @@
...
+import type { Null, Option, Struct } from '@polkadot/types'
+import type { AccountId, Extrinsic } from '@polkadot/types/interfaces'
+import type { AnyNumber } from '@polkadot/types/types'
+import type { HexString } from '@polkadot/util/types'
+
+import type { DidAuthorizationOptions, FullDidDetails } from '../DidDetails'
+import { getKiltDidFromIdentifier, parseDidUri } from '../Did.utils.js'
+
+// TODO: update with string pattern types once available
+type AccountAddress = IIdentity['address']
```
ntn-x2, line 29: I would call this `KiltAccountAddress` or similar, to distinguish it from the generic `AccountId`, which can be from any other chain.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,243 @@
...
+  const { owner, amount } = connectedDid.unwrapOrDefault().deposit
+  return {
+    amount: amount.toNumber(),
+    owner: connectedDid.isSome ? owner.toString() : null,
+  }
+}
+
+export async function getConnectedDidForAccount(
+  linkedAccount: AccountId | string
+): Promise<DidUri | null> {
```
ntn-x2, line 60: This is a chain file, so I am kind of against having types such as `DidUri`, when it should be just a `DidIdentifier`. If we want to use `DidUri`, we should at least also add functions that simply take or return the `DidIdentifier`.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,243 @@
...
+ * Builds an extrinsic to link `account` to `did`, where `account` must hold balance to cover for the transaction fees and deposit.
+ *
+ * @param account KeyringPair of the account to be linked, used for signing the did-authorized extrinsic.
+ * @param did FullDid to be linked, used for did-authorization.
+ * @param didSignCallback Keystore or callback that manages the FullDid's keys.
+ * @param signingOpts Options to be passed on to `FullDid.authorizeExtrinsic` & `Extrinsic.signAsync`.
+ * @returns A signed SubmittableExtrinsic ready to be submitted.
+ */
+export async function linkSubmitterAccount(
+  account: KeyringPair,
```
ntn-x2, line 188: I think we purposefully removed all instances of `KeyringPair`, since we most likely don't have access to the private key directly. Perhaps we should stick to that, and let the users handle the extrinsic the way they find best?

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: I don't think we care how the address is encoded, that should all go through actually

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: I am not questioning its utility, but we don't have this type of function anywhere else in the chain files. It might make sense, but perhaps in a different file, if any.

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: The point is, in my opinion, that this is a chain file, and as a chain file it should not know how to create a full DID given the identifier. That is unnecessary coupling. So a good compromise would be provide both the raw identifier and the compounded DID.

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: this was a decoded type but I guess we don't necessarily need to decode

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: ok, no more didUris

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: No, it already did very little (three lines essentially), so this is something that users can just put together themselves.

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: Going back to this issue, it now appears that if you give a generic Substrate address which is automatically encoded by the SDK, and then try to query the DID for that address, it fails as the address is not automatically encoded upon querying. There is an integration test case that is currently failing because of this.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,272 @@
...
+  did: Address
+  deposit: Deposit
+}
+
+/// Type of signatures to link accounts to DIDs.
+export type SignatureType = MultiSignature['type']
+
+/// Type of a linking payload signing function.
+export type LinkingSignerCallback = (
+  payload: Uint8Array,
```
tjwelde, line 48: Since the polkadot/extension `signRaw` function expects a `HexString` and returns an object with a signature, which is also a `HexString`, I think it would be better to define it here like this.

tjwelde, line 48: Also, since there is a `signPayload` and a `signRaw` function on the extension interface and we actually need the raw signing, I would suggest renaming it to something like `rawData`, or making a type alias, which reflects that.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,272 @@
...
+  address: KiltAddress
+) => Promise<Uint8Array>
```
tjwelde, line 50: I don't think it is clear, what we expect in return.
I suggest adding documentation, or a type alias reflecting, that we expect a signature as hex.

tjwelde (changes requested):

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: Something like this? https://github.com/KILTprotocol/sdk-js/pull/514/commits/0b3109da2b3df72208b025a936767dc290766c2d

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/514/commits/0b3109da2b3df72208b025a936767dc290766c2d

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -262,11 +262,14 @@ export async function authorizeLinkWithAccount(
   // Gets the current definition of BlockNumber (second tx argument) from the metadata.
   const blockNumberType =
     api.tx.didLookup.associateAccount.meta.args[1].type.toString()
-  // Gets the DidIdentifier definition from the metadata by unpacking a type from this pallet.
-  const didIdentifierType = api
-    .createType<Struct>('PalletDidLookupConnectionRecord')
-    .get('did')
-    ?.toRawType()
+  // This is some magic on the polkadot types internals to get the DidIdentifier definition from the metadata.
+  // We get it from the connectedDids storage, which is a double map from (DidIdentifier, Account) -> Null.
```
ntn-x2, line 266:
```suggestion
  // We get it from the connectedAccounts storage, which is a double map from (DidIdentifier, Account) -> Null.
```

rflechtner: true, this is outdated

ntn-x2:
@rflechtner can we also rename the query functions from `get...` to `query...`, as is the case for the other chain files?

rflechtner:
> @rflechtner can we also rename the query functions from `get...` to `query...`, as is the case for the other chain files?

makes sense

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,302 @@
...
+ * Return all the accounts linked to the provided DID identifier.
+ *
+ * @param linkedDid The DID to use for the lookup.
+ * @param networkPrefix The optional network prefix to use to encode the returned addresses. Defaults to generic Substrate addresses (no network-specific prefix).
+ * @returns The list of accounts linked to the DID. All accounts are encoded using the KILT 38 network prefix.
+ */
+export async function getConnectedAccountsForDid(
+  linkedDid: IDidIdentifier,
+  networkPrefix: number | undefined = undefined
+): Promise<Array<KiltAddress | SubstrateAddress>> {
```
rflechtner, line 94: The documentation is not accurate here rn. Shouldn't we at least default to the Kilt prefix, if we allow people to choose the address encoding at all?
@ntn-x2 @tjwelde

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -0,0 +1,308 @@
...
+    .toHex()
+  let signature = u8aToU8a(await signingCallback(signMe, accountAddress))
+  const { crypto, isValid } = signatureVerify(signMe, signature, accountAddress)
+  if (!isValid) throw new Error('signature not valid')
+  const sigType = getMultiSignatureTypeFromKeypairType(crypto as KeypairType)
+  try {
+    // The signature may be prefixed; if it is, it would still verify after removing the first byte (prefix).
+    // In that case, we get rid of the prefix.
+    if (
+      signatureVerify(signMe, signature.subarray(1), accountAddress).isValid
```
tjwelde, line 295: Wouldn't it be more effective to first remove the byte and do a verification and if it is valid, continue, without doing the second verification?
We would only do the second verification than in case the one with the removed byte failed, instead of always.

rflechtner: Yeah that’s the one thing I still wanted to change, would be nicer

ntn-x2: I can change that around.

ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/514/commits/e513d257e98878cc502f3620df71824e5045807e

ntn-x2: It is a bit tricky because a verifying signature can or cannot have the type annotation. So if we want to favor success cases over failures, this is what I was able to come up with.

rflechtner: can we not do something like
```
let result;
try {
  result = verifySignature(signature.subarray(1))
  assert(result.isValid) // can also be if -> throw
  signature = signature.subarray(1)
} catch {
  result = verifySignature(signature)
  assert(result.isValid)
}
const {crypto} = result
```
This way we still only verify once if successful, and it's a lot shorter?

ntn-x2: Otherwise, this probably looks more elegant and concise.

rflechtner: absolutely, we shouldn't let ourselves be held back by optimisations. on the issue of throwing, as far as I remember the error messages coming from verifySignature were actually quite helpful, so I wouldn't mind throwing them. But I can also understand if you want to catch and wrap those.

ntn-x2: Your proposal makes sense. And since we have to add the `<Bytes>...</Bytes>` padding to the default implementation, I think I will integrate your feedback.

ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/514/commits/b4ddb81e5100d8a81e1911819d7722823e43b57d includes Raphael's feedback, while https://github.com/KILTprotocol/sdk-js/pull/514/commits/369ec599abc022c33eb520a69e6e7fb6bb8d1c5b wraps the default and test signing callback with the needed `wrapBytes` call.

rflechtner:
just a thought: if we'd wrap `signMe` before handing it to the signing callback instead of wrapping within the callback, we could have support e.g. by sporran out-of-the-box. I'm pretty sure `u8aWrapBytes` checks if the input is wrapped and does not re-wrap, which would be a requirement of course

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: @rflechtner integrated your suggestion -> https://github.com/KILTprotocol/sdk-js/pull/514/commits/1c9a199eb8c696c9967cdfb5a8ca68f0354aeff4

It makes sense we expect the same on the blockchain, so good to have it padded before asking the callback to sign it.

PR #517, opened 2022-04-11
feat: use testcontainers for integration testing

rflechtner:
## fixes a big headache
This handy tool spins up individual containers for each integration test file running concurrently. No more interference in these tests, no more spinning containers up manually.  

## How to test:
Start the docker daemon, run `yarn test:integration`, and be amazed.

## Checklist:


package.json
```diff
@@ -22,8 +22,9 @@
     "style:fix": "yarn style --write",
     "test": "jest --coverage --group=unit",
     "test:ci": "yarn test --ci",
-    "test:integration:run": "jest --group=integration -b -w 3 --silent",
-    "test:integration": "{ lsof -i :9944 > /dev/null && yarn test:integration:run; } || echo 'Can not connect to chain. Is it running?'; exit 1;",
+    "test:integration": "jest --group=integration -w 3 --testTimeout=30000",
```
tjwelde, line 25: Q: why do we need to add a timeout? Is the default not sufficient?

tjwelde: also the `-b` option was removed. Why is that?

tjwelde:
just some questions above, overall I think it is great ;)

package.json
rflechtner: Oh, yes, I figured we don't actually want it to bail (fail fast) unless we test in ci mode, because that means we wouldn't known what else is broken. At least it was annoying me quite a bit.

rflechtner: On the timeout, it does work without this but I though 30s is a reasonable default for our integration tests. And because we don't have a separate jest config for integration testing, I set it here.

tjwelde (approved):
LGTM

PR #518, opened 2022-04-12
feat: add renovate config

rflechtner:
## fixes KILTProtocol/ticket#1832

This adds a renovate config to the repo. Repo maintainers would have to add the [renovate app](https://github.com/marketplace/renovate) to the repo for this to have an effect.

## How to test:
I set this up on a fork so I can test it. Look at the PRs it has created, works quite nicely.

## Checklist:

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -2,7 +2,8 @@
   "extends": [
     ":preserveSemverRanges",
     ":disableDevDependencies",
-    ":disablePeerDependencies"
+    ":disablePeerDependencies",
+    ":semanticCommits"
   ],
   "enabledManagers": ["npm"],
   "semanticCommitType": "chore",
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -15,7 +15,7 @@
     "yarnDedupeHighest"
   ],
   "prConcurrentLimit": 10,
-  "prHourlyLimit": 2,
+  "prHourlyLimit": 10,
   "packageRules": [
     {
       "matchUpdateTypes": [
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,5 @@
 {
   "extends": [
-    ":prHourlyLimit2",
-    ":prConcurrentLimit10",
     ":preserveSemverRanges",
     ":disableDevDependencies",
     ":disablePeerDependencies"
@@ -16,6 +14,8 @@
   "postUpdateOptions": [
     "yarnDedupeHighest"
   ],
+  "prConcurrentLimit": 10,
+  "prHourlyLimit": 2,
   "packageRules": [
     {
       "matchUpdateTypes": [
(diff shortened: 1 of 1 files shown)


tjwelde:
It looks _very_ nice. Example PR in raphaels fork: https://github.com/rflechtner/kilt-sdk/pull/8

tjwelde:
and OMG, it really supports updating polkadot dependencies at the same time 🤯
https://github.com/rflechtner/kilt-sdk/pull/6
I am ecstatic!

.github/renovate.json
```diff
@@ -0,0 +1,41 @@
+{
+  "extends": [
+    ":preserveSemverRanges",
+    ":disableDevDependencies",
+    ":disablePeerDependencies",
+    ":semanticCommits"
+  ],
+  "enabledManagers": ["npm"],
```
tjwelde, line 8: works better than yarn?

tjwelde (approved):
LGTM

tjwelde:
Do we need to merge before setting up the renovate app?

.github/renovate.json
rflechtner: this is a bit misleading, it's actually an opt-in for all tooling pertaining to the npm ecosystem. So it does use yarn because we have a `yarn.lock`, not a `package.lock`

.github/renovate.json
```diff
@@ -0,0 +1,41 @@
...
+  "semanticCommitType": "chore",
+  "labels": [
+    "dependencies"
+  ],
+  "dependencyDashboard": false,
+  "updateLockFiles": true,
+  "postUpdateOptions": [
+    "yarnDedupeHighest"
```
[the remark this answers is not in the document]
rflechtner: right???

rflechtner:
> Do we need to merge before setting up the renovate app?

No, I'd actually recommend adding the renovate app first, then merging this PR.

.github/renovate.json
tjwelde: ah ok

tjwelde:
@rflechtner two questions:
1. What about the base config renovatebot comes with? should we keep some stuff of that? E.g. the recommended groupings: https://docs.renovatebot.com/presets-config/#configbase ?
2. why not activate the dependency dashboard?

rflechtner:
> @rflechtner two questions:
> 
> 1. What about the base config renovatebot comes with? should we keep some stuff of that? E.g. the recommended groupings: https://docs.renovatebot.com/presets-config/#configbase ?
> 2. why not activate the dependency dashboard?

Makes sense to include `:ignoreModulesAndTests`, `workarounds:typesNodeVersioning`. On the groupings, at least the monorepos group seems to make a lot of sense, I'm just not sure whether it messes with our config.
I decided against the dashboard because it would live on the sdk repo as a ticket, which is unusual for our workflows, but we can certainly enable it if you think it helps!

PR #520, opened 2022-04-20
refactor: consistent naming

weichweich:
BREAKING CHANGE: rename `authorizeLinkWithAccount` to `getAuthorizeLinkWithAccountTx`

## No Ticket

One is not like the others!

```
function getAssociateSenderTx(): Promise<Extrinsic>
function getAccountSignedAssociationTx(): Promise<Extrinsic>
function getReclaimDepositTx(): Promise<SubmittableExtrinsic>
function getLinkRemovalByAccountTx(): Promise<SubmittableExtrinsic>
function getLinkRemovalByDidTx(): Promise<Extrinsic>
function authorizeLinkWithAccount(): Promise<Extrinsic>
```

All function that return an extrinsic are called `get...Tx` except this one, which is confusing. I expected that this does something else. 😅

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/AccountLinking.spec.ts
+++ b/packages/core/src/__integrationtests__/AccountLinking.spec.ts
@@ -180,11 +180,12 @@ describe('When there is an on-chain DID', () => {
       }, 40_000)
 
       it('should be possible to associate the account while the sender pays the deposit', async () => {
-        const linkAuthorisation = await AccountLinks.authorizeLinkWithAccount(
-          keypair.address,
-          did.identifier,
-          signingCallback
-        )
+        const linkAuthorisation =
+          await AccountLinks.getAuthorizeLinkWithAccountTx(
+            keypair.address,
+            did.identifier,
+            signingCallback
+          )
         const signedTx = await did.authorizeExtrinsic(
           linkAuthorisation,
           keystore,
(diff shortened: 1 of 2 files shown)


ntn-x2 (approved):
LGMT 👀👀👀

packages/core/src/__integrationtests__/AccountLinking.spec.ts
```diff
@@ -180,11 +180,12 @@ describe('When there is an on-chain DID', () => {
       }, 40_000)
 
       it('should be possible to associate the account while the sender pays the deposit', async () => {
-        const linkAuthorisation = await AccountLinks.authorizeLinkWithAccount(
-          keypair.address,
-          did.identifier,
-          signingCallback
-        )
+        const linkAuthorisation =
+          await AccountLinks.getAuthorizeLinkWithAccountTx(
```
wischli, line 184: Do we follow British or American English in the SDK? Before it was British (e.g. _Authori**s**ation_), now it's American (_authori**z**e_).

In the mashnet-node, we apply American as we follow Substrate.

wischli (commented):
This might introduce language inconsistency.

packages/core/src/__integrationtests__/AccountLinking.spec.ts
tjwelde: @Dudleyneedham tries to sneak in british english, but I would prefer US

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -277,7 +277,7 @@ export function defaultSignerCallback(keyring: Keyring): LinkingSignerCallback {
  * @param nBlocksValid How many blocks into the future should the account-signed proof be considered valid?
  * @returns An Extrinsic that must be did-authorized by the [[FullDid]] whose identifier was used.
  */
-export async function authorizeLinkWithAccount(
+export async function getAuthorizeLinkWithAccountTx(
```
tjwelde, line 280: If a function returns `Extrinsic`, instead of `SubmittableExtrinsic`, the convention is to call the function `get...Extrinsic`

tjwelde (changes requested):

rflechtner (commented):
Mhhh interesting. This did not follow the convention because I saw this convention as applying to the extrinsic builders, while this is a function that helps with signing and passes the result to the extrinsic builder. But on the other hand it’s ultimately just another function that builds an extrinsic, so I see your point! I’m undecided here

tjwelde:
@rflechtner At one place we used `generate...Extrinsic` for a function which does more than just returning the right extrinsic. What do you think of that?

tjwelde (approved):

PR #521, opened 2022-04-20
fix: check service endpoints in builder

weichweich:
## No Ticket

Check the service endpoints in the `FullDidUpdateBuilder`.

## Checklist:

### DIFF
--- a/packages/did/src/Did.chain.ts
+++ b/packages/did/src/Did.chain.ts
@@ -356,7 +356,7 @@ export function formatPublicKey(key: NewDidKey): PublicKeyEnum {
   return { [type]: publicKey }
 }
 
-function checkServiceEndpointInput(
+export function checkServiceEndpointInput(
   api: ApiPromise,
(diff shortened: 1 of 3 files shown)


packages/did/src/Did.chain.ts
```diff
@@ -356,7 +356,7 @@ export function formatPublicKey(key: NewDidKey): PublicKeyEnum {
   return { [type]: publicKey }
 }
 
-function checkServiceEndpointInput(
+export function checkServiceEndpointInput(
```
weichweich, line 359: this was a quick hack. I don't know what the correct approach would be to share this method.

rflechtner:
Mind if I pull that into https://github.com/KILTprotocol/sdk-js/pull/506 ?

weichweich:
> Mind if I pull that into https://github.com/KILTprotocol/sdk-js/pull/506 ?

Do that! :)

PR #522, opened 2022-04-28
feat: add RPC decorations

weichweich:
## fixes [1953](https://github.com/KILTprotocol/ticket/issues/1953)

Add custom RPC decorations to the SDK.

We should think how we can include this in the Polkadot Apps. Maybe we need to revive the [type-definitions](https://github.com/KILTprotocol/type-definitions) again. 😖 

## Checklist:


ntn-x2:
Superseded by https://github.com/KILTprotocol/sdk-js/pull/590.

PR #524, opened 2022-05-03
feat: make id of kilt ctype schema resolvable ipns link

rflechtner:
## fixes KILTProtocol/ticket#1917

By publishing the ctype metaschema on ipfs and linking to it via ipns, it is now resolvable.
The key for the ipns name can be found in 1Password.

## How to test:
- Try resolving schema (e.g. `ipns://k51qzi5uqu5dkglos1mtdukd4axyhwav7e98bga8g2nptrkgcbj9506ruoadiz/v1/ctype.json`) 
  - [ ] works via local ipfs node 
  - [ ] works via http gateway.
- Try importing ipns key from 1Password
  - [ ] works

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -243,7 +243,7 @@ Creating a CType requires specifying its structure as a [JSON schema](https://js
 
 ```typescript
 const ctype = Kilt.CType.fromSchema({
-  $schema: 'http://kilt-protocol.org/draft-01/ctype#',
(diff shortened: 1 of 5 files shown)


trusch:
@rflechtner can NOT retrieve the data anymore through the ipns entry

trusch:
I *think* you will need to import the private key of the ipns entry and import it on our ipfs node in our cluster. If I understand it correctly, there is only a short caching period and after that people looking for the data have to ask your node (the one which holds the private key to the public key from the ipns address) for the up-to-date data.

So therefore I can't resolve now, because your laptop is not online.

rflechtner:
This breaks existing CTypes and is superseded by https://github.com/KILTprotocol/ticket/issues/2378

PR #534, opened 2022-05-10
docs: add docstrings to all public functions and methods

rflechtner:
## fixes KILTProtocol/ticket#450
Enables linter rule requiring jsdoc on all public functions and methods.
Adds docstrings wherever linter complained. 

## Checklist:


tjwelde (approved):
I skimmed through it and really like what I see.
Also the documentation, which function actually throws and the changing of functions, which throw AND return boolean for validation.

LGTM!

PR #535, opened 2022-05-10
feat: convert SDKErrors to just extend Error

tjwelde:
## fixes KILTProtocol/ticket#1984
This PR converts all SDKError functions into classes extending the base Error class.
It also removes a bunch of unused SDKErrors.

## How to test:
- Run tests

## Checklist:


rflechtner (commented):
I have a few questions:
1. Did you think about multiple layers of inheritance and if so, why did you decide against it? Specifically I'm thinking it may be a good idea to extend a abstract error class like `SDKError` instead of just `Error` to allow detecting whether an error is raised within our code or by a dependency.
2. Another idea would be to additionally replicate the semantic grouping formerly expressed with the initial digits of the error code with abstract error classes, e.g. SDKError -> MissingDataError-> ERROR_CTYPE_HASH_NOT_PROVIDED
3. A lot of people also set the `name` property to the error class name. Not sure what the exact reason is here, but could provide a fallback for rare cases where `instanceof` does not work. Did you explicitly decide against that?

tjwelde:
Regarding
1: it's a cool idea, but I don't think it is worth it here. Applications will have to react on specific errors either way.
2: pretty much the same
3: the class name is taken as the name field per default and since they are already speaking, I would not add it additionally

tjwelde:
After further experimentation and discussion, we decided to use following Snippet for resolving @rflechtner points 1 and 3 without a lot of effort:
```
class SDKError extends Error {
    constructor(m) {
        super(m)
        this.name = this.constructor.name
    }
}```

rflechtner (approved):
Love it!

PR #536, opened 2022-05-10
chore(tsconfig): only include packages of interest

tjwelde:
## fixes KILTProtocol/ticket#1985
This PR changes the development tsconfig to only include relevant folders to prevent over-reporting of `yarn check`

## How to test:
- Run `yarn check`

## Checklist:


rflechtner (commented):
TBH I don't known enough about the tsconfig to really understand the implications. How does this interact with `baseUrl`? Essentially this is meant to avoid checking `tests` in the root directory, right?

PR #542, opened 2022-05-31
feat: do not require process to be available

tjwelde:
## fixes KILTProtocol/ticket#2037
This PR removes the necessity for the `process` variable to be available in `ConfigService`.
It also removes all seemingly unused polyfills and adds a test for messaging to the sdk bundle tests.

## How to test:
- Bundle tests have to go through

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "style": "prettier -l packages",
     "style:fix": "yarn style --write",
     "test": "jest --coverage --group=unit",
-    "test:ci": "yarn test --ci",
+    "test:ci": "yarn test --ci --forceExit",
     "test:integration": "jest --group=integration -w 3 --testTimeout=30000",
     "test:integration:ci": "jest --group=integration -b -w 3 --testTimeout=60000",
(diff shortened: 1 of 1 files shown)


tests/bundle-test.js
```diff
@@ -219,6 +221,31 @@ async function runAll() {
...
+    },
+    type: MessageBodyType.REQUEST_ATTESTATION,
+  }
+
+  const message = new Message(body, bob.uri, alice.uri)
+  const encryptedMessage = await message.encrypt(
+    bob.encryptionKey.id,
+    bob,
+    keystore,
+    `${alice.uri}#${alice.encryptionKey.id}`
```
rflechtner, line 237: we have
```suggestion
    alice.assembleKeyUri(alice.encryptionKey.id)
```
but still, this is crappy UI, you're right

rflechtner (approved):
I know very little about polyfills, but if the bundle tests work I guess we're good!

tests/bundle-test.js
tjwelde: yeah, I think we need some good defaults for this

PR #551, opened 2022-06-10
feat: convert bundle test to TS

arty-name:
Converted bundle test to TS to improve code quality

## How to test:
yarn bundle && yarn test:bundle

## Checklist:


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "yarn workspaces foreach -p --topological-dev --exclude '{root-workspace}' run build",
     "build:docs": "typedoc --theme default --out docs/api --tsconfig tsconfig.docs.json && touch docs/.nojekyll",
     "bundle": "yarn workspace @kiltprotocol/sdk-js run bundle",
-    "clean": "yarn workspaces foreach -p --exclude '{root-workspace}' run clean",
+    "clean": "rm -rf tests/dist && yarn workspaces foreach -p --exclude '{root-workspace}' run clean",
```
rflechtner, line 14: we're using `rimraf` elsewhere, I think that is to be more platform-agnostic

rflechtner (commented):
Given that playwright is already ts enabled, can we not import the test script in the spec file and use [`page.evaluate`](https://playwright.dev/docs/evaluating) to run the code in the browser context instead of adding the script to the html directly? This would have playwright transpile it on the fly, I assume.

arty-name:
Handing off the compilation to the playwright is smart, and I haven’t thought about it. This approach, however, would make it impossible to just open the page in the browser and run the code manually. This was an important tool for me when debugging things. Should I try for the playwright approach anyway?

rflechtner:
> Handing off the compilation to the playwright is smart, and I haven’t thought about it. This approach, however, would make it impossible to just open the page in the browser and run the code manually. This was an important tool for me when debugging things. Should I try for the playwright approach anyway?

You got a point there, and I would not know about debugging because I haven't worked with that code, so you know better if it's an essential tool. I imagine we could retain debugging options if we split the code execution in two; assign the `runAll` function to the window object on the first and run it on the second. This way you'd be able to pause execution in between and play around with the code in the browser, at least when testing headful?

package.json
rflechtner: yeah I'd say so. I don't know the exact reasons why we're working with `rimraf`, but whatever it solves, it should solve it here too.

arty-name:
I don’t even have the playwright installed, and I can open the html file and run the code now. I feel it would be practical to keep it this way.

package.json
arty-name: Done!

rflechtner (approved):
LGTM!

tests/bundle-test.ts
```diff
@@ -259,9 +289,8 @@ async function runAll() {
   )
   await BlockchainUtils.signAndSubmitTx(authorizedAttTx, devFaucet, {
     resolveOn: BlockchainUtils.IS_IN_BLOCK,
-    reSign: true,
   })
-  if (credential.verify()) {
+  if (await credential.verify()) {
```
tjwelde, line 293: Good catch

PR #555, opened 2022-06-14
feat: remove Keystore

arty-name:

To avoid over specifying the way to handle private keys, we’d like to replace the usage of Keystore with simple callbacks handling the cryptographic operations. 

## How to test:

yarn test

## Checklist:


packages/types/src/Keystore.ts
```diff
@@ -48,98 +49,40 @@ export interface ResponseData<A extends string> {
...
   alg: A
 }
 
 /**
- * A Keystore instance which supports {en/de}cryption using nacl's 'x25519-xsalsa20-poly1305' algorithm.
+ * Methods for {en/de}cryption using nacl's 'x25519-xsalsa20-poly1305' algorithm.
  */
-export interface NaclBoxCapable
-  extends Keystore<any, 'x25519-xsalsa20-poly1305'> {
+export interface NaclBoxCapable {
```
tjwelde, line 82: Should we extract the `encrypt` and `decrypt` properties here into two separate callbacks as we did with the SigningCallback, too?

packages/testing/src/TestUtils.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Generates a keypair usable for signing and a few related values.
+ *
+ * @param alg The algorithm to use for the keypair.
+ * @returns The keypair, matching sign callback, a key usable as DID authentication key.
+ */
+export function makeSigningKeypair(
+  alg: SigningAlgorithms = SigningAlgorithms.Sr25519
+): KeyTool {
```
tjwelde, line 133: `makeEncryptionKeypair` returns just the keypair, whereas this function returns KeyTool, which includes more than that. Can we either make the return types consistent (e.g. by returning an encryption keytool from `makeEncryptionKeypair`, which includes the encrypt and decrypt callbacks) or change the naming of this function?

packages/core/src/credential/Credential.spec.ts
```diff
@@ -454,49 +452,40 @@ describe('create presentation', () => {
...
+        ...newKeyForMigratedClaimerDid.authenticationKey,
         id: 'new-auth',
       }
     )
-    const migratedThenDeletedKey = await keystore.generateKeypair({
-      alg: SigningAlgorithms.Ed25519,
-      seed: '//MigratedThenDeletedClaimer',
-    })
+    const migratedThenDeletedKey = makeSigningKeypair(SigningAlgorithms.Ed25519)
+    signMigratedThenDeletedClaimerLightDid = makeSignCallback(
```
tjwelde, line 484: can use sign callback from `makeSigningKeypair()`

packages/types/src/Keystore.ts
arty-name: Yes, I was considering this, will do!

packages/testing/src/TestUtils.ts
arty-name: I expected this question :) Should they become `makeSigningKeyTool` and `makeEncryptionKeyTool`?

tests/bundle-test.ts
```diff
@@ -41,38 +48,118 @@ function getDefaultMigrationHandler(submitter: KeyringPair) {
   }
 }
 
-async function createFullDidFromSeed(
-  identity: KeyringPair,
-  keystore: InstanceType<typeof Did.DemoKeystore>,
+function makeSignCallback(keypair: KeyringPair): SignCallback {
```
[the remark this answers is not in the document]
arty-name: Exactly!

PR #556, opened 2022-06-15
fix: deposit amount test

arty-name:
Makes the deposit amount test accept the old and the new value while we’re in transition

## How to test:

yarn run test:integration packages/core/src/__integrationtests__/Attestation.spec.ts

## Checklist:


rflechtner (approved):
Works for me! Can be reverted when master has the new deposit too

PR #557, opened 2022-06-15
feat: support account link pallet update

rflechtner:
## fixes KILTProtocol/ticket#2048

This adds discovery of whether the account linking pallet has updated to support ethereum alongside substrate addresses.
Extrinsic and query parameter encoding as well as decoding of results will adapt automatically.

## How to test:

As long as https://github.com/KILTprotocol/kilt-node/pull/355 is not yet merged, testing will have to be done against an image built from source on that branch.
Tests for ethereum address linking have therefore been set up to be skipped.
Integration tests against latest & latest-develop images should nevertheless still work.

## Checklist:


packages/types/src/DidDetails.ts
```diff
@@ -211,3 +211,7 @@ export type DidSignature = {
   keyUri: DidResourceUri
   signature: string
 }
+
+export type JsonEnum<K extends string, V> = K extends any
```
tjwelde, line 215: This looks just like `Record`. What does it do differently and is it important enough to have a new type for it? Because  it makes things more confusing.

tjwelde: ok, I played around with a bit and `Record` would not work here.
The only solution I could come up with is:
```
export type JsonEnum<K extends string, V> = {
  [P in K]?: V
}
```
Which I like more, because it is more straightforward and understandable. Only downside is, that it allows empty objects, but I don't see a way around that. On the other side,`makeJsonEnum` would not be needed anymore,  which currently uses type assertion with `as` and is needed, because it would not be possible to assign to the JsonEnum type without it.

rflechtner: AFAIK it's the __only__ way to turn `'a' | 'b'` into something like `{a: any} | {b: any}`, which is the sort of type that polkadot requires for the creation of enums. What you proposed gives you `{a?: any, b?: any}` which besides empty objects also allows `{a: 1, b: 2}`. And it is not assignable to the polkadot type, so I don't think that will work.

arty-name: Another way is to list the types explicitly: `{a: any} | {b: any}`. It is wordier and longer (a downside) but also explicit (an upside).

arty-name:
I think this should be discussed after #554 is merged, it will duplicate less code. I think the enum discussion also belongs to #554.

packages/types/src/DidDetails.ts
rflechtner: but what does that solve? if variable `a` is typed as a union of string literals, typescript interprets `{[a]: any}` as `{[key: string]: any}` and will not accept it for a signature that requires `{a: any} | {b: any}`

arty-name (approved):
Looking forward to some cleanup after Ethereum is added to all our blockchains :)

rflechtner:
> Looking forward to some cleanup after Ethereum is added to all our blockchains :)

I can see that! I hope we can develop a decent concept though for abstracting over the different storage versions, because if we ever want to query old state, we would need that again.

tjwelde (approved):
LGTM

rflechtner:
Seeing that there are efforts to find better ways to deal with polkadot/api's requirements for enum encoding, I'm merging this now so we have a better basis on which to continue discussions - because this feature makes quite some use of this pattern too, and has different requirements with regards to key type unions.

PR #558, opened 2022-06-16
test: allow running integration tests against a single endpoint

rflechtner:
This amends the integration test setup such that tests are run against a single endpoint if the `TEST_WS_ADDRESS` environment variable is set. If unset, testcontainers will be used as usual.
Implemented this to be able to run individual integration tests against a node compiled from source. Also helps with debugging, given that you can take your time to look at extrinsics through the apps.
Thought it may be useful to keep around.

## Checklist:


ntn-x2:
What do you mean by individual integration tests? As far as I can see now `initializeApi()` will create the same setup for all cases, right?

arty-name (approved):

rflechtner:
This is meant to say that one is advised to run only individual test files (or runInBand) when using this mode, given that there the possibility of interference when multiple tests run against the same node.

ntn-x2 (approved):
👌

PR #559, opened 2022-06-21
fix: accept fragment as typeguard for some methods

ntn-x2:
This is causing https://github.com/KILTprotocol/docs/pull/117 to fail when verifying the signatures of credentials published behind a service endpoint.


tjwelde (approved):
LGTM

PR #560, opened 2022-06-22
feat: conventional/semantic PR titles

arty-name:
## fixes KILTProtocol/ticket#2054

When squashing commits to merge the PR, the PR title becomes the commit message. We want our commits to be formatted as conventional/semantic commits, and this requirement propagates to the PR titles. This action should validate the PR title to follow the conventional commits guidelines.

## How to test:

Change the title of the PR to not match the conventional commits guidelines and see this action fail. This will not work on the current PR because of [specific configuration](https://github.com/amannn/action-semantic-pull-request#event-triggers), but failures and successes can be seen in the [action history](https://github.com/KILTprotocol/sdk-js/actions/workflows/semantic-pull-request.yml).

## Checklist:

### DIFF
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -1,7 +1,7 @@
 name: "Validate PR title"
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - edited
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -1,7 +1,7 @@
 name: "Validate PR title"
 
 on:
-  pull_request_target:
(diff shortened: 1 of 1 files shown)


arty-name:
Shall I also remove the commitlint to drop this requirement for individual commits?

.github/workflows/semantic-pull-request.yml
```diff
@@ -0,0 +1,16 @@
+name: "Validate PR title"
+
+on:
+  pull_request_target:
```
tjwelde, line 4: nice option. didn't know it existed

tjwelde (approved):
LGTM
If you don't mind I would keep the conventional commit checker on the commit basis for now

PR #562, opened 2022-06-23
feat: ss58Format constant

arty-name:

I have added a new constant with the value of the KILT blockchain network prefix and reused it across the SDK instead of the magic number 38.

## How to test:

yarn test

## Checklist:


packages/utils/src/ss58Format.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Copyright (c) 2018-2022, BOTLabs GmbH.
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The network SS58 prefix registered for KILT blockchain
```
arty-name, line 9: Do we also need something like @packageDocumentation here?

tjwelde: not sure. But we will have to go through the API documentation either way at the end, because some package documentations might already not work anymore.

tjwelde (approved):
LGTM

PR #565, opened 2022-06-24
feat: use arrow functions shorthand

arty-name:

The shorthand form of the arrow functions is more declarative and should be used where possible

## How to test:

yarn test

## Checklist:


tjwelde (approved):
LGTM

PR #566, opened 2022-06-24
feat: replace .then() with await

arty-name:

Use modern syntax instead of Promise callbacks for better readability

## How to test:

yarn tset

## Checklist:


tjwelde (approved):
LGTM

PR #568, opened 2022-06-27
feat: prefer named functions over arrow functions

arty-name:

The declarative function syntax for named functions has some advantages over imperative assigning of arrow functions to a constant. I have updated the few occurrences of those in the codebase.

## How to test:

yarn test

## Checklist:


tjwelde (approved):
LGTM

PR #574, opened 2022-07-06
feat: remove enums

arty-name:

[TypeScript enums](https://www.typescriptlang.org/docs/handbook/enums.html) have some downsides:
* https://github.com/typescript-cheatsheets/react#enum-types
* https://fettblog.eu/tidy-typescript-avoid-enums/

Replacing enums with union types simplifies code and results in less type casting

## Checklist:

### DIFF
--- a/packages/did/src/types.ts
+++ b/packages/did/src/types.ts
@@ -13,10 +13,9 @@ import type {
   KeyRelationship,
   NewDidEncryptionKey,
   NewDidVerificationKey,
-  VerificationKeyType,
 } from '@kiltprotocol/types'
 
-export { SigningAlgorithms, EncryptionAlgorithms } from '@kiltprotocol/types'
+export { SigningAlgorithms } from '@kiltprotocol/types'
 
 /**
  * Map from a key relationship (including the 'none' relationship) -> set of key IDs.
@@ -59,9 +58,7 @@ export type FullDidCreationDetails = {
 }
 
 // Ecdsa not supported.
-export type LightDidSupportedVerificationKeyType =
-  | VerificationKeyType.Ed25519
-  | VerificationKeyType.Sr25519
+export type LightDidSupportedVerificationKeyType = 'ed25519' | 'sr25519'
 
 /**
  * A new public key specified when creating a new light DID.
--- a/packages/types/src/Delegation.ts
+++ b/packages/types/src/Delegation.ts
@@ -9,10 +9,11 @@ import type { ICType } from './CType'
 import type { IDidDetails } from './DidDetails'
 
 /* eslint-disable no-bitwise */
-export enum Permission {
-  ATTEST = 1 << 0, // 0001
-  DELEGATE = 1 << 1, // 0010
-}
+export const Permission = {
+  ATTEST: 1 << 0, // 0001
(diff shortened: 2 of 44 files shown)


packages/types/src/Delegation.ts
```diff
@@ -9,18 +9,19 @@ import type { ICType } from './CType'
...
 /* eslint-disable no-bitwise */
-export enum Permission {
-  ATTEST = 1 << 0, // 0001
-  DELEGATE = 1 << 1, // 0010
-}
+export const Permission = {
+  ATTEST: 1 << 0, // 0001
+  DELEGATE: 1 << 1, // 0010
+} as const
+export type Permissions = typeof Permission[keyof typeof Permission]
```
tjwelde, line 16: wow. not easy to come up with that :D

tjwelde: seems to just resolve to a simple `number`. But maybe future TS versions would support more explicit typing here.

tjwelde (approved):
I like it

packages/types/src/Delegation.ts
arty-name: Interesting, it works better for strings and other types. Maybe this is already a bug in TS.

packages/did/src/types.ts
```diff
@@ -59,9 +58,7 @@ export type FullDidCreationDetails = {
 }
 
 // Ecdsa not supported.
-export type LightDidSupportedVerificationKeyType =
-  | VerificationKeyType.Ed25519
-  | VerificationKeyType.Sr25519
+export type LightDidSupportedVerificationKeyType = 'ed25519' | 'sr25519'
```
ntn-x2, line 61: Can we also define this as a subset of the `VerificationKeyType`?

ntn-x2 (commented):
I have some comments all regarding linking type definitions to each other, if I read and understand the code correctly.

packages/types/src/DidDetails.ts
```diff
@@ -30,42 +30,39 @@ export type DidUri =
...
-  Ed25519 = 'ed25519',
-  Ecdsa = 'ecdsa',
-}
-export type LightDidSupportedVerificationKeyType =
-  | VerificationKeyType.Ed25519
-  | VerificationKeyType.Sr25519
+export const verificationKeyTypes = ['sr25519', 'ed25519', 'ecdsa'] as const
+export type VerificationKeyType = typeof verificationKeyTypes[number]
+
+export type LightDidSupportedVerificationKeyType = 'ed25519' | 'sr25519'
```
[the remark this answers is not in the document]
arty-name: Yes, good point! The [126fbb8](https://github.com/KILTprotocol/sdk-js/pull/574/commits/126fbb827c1c174b15d7dbfbe66666b989d8797c) commit restores the subset relationship between this and other types

packages/types/src/Delegation.ts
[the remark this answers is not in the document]
ntn-x2: Yes, `PermissionType` sounds also better, but I am afraid it might be confusing. What about renaming the `Permission` type to `PermissionType`, and the exported type `Permission`? So basically switching them around?

ntn-x2 (approved):
Love it!

PR #575, opened 2022-07-08
feat: use DID types directly

arty-name:

The DID is a foundational feature of the KILT protocol, so I think it’s fair to treat the types `DidUri` and `DidResourceUri` as foundational types that can be used directly in most places, excluding maybe those where indirection provides helpful hints.

## Checklist:


ntn-x2 (approved):
Good point, 100% agree 👌

PR #577, opened 2022-07-08
feat!: idiomatic consistent errors

arty-name:

Errors in the JavaScript ecosystem have TitleCase names and `Error` as a suffix. The first letter of the error message is capitalized, and messages usually do not end with a dot. The cited values are put in quotes. Errors are instantiated using the `new` keyword.

As a follow-up, I suggest dropping the common namespace for errors and import them directly. This will also allow moving module-specific errors into the modules.

## Checklist:

### DIFF
--- a/packages/messaging/src/Message.utils.ts
+++ b/packages/messaging/src/Message.utils.ts
@@ -39,33 +39,28 @@ export function errorCheckDelegationData(
   const { permissions, id, parentId, isPCR, account } = delegationData
 
   if (!id) {
-    throw new SDKErrors.ERROR_DELEGATION_ID_MISSING()
-  } else if (typeof id !== 'string') {
-    throw new SDKErrors.ERROR_DELEGATION_ID_TYPE()
-  } else if (!isHex(id)) {
-    throw new SDKErrors.ERROR_DELEGATION_ID_TYPE()
+    throw new SDKErrors.DelegationIdMissingError()
+  } else if (typeof id !== 'string' || !isHex(id)) {
+    throw new SDKErrors.DelegationIdTypeError()
   }
 
   if (!account) {
-    throw new SDKErrors.ERROR_OWNER_NOT_PROVIDED()
-  } else DidUtils.validateKiltDidUri(account)
+    throw new SDKErrors.OwnerMissingError()
+  }
+  DidUtils.validateKiltDidUri(account)
 
   if (typeof isPCR !== 'boolean') {
     throw new TypeError('isPCR is expected to be a boolean')
   }
 
   if (permissions.length === 0 || permissions.length > 3) {
-    throw new SDKErrors.ERROR_UNAUTHORIZED(
+    throw new SDKErrors.UnauthorizedError(
       'Must have at least one permission and no more then two'
     )
   }
 
-  if (parentId) {
-    if (typeof parentId !== 'string') {
-      throw new SDKErrors.ERROR_DELEGATION_ID_TYPE()
-    } else if (!isHex(parentId)) {
-      throw new SDKErrors.ERROR_DELEGATION_ID_TYPE()
-    }
+  if (parentId && (typeof parentId !== 'string' || !isHex(parentId))) {
+    throw new SDKErrors.DelegationIdTypeError()
   }
 }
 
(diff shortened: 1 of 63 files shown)


packages/messaging/src/Message.utils.ts
```diff
@@ -244,19 +239,19 @@ export function errorCheckMessage(message: IMessage): void {
  *
  * @param requiredProperties The list of required properties that need to be verified against a [[CType]].
  * @param cType A [[CType]] used to verify the properties.
- * @throws [[ERROR_CTYPE_HASH_NOT_PROVIDED]] when the properties do not match the provide [[CType]].
+ * @throws [[CTypeHashMissingError]] when the properties do not match the provide [[CType]].
```
tjwelde, line 242: This seems to rather throw `CTypeUnknownPropertiesError`, plus the ones from `CType.verifyDataStructure`. Is it best practice to also mention all the possible errors of sub-functions?

packages/utils/src/SDKErrors.ts
```diff
@@ -6,348 +6,225 @@
...
  * @packageDocumentation
  */
 
 /* eslint-disable max-classes-per-file */
 
-export abstract class SDKError extends Error {
-  constructor(...args: ConstructorParameters<ErrorConstructor>) {
-    super(...args)
+export class SDKError extends Error {
+  constructor(message?: string) {
```
tjwelde, line 17: There are potentially more arguments on the Error class, which we want to use in the future, e.g. `options.cause`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

tjwelde (changes requested):

packages/utils/src/SDKErrors.ts
arty-name: Yeah, too bad the current version of TypeScript doesn’t allow them. Could we add them when this will be possible? The `...args` type seems a bit clumsy?

packages/messaging/src/Message.utils.ts
arty-name: Frankly, this is the first project I see utilizing `@throws` in the JSDoc, so I have no idea. For the consumer of the JSDoc it doesn’t make much sense to only see what the function itself throws. On the other hand, without language-level support, it is unmaintainable to list the exceptions thrown from sub-functions. However since this function doesn’t throw anything by itself, and still has `@throws` in JSDoc, I assumed that you do list exceptions of sub-functions. I could just remove this line completely.

packages/utils/src/SDKErrors.ts
tjwelde: With this commit it is now supported ;) b57c174ddbe63ef83ab92ea9dc658a21b62174fe

arty-name: I’ve updated the PR and also included causes in the places where we wrap errors

tjwelde:
Does this fix https://github.com/KILTprotocol/ticket/issues/1774 now?

packages/utils/src/SDKErrors.ts
```diff
@@ -6,348 +6,225 @@
...
-  constructor() {
-    super('DelegationId is missing')
-  }
-}
+export class DelegationIdMissingError extends SDKError {}
 
-export class ERROR_DELEGATION_SIGNATURE_MISSING extends SDKError {
+export class DelegationSignatureMissingError extends SDKError {
   constructor() {
+    // TODO: better name
```
arty-name, line 92: Still need to address this

tjwelde: In the node code we have the naming 'delegator' and 'delegate'

packages/utils/src/SDKErrors.ts
```diff
@@ -6,348 +6,220 @@
...
 
 /* eslint-disable max-classes-per-file */
 
-export abstract class SDKError extends Error {
-  constructor(...args: ConstructorParameters<ErrorConstructor>) {
-    super(...args)
+export class SDKError extends Error {
+  constructor(message?: string, options?: ErrorOptions) {
+    super(message, options)
+    // this line is the only reason for using SDKError
```
rflechtner, line 19: that and you are able to identify errors thrown by us with `instanceof SDKError`

rflechtner (approved):
LGTM!

PR #580, opened 2022-07-13
chore(deps): remove unused dependencies

rflechtner:
I went through dependencies with `depcheck` and removed what seemed unused.

## How to test:

If the tests don't break I think we're good.

## Checklist:


tjwelde (approved):
oh, very nice

PR #581, opened 2022-07-21
feat!: remove DidBatcher

arty-name:

The MVP functionality of DID batchers seems to be covered by one existing and one new utility function

## Checklist:


tjwelde (commented):
Need to review more, but I have this comment lying around for some time now, so pushing it now.

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
```diff
@@ -0,0 +1,145 @@
...
+export async function didAuthorizeExtrinsics({
+  batchFunction,
+  did,
+  extrinsics,
+  nonce,
+  selectKey = defaultSelectKey,
+  sign,
+  submitter,
+}: {
+  batchFunction: SubmittableExtrinsicFunction<'promise'>
```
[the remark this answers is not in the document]
arty-name: I have considered this and thought it would be more explicit this way, but this is not a strong opinion

packages/testing/src/TestUtils.ts
```diff
@@ -330,15 +327,27 @@ export async function createFullDidFromLightDid(
...
-  return FullDidCreationBuilder.fromLightDidDetails(api, lightDidForId)
-    .setAttestationKey(lightDidForId.authenticationKey)
-    .setDelegationKey(lightDidForId.authenticationKey)
-    .buildAndSubmit(sign, payer.address, async (tx) => {
-      await Blockchain.signAndSubmitTx(tx, payer, {
-        resolveOn: Blockchain.IS_IN_BLOCK,
-      })
-    })
+  const { authenticationKey } = lightDidForId
+  const tx = await Chain.generateCreateTxFromCreationDetails(
```
tjwelde, line 331: as mentioned on the other PR: kind of a long and convoluted name for a function you will now have to use to make a full did.

packages/utils/src/SDKErrors.ts
```diff
@@ -47,6 +47,7 @@ export class DidError extends SDKError {}
 
 export class DidExporterError extends SDKError {}
 
+// TODO: rename me
```
tjwelde, line 50: haha, saw that one removed on the other PR ;)

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+          extrinsics: [extrinsic, extrinsic],
+          sign,
+          submitter: keypair.address,
+        })
+      ).rejects.toMatchInlineSnapshot(
+        '[DidBuilderError: Can only batch extrinsics that require a DID signature]'
+      )
+    })
+
+    it('fails if the extrinsic is a utility (batch) extrinsic containing valid extrinsics', async () => {
```
tjwelde, line 57: That sounds like, it shouldn't fail, if it contains valid extrinsics? Maybe it should say something else.

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+      )
+    })
+    it.todo('successfully create a batch with only 1 extrinsic')
+    it.todo('successfully create a batch with 1 extrinsic per required key')
+    it.todo('successfully create a batch with 2 extrinsics per required key')
+    it.todo(
+      'successfully create a batch with 1 extrinsic per required key, repeated two times'
+    )
+  })
+})
```
tjwelde, line 170: What do you think about tests, with changing did keys?

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
```diff
@@ -0,0 +1,145 @@
...
+    const keyRelationship = getKeyRelationshipForExtrinsic(extrinsic)
+    if (keyRelationship === 'paymentAccount') {
+      throw new SDKErrors.DidBuilderError(
+        'Can only batch extrinsics that require a DID signature'
+      )
+    }
+    return { extrinsic, keyRelationship }
+  })
+
+  const groups: GroupedExtrinsics = [
```
tjwelde, line 62: This might throw with trying to access undefined, when `extrinsics` is an empty array. I think it would be good to check for that.

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
```diff
@@ -0,0 +1,145 @@
...
+    {
+      extrinsics: [first.extrinsic],
+      keyRelationship: first.keyRelationship,
+    },
+  ]
+
+  rest.forEach(({ extrinsic, keyRelationship }) => {
+    const currentGroup = groups[groups.length - 1]
+    const useCurrentGroup = keyRelationship === currentGroup.keyRelationship
```
tjwelde, line 71: wouldn't the extrinsics needs to be sorted by keyrelationship for this to work properly?
What if i have `assertionKey -> delegationkey -> assertionKey`?

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.spec.ts
arty-name: I carried it over from another test without many changes https://github.com/KILTprotocol/sdk-js/blob/013633a/packages/did/src/DidBatcher/DidBatchBuilder.spec.ts#L77-L83

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.spec.ts
arty-name: Again, this was carried over from https://github.com/KILTprotocol/sdk-js/blob/013633a/packages/did/src/DidBatcher/DidBatchBuilder.spec.ts#L135-L148

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
arty-name: This is not a public function however, and it’s only called with array of 2 or more. Should I still add the check?

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
arty-name: We cannot sort or otherwise change the order of extrinsics, as it affects the result. In the example you have provided we will sign the extrinsics individually without creating batches. Batches will only be created in case there is a sequence of extrinsics all requiring the same key. This is how it was already done in the DidBatchBuilder, and this is the only reason for its existence afaik.

packages/testing/src/TestUtils.ts
arty-name: I agree and I want to change it!

arty-name:
I would also like some advice on what to do with the file `packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts`. Should this utility function be simply moved into `Did.utils.ts`?

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
tjwelde: ok, let's keep it than to how it is

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
tjwelde: ah ok, didn't see that. hmm, personally still feels better to have it, just in case, but you can decide.

packages/testing/src/TestUtils.ts
tjwelde: if that stays to be the main function people will have to use to upgrade to an on-chain did, would it also make sense to put it rather in the DID, or FullDID namespace?

arty-name: I guess it should follow the pattern of `getStoreTx` functions for other data objects. For the CType and Attestation the methods from the chain file are exported on the top level. For whatever reason the methods from the DID chain file are only exported in the Chain sub-namespace. Should I re-export just `generateCreateTxFromCreationDetails` and `generateCreateTxFromDidDetails` in the DID namespace? Or should it be done in some next PR?

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
arty-name: I feel that extra checks obstruct the intention of the code, so I’d prefer to keep it out.

packages/testing/src/TestUtils.ts
arty-name: I’ve renamed it in 28659889 in the other PR

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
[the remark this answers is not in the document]
tjwelde: as discussed, documentation is added in the did class removal PR, so people will know than.

tjwelde (approved):
LGTM

PR #582, opened 2022-07-21
refactor: did enum types

tjwelde:
This PR refactors some DID enum types for more clarity.

Relates to https://github.com/KILTprotocol/ticket/issues/2040

## How to test:
- Tests need to run through
- Type system shouldn't complain

## Checklist:


arty-name (approved):
This is a significant improvement! I had some suggestions but realized that it would be simpler to showcase them in a PR, so I created #584

rflechtner (changes requested):
As mentioned in Tom's linked PR, I like the approach here to handle type system limitations on the level of the type system and reducing the number of functions that simply exists to overcome those. But we should make sure not to export a function that does not, in fact, return what it says it does.

packages/did/src/Did.chain.ts
```diff
@@ -431,12 +399,9 @@ interface IDidAuthorizedCallOperation extends Struct {
  */
 export function formatPublicKey<K extends NewDidKey>(
```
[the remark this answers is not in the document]
tjwelde: It is used in multiple other places. What would be a better name?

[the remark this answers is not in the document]
rflechtner: I'm talking about the ts return type, which is just a lie ;)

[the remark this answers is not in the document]
tjwelde: [740f7fc](https://github.com/KILTprotocol/sdk-js/pull/582/commits/740f7fca4523923e156017708363deb5c9782cca)

packages/did/src/DidBatcher/FullDidBuilder.utils.ts
```diff
@@ -13,27 +13,29 @@ import { ApiPromise } from '@polkadot/api'
...
-    keyClass = 'PublicVerificationKey'
-  } else if (isEncryptionKey(key)) {
-    keyClass = 'PublicEncryptionKey'
-  } else {
-    throw new SDKErrors.DidBuilderError('Unsupported key type')
+    return new (api.registry.getOrThrow<ChainDidPublicKey>(
+      'DidDidDetailsDidPublicKey'
+    ))(api.registry, {
+      PublicVerificationKey: encodePublicKey(key),
+    })
```
rflechtner, line 29: in the magic world of augmented apis, this can now just be 
```suggestion
    return api.createType('DidDidDetailsDidPublicKey', {
      PublicVerificationKey: encodePublicKey(key),
    })
```

packages/did/src/DidBatcher/FullDidBuilder.utils.ts
```diff
@@ -13,27 +13,29 @@ import { ApiPromise } from '@polkadot/api'
...
-    'DidDidDetailsDidPublicKey'
-  ))(api.registry, {
-    [keyClass]: formatPublicKey(key),
-  })
+  if (isEncryptionKey(key)) {
+    return new (api.registry.getOrThrow<ChainDidPublicKey>(
+      'DidDidDetailsDidPublicKey'
+    ))(api.registry, {
+      PublicEncryptionKey: encodePublicKey(key),
+    })
```
rflechtner, line 36: same here

rflechtner (approved):
I got one suggestion, other than that we're good to go!

packages/did/src/DidBatcher/FullDidBuilder.utils.ts
arty-name: FullDidBuilder is gone now anyway

packages/did/src/Did.utils.ts
```diff
@@ -264,10 +262,20 @@ export function isVerificationKey(
 export function isEncryptionKey(
   key: Partial<NewDidKey | DidKey> & Pick<NewDidKey | DidKey, 'type'>
 ): key is NewDidEncryptionKey | DidEncryptionKey {
-  const keyType = key.type.toLowerCase()
-  return encryptionKeyTypes.some((kt) => kt === keyType)
+  return encryptionKeyTypes.some((kt) => kt === key.type)
 }
 
+export type EncodedVerificationKey =
```
tjwelde, line 268: Should these maybe be moved into the types package, for ease-of-use?
Otherwise it is hard to import them, when I need it externally

tjwelde: we decided to rather reintroduce the `encodePublicKey` function.

PR #584, opened 2022-07-22
feat: remove Capitalize, Lowercase, TypedValue

arty-name:
Some suggestions for further improvements of #582. I recommend reading my comments in proper order on the conversations tab first.

## Checklist:

### DIFF
--- a/packages/did/src/Did.chain.ts
+++ b/packages/did/src/Did.chain.ts
@@ -25,10 +25,8 @@ import {
   SignCallback,
   SigningOptions,
   SubmittableExtrinsic,
-  TypedValue,
   NewDidVerificationKey,
   NewDidEncryptionKey,
-  VerificationKeyType,
 } from '@kiltprotocol/types'
 import { ConfigService } from '@kiltprotocol/config'
 import { BlockchainApiConnection } from '@kiltprotocol/chain-helpers'
@@ -37,10 +35,8 @@ import { Crypto, SDKErrors } from '@kiltprotocol/utils'
 import type {
   DidDidDetails,
   DidDidDetailsDidAuthorizedCallOperation,
-  DidDidDetailsDidEncryptionKey,
   DidDidDetailsDidPublicKey,
   DidDidDetailsDidPublicKeyDetails,
-  DidDidDetailsDidVerificationKey,
   DidServiceEndpointsDidEndpoint,
 } from '@kiltprotocol/augment-api'
 import type { DidDetails } from './DidDetails/index.js'
@@ -48,6 +44,10 @@ import type { FullDidCreationDetails } from './types.js'
 import {
   checkServiceEndpointSizeConstraints,
   checkServiceEndpointSyntax,
+  EncodedEncryptionKey,
+  EncodedKey,
+  EncodedSignature,
+  EncodedVerificationKey,
   getSigningAlgorithmForVerificationKeyType,
   getVerificationKeyTypeForSigningAlgorithm,
   isEncryptionKey,
@@ -57,10 +57,6 @@ import {
 const log = ConfigService.LoggingFactory.getLogger('Did')
 
 // ### Chain type definitions
-
-export type ChainDidKey =
-  | DidDidDetailsDidVerificationKey
-  | DidDidDetailsDidEncryptionKey
 
 export type ChainDidPublicKey = DidDidDetailsDidPublicKey
 export type ChainDidPublicKeyDetails = DidDidDetailsDidPublicKeyDetails
@@ -367,10 +363,10 @@ export type AuthorizeCallInput = {
 
 // ### EXTRINSICS
 
-type FormattedPublicKey<K extends NewDidKey> = TypedValue<
-  K['type'],
-  K['publicKey']
->
+export function encodePublicKey(
+  key: NewDidVerificationKey
+): EncodedVerificationKey
+export function encodePublicKey(key: NewDidEncryptionKey): EncodedEncryptionKey
 
 /**
  * Transforms a DID public key record to an enum-type key-value pair required in many key-related extrinsics.
@@ -378,10 +374,9 @@ type FormattedPublicKey<K extends NewDidKey> = TypedValue<
  * @param key Object describing data associated with a public key.
  * @returns Data restructured to allow SCALE encoding by polkadot api.
  */
-export function formatPublicKeyForChain<K extends NewDidKey>(
-  key: K
-): FormattedPublicKey<K> {
-  return { [key.type]: key.publicKey } as FormattedPublicKey<K>
+export function encodePublicKey(key: NewDidKey): EncodedKey {
+  // TypeScript can't infer type here, so we have to add a type assertion.
+  return { [key.type]: key.publicKey } as EncodedKey
 }
 
 function checkServiceEndpointInput(
@@ -440,15 +435,15 @@ export async function generateCreateTxFromCreationDetails(
     )
   }
 
-  const newKeyAgreementKeys = keyAgreementKeys.map(({ publicKey }) =>
-    formatPublicKeyForChain({ type: 'x25519', publicKey })
+  const newKeyAgreementKeys = keyAgreementKeys.map(
+    ({ publicKey, type }) => ({ [type]: publicKey } as EncodedEncryptionKey)
   )
 
   const newAssertionKey = assertionKey
-    ? formatPublicKeyForChain(assertionKey)
+    ? ({ [assertionKey.type]: assertionKey.publicKey } as EncodedKey)
     : undefined
   const newDelegationKey = delegationKey
-    ? formatPublicKeyForChain(delegationKey)
+    ? ({ [delegationKey.type]: delegationKey.publicKey } as EncodedKey)
     : undefined
 
   const maxNumberOfServicesPerDid =
@@ -489,11 +484,8 @@ export async function generateCreateTxFromCreationDetails(
     alg: getSigningAlgorithmForVerificationKeyType(authenticationKey.type),
   })
   const keyType = getVerificationKeyTypeForSigningAlgorithm(signature.alg)
-  const signatureForPolkadot = { [keyType]: signature.data } as TypedValue<
-    typeof keyType,
-    typeof signature.data
-  >
-  return api.tx.did.create(encodedDidCreationDetails, signatureForPolkadot)
+  const encodedSignature = { [keyType]: signature.data } as EncodedSignature
+  return api.tx.did.create(encodedDidCreationDetails, encodedSignature)
 }
 
 /**
@@ -575,7 +567,7 @@ export async function getSetKeyExtrinsic(
       }`
     )
   }
-  const typedKey = formatPublicKeyForChain(key)
+  const typedKey = { [key.type]: key.publicKey } as EncodedVerificationKey
   const api = await BlockchainApiConnection.getConnectionOrConnect()
   switch (keyRelationship) {
     case 'authentication':
@@ -649,7 +641,7 @@ export async function getAddKeyExtrinsic(
           (key as any).type
         }`
       )
-    const keyAsEnum = formatPublicKeyForChain(key)
+    const keyAsEnum = { [key.type]: key.publicKey } as EncodedEncryptionKey
     return api.tx.did.addKeyAgreementKey(keyAsEnum)
   }
   throw new SDKErrors.DidError(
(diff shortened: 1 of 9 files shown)


packages/did/src/Did.utils.ts
```diff
@@ -263,8 +264,53 @@ export function isVerificationKey(
...
+}
+
+export const verificationKeyTypeToChainKeyType: Record<
+  VerificationKeyType,
+  DidDidDetailsDidVerificationKey['type']
+> = {
+  sr25519: 'Sr25519',
+  ed25519: 'Ed25519',
+  ecdsa: 'Ecdsa',
+}
```
arty-name, line 277: I do like the explicit declarative mappings more than imperative case changes, and I want to preserve them. A difference though is the split for verification and encryption keys.

Meanwhile this also makes `getMultiSignatureTypeFromKeypairType` obsolete and allows removing the comments and knowledge about the blockchain being case-agnostic here.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -231,23 +232,6 @@ export async function getLinkRemovalByDidExtrinsic(
...
-    case 'sr25519':
-      return 'Sr25519'
-    case 'ecdsa':
-      return 'Ecdsa'
-    default:
-      throw new SDKErrors.DidError(
-        `Unsupported signature algorithm "${keypairType}"`
-      )
-  }
-}
```
arty-name, line 249: This is not needed anymore

packages/did/src/Did.utils.ts
```diff
@@ -263,8 +264,53 @@ export function isVerificationKey(
...
+  verificationKeyTypeToChainKeyType
+).reduce((obj, [key, value]) => ({ ...obj, [value]: key }), {}) as Record<
+  DidDidDetailsDidVerificationKey['type'],
+  VerificationKeyType
+>
+
+export type EncodedVerificationKey =
+  | { Sr25519: Uint8Array }
+  | { Ed25519: Uint8Array }
+  | { Ecdsa: Uint8Array }
```
arty-name, line 289: In my opinion this is much more preferable to the unholy
```ts
K extends any
  ? { [P in K]: V }
  : never

File: packages/did/src/Did.utils.ts [hunk 4]
```diff
@@ -263,8 +264,53 @@ export function isVerificationKey(
...
+).reduce((obj, [key, value]) => ({ ...obj, [value]: key }), {}) as Record<
+  DidDidDetailsDidEncryptionKey['type'],
+  EncryptionKeyType
+>
+
+export type EncodedEncryptionKey = { X25519: Uint8Array }
+
+export type EncodedKey = EncodedVerificationKey | EncodedEncryptionKey
+
+export type EncodedSignature = EncodedVerificationKey
```
arty-name, line 309: Since I suggest removing the generic TypedValue we get a tiny bit of duplication, which I am OK to live with. Makes things explicit as well.

packages/did/src/DidBatcher/FullDidBuilder.utils.ts
```diff
@@ -13,27 +13,29 @@ import { ApiPromise } from '@polkadot/api'
...
 
 function encodeToChainKey(api: ApiPromise, key: NewDidKey): ChainDidPublicKey {
-  let keyClass: string
   if (isVerificationKey(key)) {
-    keyClass = 'PublicVerificationKey'
-  } else if (isEncryptionKey(key)) {
-    keyClass = 'PublicEncryptionKey'
-  } else {
-    throw new SDKErrors.DidBuilderError('Unsupported key type')
+    return new (api.registry.getOrThrow<ChainDidPublicKey>(
```
arty-name, line 25: This duplication is not as minimal as the other one, but I still consider it a reasonable cost

packages/did/src/Did.chain.ts
```diff
@@ -387,21 +391,20 @@ interface IDidAuthorizedCallOperation extends Struct {
...
 /**
  * Transforms a DID public key record to an enum-type key-value pair required in many key-related extrinsics.
  *
  * @param key Object describing data associated with a public key.
  * @returns Data restructured to allow SCALE encoding by polkadot api.
  */
-export function formatPublicKey<K extends NewDidKey>(
-  key: K
-): FormattedPublicKey<K> {
+export function encodePublicKey(key: NewDidKey): EncodedKey {
```
arty-name, line 405: I think this is a good moment to offer this renaming. Our naming convention for preparing some data for the blockchain is "encode", and it should be used here as well. This renaming is responsible for the most of the diff here.

packages/did/src/Did.chain.ts
```diff
@@ -387,21 +391,20 @@ interface IDidAuthorizedCallOperation extends Struct {
...
 // ### EXTRINSICS
 
-type FormattedPublicKey<K extends NewDidKey> = TypedValue<
-  Capitalize<K['type']>,
-  K['publicKey']
->
+export function encodePublicKey(
+  key: NewDidVerificationKey
+): EncodedVerificationKey
+export function encodePublicKey(key: NewDidEncryptionKey): EncodedEncryptionKey
```
arty-name, line 397: These declarations provide the narrowing of types, necessary for some cases

arty-name (commented):
I think there’s no obligation to take all of these in bulk, cherry-picking should be possible as well.

packages/did/src/Did.utils.ts
rflechtner: Personally, being more familiar with Typescript than with writing JS directly, I like the TypedValue utility type a lot (note that something like Record is also just an example of such a utility type) and IMO it's clear what is happening from its documentation. While I had a small gripe with the repeated type forcing, I did like the fact that we solved what is a pure type system issue on the level of the type system - leaving no trace in the final JS code.

packages/did/src/Did.chain.ts
rflechtner: I do like the idea of using function overloads to deal with key type subsets, but we will have other subsets too. For example in the account linking pallet we will be be dealing with an additional signature type soon: `Ethereum`

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: As mentioned in my other comment, this will be extended with `Ethereum` and thus not be interoperable with what lives in the Did.utils

rflechtner:
In toto, I like the solution of using function overloads, but given that it already conflicts with the changes pending in the account linking PR, I fear removing the TypedValue may not be particularly future proof. We will have for example have signature type sets in the Did space that accept only `ecdsa`, `ed25519`, and `sr25519` (in line with the key types it can handle), while account linking additionally needs to known about `ethereum` signatures.

packages/did/src/Did.utils.ts
arty-name: Neither solution leaves a trace in the final JS code, I think, since all the types are removed. I see the value in fixing the type system issue using the type system itself, however I dislike the much larger downside of the solution being extremely obscure. The "RTFM" is pretty much never a good argument :)

packages/did/src/DidLinks/AccountLinks.chain.ts
arty-name: My solution for that case would be adding another explicit mapping, which would have an upside of being very obvious.

rflechtner:
> In toto, I like the solution of using function overloads, but given that it already conflicts with the changes pending in the account linking PR, I fear removing the TypedValue may not be particularly future proof. We will have for example have signature type sets in the Did space that accept only `ecdsa`, `ed25519`, and `sr25519` (in line with the key types it can handle), while account linking additionally needs to known about `ethereum` signatures.

Just merged the PR - you should now be able to check how it integrates with your approach

arty-name:
@rflechtner I have rebased and adapted my changes, and the approach still looks good to me. What do you think?

@tjwelde I was able to reduce the amount of type casts as well.

rflechtner (commented):
What should I say, personally I am still in favour of solving this using the type system only to get around an issue that concerns the type system only. To elaborate, this is what you would write to create an extrinsic to set a assertionMethod key in pure js:
```js
api.tx.did.setAuthenticationKey( { [key.type]: key.publicKey } )
```
This is what we do here (type definitions omitted):
```ts
const verificationKeyTypeToChainKeyType: Record<
  VerificationKeyType,
  DidDidDetailsDidVerificationKey['type']
> = {
  sr25519: 'Sr25519',
  ed25519: 'Ed25519',
  ecdsa: 'Ecdsa',
}

function encodePublicKey(key: NewDidKey): EncodedKey {
  const chainType = verificationKeyTypeToChainKeyType[key.type]
  // TypeScript can't infer type here, so we have to add a type assertion.
  return { [chainType]: key.publicKey } as EncodedKey
}

const typedKey = encodePublicKey(key)
api.tx.did.setAuthenticationKey(typedKey)
```
I don't really see how all this extra code is justified. This is all going to be still there in transpiled and bundled code and I don't see the advantage of it either. It also doesn't strike me as more readable or maintainable than just type forcing in place with something like
```ts
const typedKey = { [key.type]: key.publicKey } as ChainTypedKeyOf<typeof key>
api.tx.did.setAuthenticationKey(typedKey)
```
which will be completely gone in the final output. None of this is something consumers of the sdk would have to worry about anyway, as this concerns the internals of our own functions only. And while I don't know if that's universal, at least vsc has no issues showing you the effective type of typedKey (e.g. `{Sr25519: Uint8Array} | {Ed25519: Uint8Array} | {Ecdsa: Uint8Array}`) on hover.

rflechtner:
I'm not saying I want to keep the current implementation; I would argue for an implementation along the lines of the above (
```ts
const typedKey = { [key.type]: key.publicKey } as ChainTypedKeyOf<typeof key>
api.tx.did.setAuthenticationKey(typedKey)
```
)
which is pretty much what @tjwelde has built in https://github.com/KILTprotocol/sdk-js/pull/582. On top of what he has proposed there I think we could even get rid of `formatPublicKey` entirely and just do the `{[type]: publicKey}` wherever that was used, so it looks like the above.

arty-name:
I think however that we should discuss two concerns separately:
1) Mapping from lowercase to capitalized for the TS definitions of the blockchain
2) Making types of Rust-style enums

Regarding lowercase and capitalized, I think this should be done as close to the blockchain call as possible. Calling this line in Javascript `api.tx.did.setAuthenticationKey( { [key.type]: key.publicKey } )` is indeed OK. Types start making issues only when we extract this into a variable and make it a parameter, and pass this object around. I would rather try to do the transformations right where we call `setAuthenticationKey`. 

Alternatively, and I don’t remember this idea mentioned before, we could modify the generated type definitions themselves. Now it makes a lot of sense to me to consider the current type definitions broken: they require capitalized string, while the blockchain happily accepts lowercase string.

On the second topic of Rust-style enums, I think it’s indisputable that `TypedValue<'AccountId20' | 'AccountId32', Uint8Array>` is more contrived than `{ AccountId20: Uint8Array } | { AccountId32: Uint8Array }`. 

An even better approach (which I haven’t explored in this PR yet) would be to skip them altogether. TS doesn’t handle them nicely, and the workaround to make it work is really a brain strain. I am for creating them only at the exact location they are needed: at the place of the blockchain call. I’ll give this idea a shot.

tjwelde:
Is this now obsolete?

arty-name:
I have rebased it and I still think it contains useful improvements

packages/did/src/Did.chain.ts
tjwelde: Not a strong opinion, but I think it doesn't hurt to use that function in this file...

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -110,7 +107,7 @@ type WithEtherumSupport = {
         ) => SubmittableExtrinsic
       >
       removeAccountAssociation: AugmentedSubmittable<
-        (account: LinkableAccountJson) => SubmittableExtrinsic,
+        (account: EncodedMultiAddress) => SubmittableExtrinsic,
```
tjwelde, line 110: I like the `encoded` naming scheme

packages/did/src/DidBatcher/FullDidBuilder.utils.ts
tjwelde: yes, I think readability is improved

rflechtner (approved):
LGTM!

tjwelde (approved):
LGTM

PR #585, opened 2022-07-28
feat!: DID classes -> Document* + functions

arty-name:

Similar to the changes done to the core classes, DIDs are now just plain old javascript objects, and there are some utility functions taking them as parameters. I’ve also changed the shape of these objects to be more similar to the DID Documents.

This PR is based on the code in #581 which is not yet merged. 

This is work in progress, I created this PR early according to the retrospective action point.

## Checklist:


packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
```diff
@@ -0,0 +1,149 @@
...
+
+import { generateDidAuthenticatedTx } from '../Did.chain.js'
+import { getSigningAlgorithmForVerificationKeyType } from '../Did.utils.js'
+
+/**
+ * Type of a callback used to select one of the key candidates for a DID to sign a given batch of extrinsics.
+ */
+type SelectKeyCallback = (
+  keys: DidVerificationKey[],
+  batch?: Extrinsic[]
```
ntn-x2, line 39: If it's an array, why do we make it optional an not just default to the empty one? Wouldn't it make it easier to deal with it?

packages/types/src/DidDetails.ts
```diff
@@ -218,3 +164,15 @@ export type DidSignature = {
 export type TypedValue<K extends string, V> = K extends any
   ? { [P in K]: V }
   : never
+
+export interface DidDetails {
+  identifier: DidIdentifier
+  uri: DidUri
+
+  authentication: [DidVerificationKey]
```
ntn-x2, line 172: Why are these optional? Since they are array, I think it would be easier to make them empty instead of undefined.

packages/types/src/DidDetails.ts
```diff
@@ -218,3 +164,15 @@ export type DidSignature = {
...
+  assertionMethod?: [DidVerificationKey]
+  capabilityDelegation?: [DidVerificationKey]
+  keyAgreement?: DidEncryptionKey[]
+
+  service?: DidServiceEndpoint[]
```
ntn-x2, line 177: Technically, the ID of each service endpoint is unique in regard to other endpoints but also in regard with the keys. I am not fully happy with the fact that we model them as an array. Of course the data comes from the chain, where it is assumed to be always correct, but it would make more sense to model the services as a record (or map), and the same for the keys, since it would then provide more efficient lookups.

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -36,168 +34,146 @@ import {
...
+ * @param did The DID data.
+ * @param extrinsic The unsigned extrinsic to sign.
+ * @param sign The callback to sign the operation.
+ * @param submitterAccount The KILT account to bind the DID operation to (to avoid MitM and replay attacks).
+ * @param signingOptions The signing options.
+ * @param signingOptions.keySelection The optional key selection logic, to choose the key among the set of allowed keys. By default it takes the first key from the set of valid keys.
+ * @param signingOptions.txCounter The optional DID nonce to include in the operation signatures. By default, it uses the next value of the nonce stored on chain.
+ * @returns The DID-signed submittable extrinsic.
+ */
+export async function authorizeExtrinsic(
```
ntn-x2, line 151: So now this function takes a `DidDetails` instance, right? It means that it is not possible to prevent the `authorizeExtrinsic` to be used with light DIDs?

ntn-x2 (commented):
It was probably intentional, but now everything that a full DID could do before, can be done by a light DID since they are both `DidDetails` now. I was pretty keen on that distinction, which is now gone.

packages/did/src/DidDetails/FullDidDetails.ts
tjwelde: Maybe we can check the DID uri, to distinguish between Light and Full Dids and throw an early error here.

ntn-x2: As discussed f2f, this would be a good compromise, with the goal that stuff should fail before it makes its way to the full node.

packages/types/src/DidDetails.ts
tjwelde: What is the correct way according to spec @ntn-x2 ? It seems like the current did resolver does leave empty things out...

packages/types/src/DidDetails.ts
tjwelde: The idea is to literally work with the DID document. And the DID Document has them as an array. The performance impact on having to iterate over the array, every time you want to access a specific service endpoint is negligible.

ntn-x2: The spec defines the property as a SET, meaning that each element (identified by its ID) cannot appear more than once -> https://www.w3.org/TR/did-core/#services. Just saying that we could enforce this at compilation time since at least with Typescript.

packages/types/src/DidDetails.ts
ntn-x2: True, but that is logic that could be left to the resolver when showing a DID Document for a given DID. At the SDK level, at least in my opinion, it is confusing to have an optional array property, where you would have to check both for existence AND length.

packages/did/src/DidDetails/FullDidDetails.ts
arty-name: Good point, it will also make more explicit why we even have the distinction

packages/types/src/DidDetails.ts
arty-name: I feel that enforcing unique IDs across all DID properties is a complex task and that TypeScript is a wrong tool to solve it. Not only would this solution be incomplete, it will result in awkward data structures. With the lean SDK being a goal, I would avoid even attempting to solve the issue in it.

packages/types/src/DidDetails.ts
arty-name: For us the developers of the SDK this might be slightly inconvenient. For the users of the SDK it would be quite confusing and inconvenient to always have some unused empty arrays hanging around. It would be even worse to have to explicitly pass them somewhere. I feel that "the required array can be empty" is a poor API.

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
arty-name: Nice catch! Looks like I wanted to use the same thing for authorizeExtrinsic and it failed, so instead of fixing the DidKeySelectionCallback I made the batch here optional. It doesn’t make sense like this I think. I’ll have to fix it properly.

arty-name:
> It was probably intentional, but now everything that a full DID could do before, can be done by a light DID since they are both DidDetails now. I was pretty keen on that distinction, which is now gone.

I was uncertain whether to remove it or not. I see the point about `authorizeExtrinsic` only working with on-chain DIDs, and `getNextNonce` doesn’t make sense for the light ones as well. Batches should not be authorized either. But looking at the diff of https://github.com/KILTprotocol/sdk-js/pull/585/commits/d56a1475e1d980400df8949f298fafc1ef0f1231 I do not see many non-test places where the distinction was important. 

On the other hand, having a single data interface instead of four different ones (interface + base class + two descendants) makes things so much simpler to grasp and use.

packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -336,34 +340,44 @@ it('creates and updates DID, and then reclaims the deposit back', async () => {
...
-    const lightDidDetails = LightDidDetails.fromDetails({
+    const lightDidDetails = fromDetails({
       authenticationKey,
       encryptionKey: makeEncryptionKeyTool(
         '0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'
       ).keypair,
     })
 
-    const migratedFullDid = await lightDidDetails.migrate(
+    const creationTx = await DidChain.generateCreateTxFromCreationDetails(
```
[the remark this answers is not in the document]
rflechtner: think I found it, I assume we would recommend using `generateCreateTxFromDidDetails` after creating a light did

packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -8,245 +8,163 @@
...
+ * Create a new instance of [[LightDidDetails]] by parsing the provided input URI.
+ * This is possible because of the self-describing and self-containing nature of light DIDs.
+ * Private keys are assumed to already live in another storage, as it contains reference only to public keys.
+ *
+ * @param uri The DID URI to parse.
+ * @param failIfFragmentPresent Whether to fail when parsing the URI in case a fragment is present or not, which is not relevant to the creation of the DID. It defaults to true.
+ *
+ * @returns The resulting [[LightDidDetails]].
+ */
+export function fromUri(uri: DidUri, failIfFragmentPresent = true): DidDetails {
```
rflechtner, line 112: could also be renamed to make sure it still expresses what it creates

packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -8,245 +8,163 @@
...
+/**
+ * Create a new instance of [[LightDidDetails]] from the provided KILT address.
+ * The resulting DID will only have an authentication key, and no encryption key nor service endpoints.
+ *
+ * @param identifier The KILT address to generate the DID from.
+ * @param keyType One of the [[LightDidSupportedVerificationKeyType]]s to set the type of the authentication key derived from the provided address. It defaults to Sr25519.
+ *
+ * @returns The resulting [[LightDidDetails]].
+ */
+export function fromIdentifier(
```
rflechtner, line 159: same here

packages/did/src/DidDocumentExporter/DidDocumentExporter.ts
```diff
@@ -8,87 +8,74 @@
...
-    .map((authKey) => {
-      result.verificationMethod.push({
-        id: `${details.uri}#${authKey.id}`,
-        controller: details.uri,
-        type: VerificationKeyTypesMap[authKey.type],
-        publicKeyBase58: base58Encode(authKey.publicKey),
-      })
-      return `${details.uri}#${authKey.id}`
+  // Populate the `verificationMethod` array and then sets the `authentication` array with the key IDs
+  result.authentication = details.authentication.map((authKey) => {
```
rflechtner, line 28: not fixing https://github.com/KILTprotocol/ticket/issues/2129 here

packages/testing/src/TestUtils.ts
```diff
@@ -297,54 +283,54 @@ export async function createLocalDemoFullDidFromKeypair(
...
-      await Blockchain.signAndSubmitTx(tx, payer, {
-        resolveOn: Blockchain.IS_IN_BLOCK,
-      })
-    })
+): Promise<DidDetails> {
+  const [authenticationKey] = lightDidForId.authentication
+  const tx = await Chain.generateCreateTxFromCreationDetails(
+    {
+      identifier: DidUtils.getIdentifierByKey(authenticationKey),
+      authenticationKey,
```
rflechtner, line 311: this is a recurrent pattern I see, plus afaik you can't even have a creation extrinsic where the identifier is not derived from the authentication key – why don't we pull this into the function ?

rflechtner (commented):
I like what I see so far, got no real issues with this refactor. We might want to consider flagging the difference between light DIDs and full DIDs on the level of the type system – helps communicating which functions will fail if you try to use a light DID on them. Combined with `isLightDid` & `isFullDid` type guards this could be quite powerful.

packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -122,26 +122,24 @@ describe('write and didDeleteTx', () => {
 
     await submitExtrinsic(tx, paymentAccount)
 
-    details = (await FullDidDetails.fromChainInfo(
+    details = (await fromChainInfo(
       DidUtils.getKiltDidFromIdentifier(newDetails.identifier, 'full')
-    )) as FullDidDetails
+    )) as DidDetails
```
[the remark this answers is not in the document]
rflechtner: nitpick alert obvs

packages/did/src/DidDetails/FullDidDetails.spec.ts
```diff
@@ -169,81 +166,71 @@ describe('When creating an instance from the chain', () => {
...
+    expect(fullDidDetails.assertionMethod).toStrictEqual<DidKey[]>([
       {
         id: 'att#1',
         publicKey: new Uint8Array(32).fill(3),
         type: 'ed25519',
         includedAt: new BN(0),
       },
     ])
-    expect(fullDidDetails?.attestationKey?.id).toStrictEqual('att#1')
+    expect(fullDidDetails.assertionMethod?.[0].id).toStrictEqual('att#1')
```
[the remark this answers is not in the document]
rflechtner: On the other hand, as Timo said, you won't find empty arrays on the actual did document - that's just how most json-ld processors work. might even be against spec to have the property but an empty array: 

> The assertionMethod property is OPTIONAL. If present, the associated value MUST be a [set](https://infra.spec.whatwg.org/#ordered-set) of one or more [verification methods](https://www.w3.org/TR/did-core/#dfn-verification-method).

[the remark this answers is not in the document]
rflechtner: withdrawing this

tjwelde (commented):
Note: I didn't review the didbatcher things here.

Overall, I like what I see. I would love, if we could work a bit on the naming though. If the only way to make a lightdid, e.g. is by using the `LightDidDetails.fromDetails` function, I would expect it to rather be named something like `LightDidDetails.create`. I would even prefer `LightDid.create`, which I know, the Did is just the uri, but wouldn't it make it more concise everywhere?

Also (and that might be another PR), wouldn't it be good to have some functions, which make it easier to generate a light, or full did from a keypair, without needing to make objects according to the `LightDidCreationDetails` interface?

packages/types/src/DidDetails.ts
arty-name: This depends on whether we want to go all in on the idea "literally work with the DID document" or just move in this direction. So far I believe that some practical differences are warranted: 

1) use key type `sr25519` instead of `Sr25519VerificationKey2020`
2) when the same key was used for multiple relationships, duplicate it instead of referencing an entry in `verificationMethods` by ID
3) consider the fragment part as "internal ID" and use it

The transformation to a proper interchangeable DID Document could be done by the DocumentExporter

packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -122,26 +122,24 @@ describe('write and didDeleteTx', () => {
 
     await submitExtrinsic(tx, paymentAccount)
 
-    details = (await FullDidDetails.fromChainInfo(
+    details = (await fromChainInfo(
```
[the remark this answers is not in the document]
arty-name: I agree that the function name by itself is now unclear. Should I import the namespace instead?

packages/types/src/DidDetails.ts
rflechtner: yeah these are good points, switching to referencing key types as `Sr25519VerificationKey2020` sounds like asking for trouble. And #2 is just a consequence of the simplified data model of the did document that we adopt in this PR. I can't really say what the most practical position is on the question just how much we want to go 'all-in' on the "literally work with the DID document" goal.

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
arty-name: I decided not to go full-on boyscout on the code and limited myself to relevant changes. This code could indeed be improved, but I dislike eslint overrides almost as must as type assertions, so I would have rather added the explicit throw if details are falsy. Even better would be having a function that returns a value or throws, but I’m not touching this topic yet.

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
rflechtner: there are 2 solutions I can imagine: using a namespace import like you suggested or adding a 'did' to the names of all these constructor functions. Which one is best probably depends on how we think sdk consumers would end up importing these functions: individually or via the did namespace

packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -215,7 +215,8 @@ describe('write and didDeleteTx', () => {
     )
     const call = await DidChain.getDeleteDidExtrinsic(storedEndpointsCount)
 
-    const submittable = await (details as FullDidDetails).authorizeExtrinsic(
+    const submittable = await authorizeExtrinsic(
```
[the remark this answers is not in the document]
arty-name: Okay, I guess using a namespace will make things more clear

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
arty-name: Good catch! This change was done while removing the `migrate()`, and I just inlined whatever it was doing. It was not using the `generateCreateTxFromDidDetails` however, because it wanted to control whether the encryption key and service endpoints should be migrated. We’re not really using this control, so `generateCreateTxFromDidDetails` is really a better candidate here!

packages/did/src/DidDetails/DidDetails.ts
```diff
@@ -7,248 +7,100 @@
...
+ * Gets all public keys associated with this Did.
+ *
+ * @param did The DID data.
+ * @returns Array of public keys.
+ */
+export function getKeys(
+  did: Partial<DidDetails> & Pick<DidDetails, 'authentication'>
+): DidKey[] {
+  return [
+    ...did.authentication,
```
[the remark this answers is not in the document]
arty-name: Probably not… However if it treated the authentication as optional, some developer would look at that code and think "why? it’s not optional!"

packages/did/src/DidDetails/DidDetails.ts
```diff
@@ -7,248 +7,100 @@
...
-      publicKey: key.publicKey,
-      alg,
-      data: Crypto.coToUInt8(payload),
-    })
-    return {
-      keyUri: this.assembleKeyUri(key.id),
-      signature: u8aToHex(signature),
-    }
+  }
+  const alg = getSigningAlgorithmForVerificationKeyType(
```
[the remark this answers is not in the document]
arty-name: I am uncertain how much we should doubt TS in the code. Here it tells us that the algorithm will always be defined. Why double-check here but not in all other places?

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -36,168 +34,146 @@ import {
...
-   * @param creationDetails The creation details.
-   * @param creationDetails.identifier The DID subject identifier.
-   * @param creationDetails.uri The full DID URI.
-   * @param creationDetails.keys The set of public keys associated with the given full DID.
-   * @param creationDetails.keyRelationships The map of key ID -> relationship (e.g., authentication, attestation).
-   * @param creationDetails.serviceEndpoints The set of service endpoints controlled by the specified DID.
-   */
-  public constructor({
+/**
+ * Create a new instance of [[FullDidDetails]] after fetching the relevant information from the blockchain.
```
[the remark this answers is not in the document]
arty-name: If we proceed to remove the distinct types, I will update the file structure and the comments. When asking for a review I wasn’t confident you would approve of that change.

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -36,168 +34,146 @@ import {
...
+    capabilityDelegationKey,
+    keyAgreementKeys,
+  } = didRec
+
+  const keys: Record<string, DidKey> = publicKeys.reduce((res, key) => {
+    res[key.id] = key
+    return res
+  }, {})
+
+  const authentication = keys[authenticationKey] as DidVerificationKey
```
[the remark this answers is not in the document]
arty-name: The actual keys are coming from `publicKeys` property having the type `DidKey[]`. So yes, here we have to invoke our knowledge about keys being of proper types.

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -36,168 +34,146 @@ import {
...
+ * @param extrinsic The unsigned extrinsic to perform the lookup.
+ *
+ * @returns All the keys under the full DID that could be used to generate valid signatures to submit the provided extrinsic.
+ */
+export function getKeysForExtrinsic(
+  did: DidDetails,
+  extrinsic: Extrinsic
+): DidVerificationKey[] {
+  const keyRelationship = getKeyRelationshipForExtrinsic(extrinsic)
+  return keyRelationship === 'paymentAccount' ? [] : did[keyRelationship] || []
```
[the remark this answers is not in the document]
arty-name: While this is true, I am surprised we can access a non-existing property of an interface. Even without a complaint from TS this feels a bit dirty.

packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -8,245 +8,163 @@
...
+ * Private keys are assumed to already live in another storage, as it contains reference only to public keys.
+ *
+ * @param details The DID creation details.
+ * @param details.authenticationKey The light DID authentication key.
+ * @param details.encryptionKey The optional light DID encryption key.
+ * @param details.serviceEndpoints The optional light DID service endpoints.
+ *
+ * @returns The resulting [[LightDidDetails]].
+ */
+export function fromDetails({
```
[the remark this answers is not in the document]
arty-name: I agree, I haven’t addressed the creation naming yet, or naming in general.

packages/did/src/DidDetails/LightDidDetails.ts
arty-name: This one is only used in a few tests and I wanted to remove it completely

packages/testing/src/TestUtils.ts
arty-name: You’re on spot, I also wanted to make the identifier prop optional

tests/bundle-test.ts
```diff
@@ -142,17 +143,39 @@ async function createFullDidFromKeypair(
 
   const sign = makeSignCallback(keypair)
 
-  const fullDid = await lightDid.migrate(
+  const creationTx = await Did.Chain.generateCreateTxFromCreationDetails(
```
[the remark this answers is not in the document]
arty-name: Thanks to Raphael’s comment above I have realized that this place also should have skipped `generateCreateTxFromCreationDetails` and used `generateCreateTxFromDidDetails` instead. The latter I would like to rename to `getStoreTx`.

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
arty-name: I think at the moment most of the examples and tests in SDK use namespaces, so I will stick to this approach for now. I am not a fan of namespaces, however. I think they impair tree-shaking among other things. Such change is a much harder sell, though, so I am not attempting it yet :)

tests/bundle-test.ts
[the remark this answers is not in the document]
arty-name: I’ve pushed an improvement for that

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
arty-name: I’ve pushed an improvement for that

packages/did/src/didAuthorizeExtrinsics/didAuthorizeExtrinsics.ts
arty-name: I’ve pushed a fix for that

packages/did/src/DidDetails/FullDidDetails.ts
arty-name: I’ve pushed a fix for that in d3f19b18

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
arty-name: I’ve pushed a fix for that in d5146ca4

packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -8,245 +8,163 @@
...
-   *
-   * @param submitterAddress The KILT address to bind the DID creation operation to. It is the same address that will have to submit the operation and pay for the deposit.
-   * @param sign The callback to sign the operation.
-   * @param migrationCallback A user-provided callback to handle the packed and ready-to-be-signed extrinsic representing the DID creation operation.
-   * @param upgradeOptions Optional.
-   * @param upgradeOptions.withEncryptionKey When set to true (default) the LightDID's encryption key is added to the on-chain DID.
-   * @param upgradeOptions.withServiceEndpoints When set to true the LightDID's ServiceEndpoints are added to the on-chain DID. This is strictly opt-in as there are more restrictive size limits for on-chain service records.
-   * @returns The migrated [[FullDidDetails]] if the user-provided callback successfully writes the full DID on the chain. It throws an error otherwise.
-   */
-  public async migrate(
```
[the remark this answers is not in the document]
arty-name: Exactly, it was really forwarding a call to another function and then querying the result

packages/testing/src/TestUtils.ts
arty-name: Done in 7a9453ca

tests/bundle-test.ts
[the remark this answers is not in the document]
arty-name: Renamed the function in 28659889

packages/did/src/DidDetails/LightDidDetails.ts
arty-name: Removed in a4b3901d

packages/did/src/DidDetails/LightDidDetails.ts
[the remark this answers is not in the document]
rflechtner: I’ll leave this here then

packages/core/src/__integrationtests__/Did.spec.ts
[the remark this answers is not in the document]
rflechtner: nice one!

packages/did/src/DidDetails/FullDidDetails.ts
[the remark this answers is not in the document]
rflechtner: Makes sense. What do you think of 'dummy types' along the lines of what I suggested in my main comment? This could be something like
```ts
interface LightDidDetails extends DidDetails {
  uri: `did:kilt:light:${string}`
}
```
and would primarily function to communicate which function works with what. In that case we could of course keep those functions in separate files.

packages/did/src/DidDetails/LightDidDetails.ts
rflechtner: I think if we go this route we need to work on the constructors. We have a function in FullDidDetails that does the exact same thing for on-chain DIDs but has a different name - we really need to think about whether we want to integrate those two, and if not work on the naming.

packages/did/src/DidDetails/FullDidDetails.ts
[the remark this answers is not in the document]
arty-name: I think at the moment we do not have any function which only accepts light DIDs. We only have two functions which reject light DIDs. I feel that introducing top-level always visible classes just to communicate limitations of two functions is not only an overkill, but also creates more confusion than it clears up.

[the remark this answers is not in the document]
rflechtner: could be overkill, yeah, especially if function names communicate constraints well (we talked about `parseDetailsFromLightDid` as a name for the one function I can think of that does only accept a light did).

packages/did/src/DidDetails/LightDidDetails.ts
rflechtner: talked about it 1 on 1, this is likely going to be `parseDetailsFromLightDid`, then we'll have a `queryDid` or `queryFullDid`. We could think about a higher level function (`resolve`?) that accepts either and chooses the correct path (parse or query) depending on the did type

arty-name: As discussed, we’ll have `createDetails`, `parseDetailsFromLightDid`, and `query`. 00d7c08f and 8ce24621

packages/core/src/credential/Credential.spec.ts
```diff
@@ -532,11 +527,10 @@ describe('Credential', () => {
     })
   })
   it('verify credentials signed by a light DID', async () => {
-    const { keypair, sign } = makeSigningKeyTool('ed25519')
-    identityDave = await LightDidDetails.fromIdentifier(
-      encodeAddress(keypair.publicKey, ss58Format),
-      'ed25519'
-    )
+    const { sign, authentication } = makeSigningKeyTool('ed25519')
+    identityDave = await Did.createDetails({
```
tjwelde, line 531: If this function creates lightdid details, I think the namespace or function name should reflect that.
e.g. `Did.createLight`

packages/core/src/delegation/DelegationNode.ts
```diff
@@ -273,25 +270,25 @@ export class DelegationNode implements IDelegationNode {
...
+      keySelection = Did.Utils.defaultKeySelectionCallback,
     }: {
       keySelection?: DidKeySelectionCallback<DidVerificationKey>
     } = {}
-  ): Promise<DidUtils.EncodedSignature> {
-    const authenticationKey = await keySelection(
-      delegateDid.getVerificationKeys('authentication')
-    )
+  ): Promise<Did.Utils.EncodedSignature> {
+    // TODO: do we need keySelection for the single authentication key here and elsewhere?
```
tjwelde, line 278: I fear we have to, because on-chain it is stored as an array. Not sure, if we currently allow having multiple authentication keys though. In that case we might be able to remove the keyselection for the time being.

packages/types/src/DidDetails.ts
```diff
@@ -139,75 +145,46 @@ export type DidServiceEndpoint = {
...
-   * @param relationship A [[KeyRelationship]] or 'none' to filter out keys with a specific key
-   * relationship, undefined to return all keys.
-   * @returns An array of all or selected [[DidEncryptionKey]]s, depending on the `relationship` parameter.
-   */
-  getEncryptionKeys(relationship: EncryptionKeyRelationship): DidEncryptionKey[]
+  service?: DidServiceEndpoint[]
+}
 
-  getKeys(): DidKey[]
+export interface DidCreationDetails {
```
tjwelde, line 179: Since this is only used to create a Full did, I think the name should reflect that.

packages/types/src/DidResolver.ts
```diff
@@ -55,7 +55,7 @@ export interface IDidResolver {
    * @returns A promise of a [[DidResolvedDetails]] object if the didUri contains no fragment, [[ResolvedDidKey]] or [[ResolvedDidServiceEndpoint]] otherwise. Null if a resource cannot be resolved.
    */
   resolve: (
-    didUri: DidUri | DidResourceUri | DidPublicServiceEndpoint['uri']
+    didUri: DidUri | DidResourceUri | DidPublicServiceEndpoint['id']
```
tjwelde, line 58: for did and did resources we enter a uri, but for service endpoints just the id? Seems inconsistent

packages/types/src/DidDetails.ts
```diff
@@ -139,75 +145,46 @@ export type DidServiceEndpoint = {
...
+}
 
-  /**
-   * Retrieves a particular public key record via its id.
-   *
-   * @param id The key ID, without the leading DID URI.
-   * @returns [[DidKey]] or undefined if no key with this id is present.
-   */
-  getKey(id: DidKey['id']): DidKey | undefined
+export interface DidDetails {
```
tjwelde, line 167: This seems to overlap with `DidCreationDetails` quite a bit and in the bundle tests the light did details are even used for `getStoreTx` to generate the transaction for a full did.
Can / Should we try to bring them together?
Or is the distinction important, e.g. for creating a full did without lightdid details?

tests/bundle-test.ts
```diff
@@ -237,18 +232,18 @@ async function runAll() {
...
     devFaucet.address,
     sign
   )
   await Blockchain.signAndSubmitTx(createTx, devFaucet, {
     resolveOn: Blockchain.IS_IN_BLOCK,
   })
-  const fullDid = await Did.FullDidDetails.fromChainInfo(
-    Did.Utils.getFullDidUriByKey(authenticationKey)
+  const fullDid = await Did.query(
+    Did.Utils.getFullDidUriByKey(authentication[0])
```
tjwelde, line 246: Is it more accurate to name this function `getFullDidUriFromKey`?

packages/did/src/Did.utils.ts
```diff
@@ -44,52 +45,78 @@ const KILT_DID_PREFIX = 'did:kilt:'
...
  *
  * @param identifier A ss58 encoded address valid on the KILT network.
- * @param didType 'full' to produce a FullDid's uri, 'light' for a LightDid.
+ * @param type 'full' to produce a FullDid's uri, 'light' for a LightDid.
  * @param version KILT DID specification version number.
  * @param encodedDetails When compiling a LightDid uri, encoded DidDetails can be appended to the end of the uri.
  * @returns A DID uri as a string.
  */
 export function getKiltDidFromIdentifier(
   identifier: DidIdentifier,
```
tjwelde, line 85: Is there a distinction between an identifier and an address here?
In the regex above it was already renamed and I think it is better to be consistent with the terminology, if possible.

packages/did/src/Did.utils.ts
```diff
@@ -104,37 +131,50 @@ export function parseDidUri(
...
+      authKeyType,
+      version: versionString,
+      encodedDetails,
+      fragment,
+    } = matches
+    const address = matches.address as KiltAddress
+    const version = versionString
+      ? parseInt(versionString, 10)
+      : LIGHT_DID_LATEST_VERSION
+    const identifier = `${authKeyType}${address}` as DidIdentifier
```
tjwelde, line 163: You might be able to type the variable here instead of using type assertion. Would give you the benefit of actual type checking.

packages/did/src/DidDetails/LightDidDetails.utils.ts
```diff
@@ -16,165 +16,158 @@ import type {
...
+    }))
   }
 
   if (Object.keys(objectToSerialize).length === 0) {
     return null
   }
 
+  const serializationFlag = 0x0
   const serialized = cborEncode(objectToSerialize)
-  // Add a flag to recognize the serialization algorithm. (Currently only custom object + cbor)
```
tjwelde, line 159: I would keep the comment about `0x0` specifying our custom object + cbor, since I think it is not documented anywhere else 😅

tjwelde (changes requested):
Good progress.
Biggest gripe for me is still the distinction between full did and light did.
Is it clear, that you get a full did, if you use `Did.query`. Is it clear, that you get a light did, when using `Did.createDetails`. Is it clear, that you can use a light did for `getStoreTx`?

packages/did/src/DidDetails/DidDetails.ts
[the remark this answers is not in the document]
tjwelde: TypeScript can maybe be trusted, but not necessarily JavaScript. Since the key is gotten via an internal function and checked for existence, it is fine in this case, though.

packages/core/src/credential/Credential.spec.ts
tjwelde: The more I think about it, the more I would prefer namespacing. `LightDid.createDetails` in this case. Or with the full did: `FullDid.Chain.getStoreTx`. What do you think?

packages/did/src/DidDetails/FullDidDetails.ts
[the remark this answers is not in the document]
tjwelde: I don't have a strong opinion on this topic

packages/did/src/DidDetails/DidDetails.ts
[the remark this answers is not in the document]
arty-name: > Since the key is gotten via an internal function and checked for existence, it is fine in this case, though.

This is why the question surprised me.

packages/types/src/DidResolver.ts
arty-name: The service endpoint is also a resource, so I am not sure why exactly my predecessors added the service endpoint here. I’d be happy to drop it.

tests/bundle-test.ts
arty-name: I’m used to thinking that I can get a kernel from a peach because the peach contains the kernel. Alternatively I can get a book by ISBN, even though the ISBN does not contain the book and only refers to it. Since the key does not, strictly speaking, contains the full DID URI, I used `By` in the name.

packages/did/src/Did.utils.ts
arty-name: The address is always `4...`. The identifier in the light DID could be `004...`, it’everything after `did:kilt:` prefix. The function was created before I put in place the proper distinction between the address and identifier.

I will review its usages to see what it should be.

packages/core/src/delegation/DelegationNode.ts
arty-name: > on-chain it is stored as an array

I don’t think this is true, and [William confirmed my understanding](https://botlabs-group.slack.com/archives/CKB1ZF0QY/p1660894109310849)

packages/did/src/DidDetails/LightDidDetails.utils.ts
arty-name: I am a big fan of the code that explains itself. Could a better name of this constant make it self-explanatory? I reused `serializationFlag` because this is what it is called in the `decode...` function. How about `versionOfEncoding` or `encodingVersion` or `serializationAlgId`?

packages/did/src/Did.utils.ts
arty-name: Since TS doesn’t have insight into the regular expression, it doesn’t know possible values of `authKeyType` and complains:

>TS2322: Type '`${string}4${string}`' is not assignable to type 'DidIdentifier'.

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -64,16 +63,16 @@ export async function resolveDoc(
...
-        details = LightDidDetails.fromUri(did, false)
+        details = Did.parseDetailsFromLightDid(did, false)
       } catch (cause) {
         throw new SDKErrors.InvalidDidFormatError(did, {
           cause: cause as Error,
         })
       }
 
-      const fullDidDetails = await queryDetails(details.identifier)
+      const fullDidDetails = await queryDetails(address)
```
arty-name, line 75: Looks like there’s a bug on one of the next lines here. Input is light DID, we parse details from it, find that it has been upgraded, but still return the original parsed details. Shouldn’t we return the details fetched from the chain, @tjwelde?

packages/did/src/Did.utils.ts
arty-name: So I took a long look at the usages of `getKiltDidFromIdentifier()`. Many of those we for the migration flow: I have stored a light DID and want to know its new URI. I have added a dedicated function for this purpose: `getFullDidUri()`. Besides, some code gets an address from the blockchain and needs to transform that to a full DID, and the same function helps with that. There are some tests that hard-code the address but not the DID, why not both? Finally, there was some raw string processing that could be simplified. After all these updates, I was able to remove the `getKiltDidFromIdentifier()` completely.

packages/types/src/DidDetails.ts
arty-name: While the interfaces are similar, they have important differences, and I don’t see a practical way to bring them together. The interfaces for creation define keys that do not need IDs, while the keys of the DidDetail have IDs. Moreover, the creation of light DIDs only supports two key types. 

An alternative approach would be to inline these type definitions. I don’t think this it’s a good API to give such a central position to a type of one parameter of one function.

packages/core/src/credential/Credential.spec.ts
arty-name: I think the `LightDid` part rather belongs in the name of the function itself, I don’t see many other candidates for including in the `LightDid` namespace.

packages/core/src/delegation/DelegationNode.ts
tjwelde: ah, good to know. I also remember some conversations around this now.
I think we should get rid of multiple keys for all the verification keys than. Makes it much simpler to use.
That would work according to w3c spec and would reflect the blockchain, too.

packages/types/src/DidResolver.ts
tjwelde: yes, I would drop it

tests/bundle-test.ts
tjwelde: But I can generate the uri _from_ the key. I can't get the book _from_ an ISBN.

packages/did/src/Did.utils.ts
tjwelde: ah, too bad

packages/did/src/DidDetails/LightDidDetails.utils.ts
tjwelde: yes, naming would also help very much

packages/did/src/DidResolver/DidResolver.ts
tjwelde: I think I had a discussion with @ntn-x2 about this before. I think it was about, that we want to inform the user, that the did was upgraded. This information is obtained from the metadata.
Not sure whether to return the full did details together with that metadata information though. Sounds to be the best way to go, I find.

packages/core/src/credential/Credential.spec.ts
arty-name: Done in 0d367a30

packages/types/src/DidResolver.ts
arty-name: Done in deb62199

packages/did/src/DidDetails/LightDidDetails.utils.ts
arty-name: Done in b603395c

packages/did/src/DidResolver/DidResolver.ts
arty-name: Yes, there will be a value called `canonicalId` if a light DID was requested but full DID found. For the current implementation the application developer should resolve a light DID, check for presence of canonicalId, and then resolve the canonicalId again. I don’t think Sporran or SKYC do that, the assumption there was that the resolve call will simply produce the latest data even if an upgrade has happened.

The naive fix has a hurdle though. The DID resolution spec says whatever ID is queried, it MUST return a document with the same ID. I haven’t yet decided on a solution for that.

tests/bundle-test.ts
arty-name: "Generate" and "get" are different verbs requiring different adverbs. I am totally with you if we were to call the function `generateFullDidUriFromKey`. We do not call it that, however, and for getting we’ll need `by`.

packages/did/src/DidDocumentExporter/DidDocumentExporter.ts
```diff
@@ -8,97 +8,74 @@
...
+      controller,
+      type,
+      publicKeyBase58: base58Encode(publicKey),
     }))
-  }
+    .filter(
+      // remove duplicates
+      ({ id }, index, array) =>
+        index === array.findIndex((key) => key.id === id)
+    )
```
rflechtner, line 54: fixes https://github.com/KILTprotocol/ticket/issues/2129

tests/bundle-test.ts
tjwelde: I am only used to `find..by`, not really `get..by`. It sounds so much like an entity lookup from a set of items. Like `get car by serialnumber`
Additionally, since we are not just getting the uri directly from the key, `derive` would be a better word?

arty-name: While `derive` would be more semantically precise, this will differ from all the other functions in the DID namespace. As an application developer I would have to learn that this particular function starts not with `get` but with `derive`, and this would be annoying.

packages/did/src/DidResolver/DidResolver.ts
arty-name: Looks like this is bigger than a simple fix, let’s address it separately in https://github.com/KILTprotocol/ticket/issues/2163

packages/types/src/DidDetails.ts
arty-name: I have put on a separate branch my proposal to improve that https://github.com/KILTprotocol/sdk-js/commit/dcb30ee5060f8d9545b545d9d7d7b2d779a5e139

arty-name: Added in https://github.com/KILTprotocol/sdk-js/pull/585/commits/398758776955a682e81c9cc8b46291f96aebbf35

tests/bundle-test.ts
arty-name: Renamed in https://github.com/KILTprotocol/sdk-js/pull/585/commits/d22642372acfdc53c8c5c883e9a550a653288332

tjwelde (approved):
LGTM

PR #587, opened 2022-08-01
feat: clarity in Credential.verify

tjwelde:
This PR adds clarity to the Credential.verify checks and always checks signature, if it is present.

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -396,13 +396,17 @@ export async function verify(
     if (!isSchemaValid) throw new SDKErrors.CredentialUnverifiableError()
   }
 
-  if (challenge || !allowUnsigned) {
-    const isSignatureCorrect = await verifySignature(credential, {
+  let isSignatureCorrect = false
+  if (challenge || credential.claimerSignature) {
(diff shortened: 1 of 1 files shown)


packages/core/src/credential/Credential.ts
```diff
@@ -396,13 +396,17 @@ export async function verify(
...
-    const isSignatureCorrect = await verifySignature(credential, {
+  let isSignatureCorrect = false
+  if (challenge || credential.claimerSignature) {
+    isSignatureCorrect = await verifySignature(credential, {
       challenge,
       resolver,
     })
-    if (!isSignatureCorrect) throw new SDKErrors.SignatureUnverifiableError()
+  } else if (allowUnsigned) {
+    isSignatureCorrect = true
```
rflechtner, line 406: the logic is sound. just wondering if we shouldn't throw a different error here - which would also allow dropping the `isSignatureCorrect` variable that does not necessarily make it more readable

tjwelde: [dd87752](https://github.com/KILTprotocol/sdk-js/pull/587/commits/dd877524199c8777cc81790892ece0ec62af2942)

rflechtner (approved):
LGTM!

PR #588, opened 2022-08-05
fix: lowercase the key types

arty-name:
This PR suggests improvements for #582

The premise is that the blockchain already accepts lower-case key types, so it is not a big stretch to consider the current generated title-case types broken. If we were to fix that, we would only have to do a single lower-case transformation when decoding the keys from the blockchain. The rest of the SDK would happily work with the same types that Polkadot packages use. 

If we take this route, #584 becomes much simpler https://github.com/KILTprotocol/sdk-js/commit/adccbb2ea2be717b3c8fd0d0c4cffb8c4e4157a4

## How to test:

     yarn test:integration:latest-develop

## Checklist:

### DIFF
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -24,7 +24,7 @@
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
     "build:types": "yarn generate:defs && yarn generate:meta && yarn build:fixes",
-    "build:fixes": "sed -i.backup -E 's/\\\\bEd25519\\\\b/ed25519/g; s/\\\\bSr25519\\\\b/sr25519/g; s/\\\\bX25519\\\\b/x25519/g; s/\\\\bEcdsa\\\\b/ecdsa/g' src/interfaces/augment-api-tx.ts && rm src/interfaces/augment-api-tx.ts.backup",
+    "build:fixes": "node scripts/fixTypes.mjs",
     "build:ts": "yarn build:cjs && yarn build:esm",
     "build:cjs": "tsc --declaration -p tsconfig.build.json && echo '{\"type\":\"commonjs\"}' > ./lib/cjs/package.json",
     "build:esm": "tsc --declaration -p tsconfig.esm.json && echo '{\"type\":\"module\"}' > ./lib/esm/package.json",
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -24,7 +24,7 @@
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
     "build:types": "yarn generate:defs && yarn generate:meta && yarn build:fixes",
-    "build:fixes": "sed -i.backup -E 's/\\\\bEd25519\\\\b/ed25519/g; s/\\\\bSr25519\\\\b/sr25519/g; s/\\\\bX25519\\\\b/x25519/g; s/\\\\bEcdsa\\\\b/ecdsa/g' src/interfaces/*.ts && rm src/interfaces/*.backup",
+    "build:fixes": "sed -i.backup -E 's/\\\\bEd25519\\\\b/ed25519/g; s/\\\\bSr25519\\\\b/sr25519/g; s/\\\\bX25519\\\\b/x25519/g; s/\\\\bEcdsa\\\\b/ecdsa/g' src/interfaces/augment-api-tx.ts && rm src/interfaces/augment-api-tx.ts.backup",
     "build:ts": "yarn build:cjs && yarn build:esm",
     "build:cjs": "tsc --declaration -p tsconfig.build.json && echo '{\"type\":\"commonjs\"}' > ./lib/cjs/package.json",
     "build:esm": "tsc --declaration -p tsconfig.esm.json && echo '{\"type\":\"module\"}' > ./lib/esm/package.json",
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -23,7 +23,8 @@
   "scripts": {
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
-    "build:types": "yarn generate:defs && yarn generate:meta",
+    "build:types": "yarn generate:defs && yarn generate:meta && yarn build:fixes",
+    "build:fixes": "sed -i.backup -E 's/\\\\bEd25519\\\\b/ed25519/g; s/\\\\bSr25519\\\\b/sr25519/g; s/\\\\bX25519\\\\b/x25519/g; s/\\\\bEcdsa\\\\b/ecdsa/g' src/interfaces/*.ts && rm src/interfaces/*.backup",
     "build:ts": "yarn build:cjs && yarn build:esm",
     "build:cjs": "tsc --declaration -p tsconfig.build.json && echo '{\"type\":\"commonjs\"}' > ./lib/cjs/package.json",
     "build:esm": "tsc --declaration -p tsconfig.esm.json && echo '{\"type\":\"module\"}' > ./lib/esm/package.json",
--- a/packages/augment-api/src/interfaces/augment-api-tx.ts
+++ b/packages/augment-api/src/interfaces/augment-api-tx.ts
@@ -391,7 +391,7 @@ declare module '@polkadot/api-base/types/submittable' {
        * - Writes: Delegations
        * # </weight>
        **/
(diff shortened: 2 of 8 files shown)


packages/did/src/Did.chain.ts
```diff
@@ -153,15 +153,9 @@ function decodeDidPublicKeyDetails(
   const key = keyDetails.key.isPublicEncryptionKey
     ? keyDetails.key.asPublicEncryptionKey
     : keyDetails.key.asPublicVerificationKey
-  const keyType = key.type.toLowerCase() as Lowercase<typeof key.type>
-  if (!keyType) {
```
arty-name, line 157: No way this will happen

ntn-x2 (approved):
LGTM!

packages/augment-api/src/interfaces/types-lookup.ts
```diff
@@ -1922,7 +1922,7 @@ declare module '@polkadot/types/lookup' {
     readonly asSr25519: SpCoreSr25519Signature;
     readonly isEcdsa: boolean;
     readonly asEcdsa: SpCoreEcdsaSignature;
-    readonly type: 'Ed25519' | 'Sr25519' | 'Ecdsa';
+    readonly type: 'ed25519' | 'sr25519' | 'ecdsa';
```
rflechtner, line 1925: These enum types unfortunately are dishonest now, which could be a real tricky trap for the ts user. If you interact with such an enum instance, the `type` string will still be sentence cased. This is because the types as used by the api instance are read directly from the chain's metadata and will therefore not respond to the changes made here.

packages/augment-api/src/interfaces/augment-api-tx.ts
```diff
@@ -391,7 +391,7 @@ declare module '@polkadot/api-base/types/submittable' {
        * - Writes: Delegations
        * # </weight>
        **/
-      addDelegation: AugmentedSubmittable<(delegationId: H256 | string | Uint8Array, parentId: H256 | string | Uint8Array, delegate: AccountId32 | string | Uint8Array, permissions: DelegationDelegationHierarchyPermissions | { bits?: any } | string | Uint8Array, delegateSignature: DidDidDetailsDidSignature | { Ed25519: any } | { Sr25519: any } | { Ecdsa: any } | string | Uint8Array) => SubmittableExtrinsic<ApiType>, [H256, H256, AccountId32, DelegationDelegationHierarchyPermissions, DidDidDetailsDidSignature]>;
+      addDelegation: AugmentedSubmittable<(delegationId: H256 | string | Uint8Array, parentId: H256 | string | Uint8Array, delegate: AccountId32 | string | Uint8Array, permissions: DelegationDelegationHierarchyPermissions | { bits?: any } | string | Uint8Array, delegateSignature: DidDidDetailsDidSignature | { ed25519: any } | { sr25519: any } | { ecdsa: any } | string | Uint8Array) => SubmittableExtrinsic<ApiType>, [H256, H256, AccountId32, DelegationDelegationHierarchyPermissions, DidDidDetailsDidSignature]>;
```
rflechtner, line 394: Seems like all we really need are the changes in this file, how about we restrict ourselves to a search and replace here?

packages/did/src/Did.chain.ts
rflechtner: true, I wonder if it was meant to be `key.type?.toLowerCase()` ? But even then, the enum type always has a value.

packages/did/src/Did.chain.ts
```diff
@@ -153,15 +153,9 @@ function decodeDidPublicKeyDetails(
...
-    throw new SDKErrors.DidError(
-      `Unsupported key type "${key.type}" found on chain`
-    )
-  }
   return {
     id: keyId.toHex(),
-    type: keyType,
+    type: key.type.toLowerCase() as typeof key.type,
```
rflechtner, line 158: that's dealing with the fact that enum type strings are still sentence case, right?

rflechtner (changes requested):
That's actually a really good idea! I think there is one thing we would need to fix, however, and that is the fact that these types are now a little dishonest; if you interact with one of the enum types we've overwritten here, it will still be sentence cased.
I see two possible routes to fix this:
1) We could try to find a way to load the `lookup.ts` types when creating the api; as far as I understand this should load the `type` strings from these definitions when creating an instance of the relevant enums, which would now be lower-cased. But even if this works, it does not seem like a great way to prepare our 'bring your own api' approach that we were planning.
2) We restrict the search-and-replace to `[augment-api-tx.ts](https://github.com/KILTprotocol/sdk-js/pull/588/files#diff-cb240385152f52b2e9307fd8cdf627bd6df88e162679f7d4a7e718c0e7b4aa6c)`, effectively only changing what sort of input the extrinsic generation accepts, leaving the types of the enum codec untouched. Am I right in assuming this is the relevant part that we need changed?

packages/augment-api/src/interfaces/types-lookup.ts
arty-name: Yep, let’s undo that

packages/augment-api/src/interfaces/augment-api-tx.ts
arty-name: I’m all for surgical fixes! I don’t really know what I am doing with these types, so I went for the mass change initially.

packages/did/src/Did.chain.ts
arty-name: Yeah, and might be addressed when we apply a narrower update

arty-name:
Thank you for the great suggestion @rflechtner! Hope you like the updated implementation better.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -62,7 +62,7 @@ export type EthereumAddress = HexString
 export type Address = KiltAddress | SubstrateAddress | EthereumAddress
 
 /// Type of signatures to link accounts to DIDs.
-export type SignatureType = MultiSignature['type'] | 'Ethereum'
+type SignatureType = Lowercase<MultiSignature['type']> | 'ethereum'
```
rflechtner, line 65: we could explore if can't just use `KeyPair['type']` here

packages/augment-api/package.json
```diff
@@ -23,7 +23,8 @@
   "scripts": {
     "clean": "rimraf ./lib",
     "build": "yarn clean && yarn build:ts",
-    "build:types": "yarn generate:defs && yarn generate:meta",
+    "build:types": "yarn generate:defs && yarn generate:meta && yarn build:fixes",
+    "build:fixes": "sed -i.backup -E 's/\\\\bEd25519\\\\b/ed25519/g; s/\\\\bSr25519\\\\b/sr25519/g; s/\\\\bX25519\\\\b/x25519/g; s/\\\\bEcdsa\\\\b/ecdsa/g' src/interfaces/augment-api-tx.ts && rm src/interfaces/augment-api-tx.ts.backup",
```
rflechtner, line 27: the file we are changing is auto-generated in the previous step, do we even need a backup?

rflechtner (approved):
Love it! So coooool

packages/augment-api/package.json
arty-name: It’s unavoidable with MacOS default `sed`. I actually need someone on Mac to confirm this line works :)

packages/did/src/DidLinks/AccountLinks.chain.ts
arty-name: Good idea!

arty-name: Seems to work in https://github.com/KILTprotocol/sdk-js/pull/588/commits/99fc11618c6ba81bab8437963d2cc939d63cf639

arty-name:
What do you think @tjwelde, should we merge this into your branch?

packages/augment-api/package.json
rflechtner: seems like `\b` does not work on macOS, not even with all the backslashes :( seems like we can only use actual whitespace. `sed -I ''` does not work on linux? (switches off backups on Mac)

arty-name: Wow, MacOS, just wow. Promises extended regex support, but doesn’t support word boundary \b. Well, the sed script was ugly anyway, I’ve replaced it with a bit of javascript in https://github.com/KILTprotocol/sdk-js/pull/588/commits/99fc11618c6ba81bab8437963d2cc939d63cf639

rflechtner: Yeah... honestly I'm not even trying to understand the differences between all the different regex dialects bc it's all so inconsistent. Using JS is definitely the way to go to make it work cross-platform (it's also why we use `rimraf` for file removal).
And this means we could even do:
```js
const fixed = source.replace(/\b(Ed25519|Sr25519|X25519|Ecdsa)\b/g, (match) =>
  match.toLowerCase()
)
```

arty-name: Yes, this is even better!

tjwelde (approved):
Yes. LGTM!

PR #591, opened 2022-08-16
fix: testcontainer startup await once and for all

rflechtner:
## fixes all those broken workflows

Don't known why I didn't do that already, but if we're searching the logs for the specified ws port to determine if an image is ready, this should fix all broken waits once and for all.

## How to test:
Test integration & bundle with a bunch of different images.

## Checklist:


arty-name (approved):

tjwelde (approved):
LGTM

PR #595, opened 2022-08-24
feat: use finalization defaults

arty-name:

We already have reasonable defaults for when to resolve the submit transaction promise. I removed the boilerplate-y over-specification of those.

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/Deposit.spec.ts
+++ b/packages/core/src/__integrationtests__/Deposit.spec.ts
@@ -28,7 +28,6 @@ import {
   SubmittableExtrinsic,
 } from '@kiltprotocol/types'
 import { DecoderUtils } from '@kiltprotocol/utils'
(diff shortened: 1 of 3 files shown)


packages/core/src/__integrationtests__/Deposit.spec.ts
```diff
@@ -68,7 +67,7 @@ async function checkDeleteFullDid(
   const didResult = await Did.Chain.queryDetails(fullDid.identifier)
   const didDeposit = didResult!.deposit
 
-  await submitExtrinsic(tx, identity, Blockchain.IS_FINALIZED)
+  await submitExtrinsic(tx, identity)
```
arty-name, line 70: It’s a test, I see no reason to wait for finalization in it

tjwelde (approved):
Should be fine, if the tests run through

PR #596, opened 2022-08-24
fix: more realistic bundle tests

arty-name:

The refactoring accumulated changes to the bundle tests and stopped being an example of what application developers might write. I try to bring it back to its senses.

## Checklist:


tjwelde (approved):
Looks nice.
The ethereum key-type exclusion is needed for this PR, or has it just sneaked in?

arty-name:
Yes, when simplifying the bundle I have noticed the key type issue and wanted to address it

PR #597, opened 2022-08-24
feat: convert Message to functional style

tjwelde:
## fixes KILTProtocol/ticket#2164
This PR changes the Message class to the functional programming style.

## TODO
- [x] Move message.utils functions into base message file

## Checklist:


packages/messaging/src/Message.ts
```diff
@@ -26,312 +25,287 @@ import { DidResolver } from '@kiltprotocol/did'
...
+ * Constructs a message from a message body.
+ * This should be encrypted with [[Message.encrypt]] before sending to the receiver.
+ *
+ * @param body The body of the message.
+ * @param sender The DID of the sender.
+ * @param receiver The DID of the receiver.
+ * @returns The message created.
+ */
+export function fromBody(
+  body: MessageBody | CompressedMessageBody,
```
tjwelde, line 213: The compression for messages seems to be not thought through. You can only compress the message body (not a full message) and if you try to make a new `Message` object from a compressed message body, it is decompressed, before making the message.
Seems like a compressed message body can never really be used for message sending.

tjwelde: I would like to at least get rid of the option to give a compressed message body to this function. What d you think?

arty-name: If it is so half-baked, let’s rather drop it :+1:

tjwelde: I also added a ticket for getting rid of compression altogether

arty-name (approved):

packages/messaging/src/Message.ts
```diff
@@ -5,330 +5,525 @@
...
-        }
-        break
-      default:
-    }
+/**
+ * Checks if delegation data is well formed.
+ *
+ * @param delegationData Delegation data to check.
+ */
+export function errorCheckDelegationData(
```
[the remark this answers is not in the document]
tjwelde: Yeah, there is even more i would like to move / cleanup. Will do in another ticket though

PR #598, opened 2022-08-24
feat!: drop keySelection

arty-name:

Looks like we only ever choose from one key, let’s remove the keySelection for now to avoid confusion.

## Checklist:


tjwelde (approved):
Yes, please!

PR #599, opened 2022-08-24
feat!: replace resolver with resolve

arty-name:

Same as remove KeyStore but for DidResolver. Instead of providing a full set of functions only pass the function we need.

Also removes the universal but effectively unusable `resolve()` function and uses this name for `resolveDoc()`

## Checklist:


packages/core/src/quote/Quote.ts
```diff
@@ -104,22 +104,22 @@ export async function createAttesterSignedQuote(
  *
  * @param quote The object which to be verified.
  * @param options Optional settings.
- * @param options.resolver DidResolver used in the process of verifying the attester signature.
+ * @param options.resolve Resolve function used in the process of verifying the attester signature.
```
tjwelde, line 107: I think we should mention, that it is for resolving Dids. 
Or maybe we should rename the parameter to reflect that.
(This comment applies to all functions taking a resolve function)

arty-name:
The problem with the generic resolver is that we do not provide any means to detect what exactly it has returned

tjwelde (approved):
LGTM

PR #590, opened 2022-08-12
feat: api augmentation for RPC and state_calls

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/1953.

This PR is a WIP, and is to be used as an example of how to decorate the `api` object with additional types, RPCs, and state_calls.

## To Note

1. This PR uses a locally-built version of `@kiltprotocol/type-definitions` as it is currently defined in https://github.com/KILTprotocol/type-definitions/pull/32 at commit https://github.com/KILTprotocol/type-definitions/pull/32/commits/096400187ab7b127dce1a557b9f4caf10549b4aa. Once the PR is merged and the new 0.2.0 version released, it should be sufficient to just switch to a published version of the package, with no other implications
2. This PR has fetched metadata from a standalone instance of the runtime in PR https://github.com/KILTprotocol/kilt-node/pull/378 as of commit https://github.com/KILTprotocol/kilt-node/pull/378/commits/7fe38a08e9dc55ac03715cb647d532a3e3efa955. Not sure how we could handle these different runtimes and type definitions for development purposes tho.
3. The typegen stuff still seems to be not 100% complete, since some types are correctly used in the definitions, but are not imported and need to be manually imported. I opened an issue in the `@polkadot/api` repository and you can follow its development here: https://github.com/polkadot-js/api/issues/5151.

## Checklist

- [ ] Switch to the released version of https://www.npmjs.com/package/@kiltprotocol/type-definitions once https://github.com/KILTprotocol/type-definitions/pull/32 is merged
- [ ] Update metadata and type definitions against the latest release of Spiritnet
- [ ] Fix any outstanding issues, also by monitoring the development of https://github.com/polkadot-js/api/issues/5151

PR #601, opened 2022-08-24
fix: augmentation hacks

rflechtner:
## no ticket

For es module imports every import needs a file extension, which we generally check with an eslint rule.
We don't for autogenerated files in the augment-api package, however, because these are marked with eslint-disable.
This has resulted in import issues.

This fix runs a regex-based search-replace to fix any affected import in the directory where autogenerated files are placed.

## How to test:

All imports should now have an extension.

## Checklist:

### DIFF
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -10,11 +10,6 @@
       "types": "./lib/cjs/index.d.ts",
       "import": "./lib/esm/index.js",
       "require": "./lib/cjs/index.js"
-    },
-    "./augment": {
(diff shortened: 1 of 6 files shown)


tjwelde (approved):
LGTM

ntn-x2:
@rflechtner how does this relate to https://github.com/KILTprotocol/sdk-js/pull/590?

rflechtner:
> @rflechtner how does this relate to #590?

This simply adds the .js extension to some imports in autogenerated files so we don't have to do it manually. I think that is independent of #590

PR #602, opened 2022-08-25
feat: align from *Extrinsic to *Tx method names

arty-name:

Renamed `authorizeExtrinsic()` to `authorizeTx()` and applied the same to all the methods with `Extrinsic` in their names. Inconsistent naming is confusing, and even the word "extrinsic" is not the most tongue-friendly.

authorizeExtrinsic -> authorizeTx
submitExtrinsic -> submitTx
getKeyRelationshipForExtrinsic -> getKeyRelationshipForTx
getKeysForExtrinsic -> getKeysForTx

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/AccountLinking.spec.ts
+++ b/packages/core/src/__integrationtests__/AccountLinking.spec.ts
@@ -29,7 +29,7 @@ import {
   createEndowedTestAccount,
   fundAccount,
   initializeApi,
-  submitExtrinsic,
+  submitTx,
 } from './utils'
 import { disconnect } from '../kilt'
 
@@ -80,7 +80,7 @@ describe('When there is an on-chain DID', () => {
       ).toBe(true)
 
       const associateSenderTx = api.tx.didLookup.associateSender()
-      const signedTx = await Did.authorizeExtrinsic(
+      const signedTx = await Did.authorizeTx(
         did.uri,
         associateSenderTx,
         didKey.getSignCallback(did),
@@ -89,7 +89,7 @@ describe('When there is an on-chain DID', () => {
       const balanceBefore = (
         await api.query.system.account(paymentAccount.address)
       ).data
-      await submitExtrinsic(signedTx, paymentAccount)
+      await submitTx(signedTx, paymentAccount)
 
       // Check that the deposit has been taken from the sender's balance.
       const balanceAfter = (
@@ -123,7 +123,7 @@ describe('When there is an on-chain DID', () => {
     }, 30_000)
     it('should be possible to associate the tx sender to a new DID', async () => {
       const associateSenderTx = api.tx.didLookup.associateSender()
-      const signedTx = await Did.authorizeExtrinsic(
+      const signedTx = await Did.authorizeTx(
         newDid.uri,
         associateSenderTx,
         newDidKey.getSignCallback(newDid),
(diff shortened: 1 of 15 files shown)


tjwelde:
Ah, this reminds me of the convention, we wanted to have.
Function names ending in `Extrinsic` would return something, that has to be did authorized first. Function names ending in `Tx` return something, that can be signed for fee payment.
This ticket contains a discussion about that: https://github.com/KILTprotocol/ticket/issues/1932

tjwelde:
So though I agree, that the naming is bad and inconsistent. I think we should concentrate on better naming for the transactions, that need to be did authorized first.

arty-name:
Actually, many of these are simple wrappers that we plan to remove anyway. Only a few will remain afterwards, let’s see whether it would make sense to come up with a naming convention for them :smiling_imp:

arty-name:
I think it would be an improvement over the current state:
1) we will not have a simple way to identify what needs authorization
2) we will have consistent naming
3) we will not mislead developers with current mixed signals

arty-name:
We’re actually discussing 3 names only: authorizeExtrinsic, getKeyRelationshipForExtrinsic, getKeysForExtrinsic

rflechtner:
> We’re actually discussing 3 names only: authorizeExtrinsic, getKeyRelationshipForExtrinsic, getKeysForExtrinsic

Timo and I found out that the terminology that substrate proposes is the following:
- A `Call` (or `Method`? they both exist and seem to have similar functions) describes what should be triggered on-chain; it comprises the call index (indentifying call pallet and method) and any arguments.
- An `Extrinsic` is a general term; it's essentially a Call plus caller (origin) information. Depending on the origin, we would call that:
  - A `Transaction` if the origin is an account, which would be expected to sign it.
  - An `Inherent` if the origin is internal to the runtime, requiring no signature.

I noted that all three remaining functions produce objects that function as `Call`s to be passed to did authorisation - the are not meant to be submitted as transactions (are are not inherents either). Should the name refer to `Call` then?

tjwelde:
AFAICS, the `Extrinsic` does not even have to contain the origin information, since it can be unsigned.
The call is _included_ in the Extrinsic in the `method` property, so as long as we don't want to change what the functions expect and return, I wouldn't go with it.
The type augmentation gives users `SubmittableExtrinsics` for transactions, which are `Extrinsics` with a `send` method, since they are already connected to the blockchain.
So I think we should just keep the `Extrinsic` name, tbh

arty-name:
I was trying to relay the viewpoint of an external application developer. It goes like that. 

> The SDK has a bunch of methods with names ending in `Tx`, short for "transaction". There are also three methods which for unexplicable reasons end their name not with `Tx` but with a tongue-twister `Extrinsic`, whatever that means. I guess I’ll either try to memorize that or keep stumbling upon them here and there.

tjwelde:
But now all the functions ending in `Tx` are gone and only 3 functions ending in `Extension` remain. correct?

arty-name:
No, we still have quite some `Tx` functions: 
signAndSubmitTx
submitSignedTx
dispatchTx
Delegation.getStoreTx
Delegation.getRevokeTx
Delegation.getReclaimDepositTx
Did.getStoreTx
Did.generateDidAuthenticatedTx
submitTx

rflechtner (approved):
needs conflict resolution of course, but let's go with the approach of naming everything `tx` that handles or produces extrinsics etc.

tjwelde (approved):
LGTM

PR #603, opened 2022-08-25
fix: capitalization of DID

arty-name:

The comments were a bit lax in the proper capitalization of the term `DID`.

## Checklist:


rflechtner (approved):
I was under the impression that `FullDid` was just vocabulary we use in our DID specifications, but that does not seem to be the case. In that case, I'm happy to move to `full DID` instead.

PR #604, opened 2022-08-25
feat!: replace function calls with map lookups

arty-name:

The functions seem to bring no value, and even their names are longer

## Checklist:

### DIFF
--- a/packages/did/src/Did.utils.ts
+++ b/packages/did/src/Did.utils.ts
@@ -127,32 +127,32 @@ export function isSameSubject(didA: DidUri, didB: DidUri): boolean {
 export const signatureAlgForKeyType: Record<
   VerificationKeyType,
   SigningAlgorithms
-> = {
+> = Object.freeze({
   ed25519: 'ed25519',
   sr25519: 'sr25519',
   ecdsa: 'ecdsa-secp256k1',
-}
+})
 
-export const keyTypeForSignatureAlg = Object.entries(
-  signatureAlgForKeyType
-).reduce((obj, [key, value]) => ({ ...obj, [value]: key }), {}) as Record<
-  SigningAlgorithms,
-  VerificationKeyType
->
+export const keyTypeForSignatureAlg = Object.freeze(
+  Object.entries(signatureAlgForKeyType).reduce(
+    (obj, [key, value]) => ({ ...obj, [value]: key }),
+    {}
+  )
+) as Record<SigningAlgorithms, VerificationKeyType>
 
 export const encryptionAlgForKeyType: Record<
   EncryptionKeyType,
   EncryptionAlgorithms
-> = {
+> = Object.freeze({
   x25519: 'x25519-xsalsa20-poly1305',
-}
+})
 
 export const keyTypeForEncryptionAlg: Record<
   EncryptionAlgorithms,
   EncryptionKeyType
-> = {
+> = Object.freeze({
   'x25519-xsalsa20-poly1305': 'x25519',
-}
+})
 
 export type EncodedVerificationKey =
   | { sr25519: Uint8Array }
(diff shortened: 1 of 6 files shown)


packages/did/src/Did.utils.ts
```diff
@@ -124,72 +124,34 @@ export function isSameSubject(didA: DidUri, didB: DidUri): boolean {
   return parseDidUri(didA).address === parseDidUri(didB).address
 }
 
-const signatureAlgForKeyType: Record<VerificationKeyType, SigningAlgorithms> = {
+export const signatureAlgForKeyType: Record<
+  VerificationKeyType,
+  SigningAlgorithms
+> = {
```
rflechtner, line 130:
```suggestion
export const signatureAlgForKeyType: Record<
  VerificationKeyType,
  SigningAlgorithms
> = Object.freeze({
```

packages/did/src/Did.utils.ts
```diff
@@ -124,72 +124,34 @@ export function isSameSubject(didA: DidUri, didB: DidUri): boolean {
...
-  return keyTypeForSignatureAlg[signatureAlg]
-}
-
-const encryptionAlgForKeyType: Record<EncryptionKeyType, EncryptionAlgorithms> =
-  {
-    x25519: 'x25519-xsalsa20-poly1305',
-  }
+export const keyTypeForSignatureAlg = Object.entries(
+  signatureAlgForKeyType
+).reduce((obj, [key, value]) => ({ ...obj, [value]: key }), {}) as Record<
```
rflechtner, line 138:
```suggestion
export const keyTypeForSignatureAlg = Object.freeze(
  Object.entries(
  signatureAlgForKeyType
).reduce((obj, [key, value]) => ({ ...obj, [value]: key }), {})) as Record<
```

packages/did/src/Did.utils.ts
```diff
@@ -124,72 +124,34 @@ export function isSameSubject(didA: DidUri, didB: DidUri): boolean {
...
- * Given the identifier of a key type, returns the identifier of the encryption algorithm for which it is used.
- *
- * @param keyType Key type identifier.
- * @returns Encryption algorithm identifier.
- */
-export function getEncryptionAlgorithmForEncryptionKeyType(
-  keyType: EncryptionKeyType
-): EncryptionAlgorithms {
-  return encryptionAlgForKeyType[keyType]
+export const encryptionAlgForKeyType: Record<
```
rflechtner, line 143: same

packages/did/src/Did.utils.ts
```diff
@@ -124,72 +124,34 @@ export function isSameSubject(didA: DidUri, didB: DidUri): boolean {
...
- * Given the identifier of an encryption algorithm, returns the identifier of the key type required for it.
- *
- * @param encryptionAlg Encryption algorithm identifier.
- * @returns Key type identifier.
- */
-export function getEncryptionKeyTypeForEncryptionAlgorithm(
-  encryptionAlg: EncryptionAlgorithms
-): EncryptionKeyType {
-  return keyTypeForEncryptionAlg[encryptionAlg]
+export const keyTypeForEncryptionAlg: Record<
```
rflechtner, line 150: same

rflechtner (commented):
Exporting the maps technically opens up the possibility to mess with these mappings, which I think was one of the reasons for the function approach. But how about we just add an `Object.freeze` here?

rflechtner (approved):
LGTM!

arty-name:
I could live with `Object.freeze`, but the dynamic nature of JS still allows anyone to mess with the modules to the great extent. Whenever the SDK code accesses these maps as `Did.Utils.encryptionAlgForKeyType`, the external developer can simply do `Did.Utils.encryptionAlgForKeyType = { foo: bar }`. With these glaring holes being open by the dynamic nature of the language itself, I don’t see much use in attempts to protect something.

PR #605, opened 2022-08-26
feat!: better account linking

arty-name:

I have noted for myself a few improvements for account linking code when I was reviewing the package a couple of months ago. The default signer callback was keyring-based, I aligned it with how the sign callbacks work now in the rest of SDK. One comment was perplexing. The verification of potentially prefixed signature could be cleaner.

## Checklist:


packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -59,12 +56,11 @@ export type Address = KiltAddress | SubstrateAddress | EthereumAddress
...
- * It takes the HEX-encoded tuple (DidAddress, BlockNumber) and expects the HEX-encoded signature generated by the provided address in return.
+ * It takes the HEX-encoded tuple (DidAddress, BlockNumber) and returns the Uint8Array signature generated by the provided address.
  */
-export type LinkingSignerCallback = (
-  encodedLinkingDetails: HexString,
-  address: SubstrateAddress
-) => Promise<HexString>
+export type LinkingSignCallback = (
+  encodedLinkingDetails: HexString
+) => Promise<Uint8Array>
```
arty-name, line 63: It was specified to return a hex string, but that was instantly converted back to bytes, so I changed this. If there’s a reason we need that, I’m ready to undo it.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -406,14 +402,45 @@ export async function getLinkRemovalByDidExtrinsic(
...
- * @param keyring The keyring to retrieve the signing key.
+ * @param keypair The keypair to sign the data with.
  * @returns The signature generating callback that uses the keyring to sign the input payload using the input address.
  */
-export function defaultSignerCallback(keyring: Keyring): LinkingSignerCallback {
-  return (payload: HexString, address: Address): Promise<HexString> =>
-    Promise.resolve(
-      u8aToHex(keyring.getPair(address).sign(payload, { withType: false }))
-    )
+export function makeLinkingSignCallback(
```
arty-name, line 408: I changed the name since the function is not a callback by itself, it makes callbacks.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -406,14 +402,45 @@ export async function getLinkRemovalByDidExtrinsic(
...
+  keypair: KeyringPair
+): LinkingSignCallback {
+  return async function sign(payload: HexString): Promise<Uint8Array> {
+    return keypair.sign(payload, { withType: false })
+  }
+}
+
+function getUnprefixedSignature(
```
arty-name, line 416: Extracting this into function allows to more clearly communicate the intent and to have a neater code

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -406,14 +402,45 @@ export async function getLinkRemovalByDidExtrinsic(
...
+
+  const { crypto, isValid } = signatureVerify(message, signature, address)
+  if (isValid) {
+    return {
+      signature,
+      type: crypto as KeypairType,
+    }
+  }
+
+  throw new SDKErrors.SignatureUnverifiableError()
```
arty-name, line 443: I replaced the polkadot asserts (not used anywhere else in the SDK) with a proper SDK error

packages/core/src/__integrationtests__/AccountLinking.spec.ts
```diff
@@ -38,15 +37,12 @@ import { disconnect } from '../kilt'
 
 let paymentAccount: KiltKeyringPair
 let linkDeposit: BN
-let keyring: Keyring
-let signingCallback: AccountLinks.LinkingSignerCallback
+let sign: AccountLinks.LinkingSignCallback
```
rflechtner, line 40: maybe we could investigate if the sign callback signature that we use everywhere does not also work here

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: what does sporran return when we sign? is that a reason to make it a string?

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: that's true! Tbh I still don't know 100% if this check is actually required or if the chain would also accept the prefix. Was hard to find out and we were pressed on time so we just left it. Maybe @trusch could try to figure that out?

packages/did/src/DidLinks/AccountLinks.chain.ts
rflechtner: makes sense

rflechtner (approved):
Nice, that is much cleaner! Two things may be investigated before we merge: 
1. Do we actually need to remove the prefix?
2. Can we use the same sign interface everywhere?

packages/did/src/DidLinks/AccountLinks.chain.ts
arty-name: I think it should be the responsibility of the code talking to Sporran to convert hex to bytes before feeding this into the SDK. Do you agree?

tjwelde (commented):
I would be fine with this

PR #606, opened 2022-08-26
feat: do not assert codec type

arty-name:

I vaguely remember someone saying that asserting the codec is not really needed now that we have type generation. Also it doesn’t look like we’re consistently using it across the codebase. Finally, removing the wrapper functions becomes more straightforward without these assertions. What do you guys think?

## Checklist:


tjwelde:
But we do not have tests for the generated chain types, or do we?
These tests would fail, e.g. when we do our cron tests and the chain code changed (while the generated types stay the same)

arty-name:
Should we then add the assertions in the 80% of cases where they are missing?

rflechtner:
FYI these assertions are a little weird, because they only work on the level of type names or whatever .`toRawType()` outputs. That does not protect us from changes to the type while the name stays the same...
If we want to assert, we'd better do it in a way that is actually useful.
I developed a way to walk the entire definition tree of even complex composite types, resulting in essentially a merkle root uniquely identifying a given type definition. Could be a little overkill. But if we wanted to go that route, I probably still have the branch somewhere.

rflechtner (approved):
LGTM. Was wondering for a second if any consumer of the sdk uses this function, but probably not

PR #607, opened 2022-08-26
feat: streamline resolution of DIDs, keys, endpoints

arty-name:

There was much duplication of code, also the endpoints and keys were not getting the latest values.

## Checklist:

### DIFF
--- a/packages/did/src/DidResolver/DidResolver.ts
+++ b/packages/did/src/DidResolver/DidResolver.ts
@@ -6,7 +6,6 @@
  */
 
 import {
-  DidDocument,
   DidResolutionResult,
   DidResourceUri,
   DidUri,
@@ -16,11 +15,7 @@ import {
 import { SDKErrors } from '@kiltprotocol/utils'
 
 import * as Did from '../index.js'
-import {
-  queryDetails,
-  queryDidDeletionStatus,
-  queryServiceEndpoint,
-} from '../Did.chain.js'
+import { queryDidDeletionStatus, queryServiceEndpoint } from '../Did.chain.js'
 import { getFullDidUri, parseDidUri } from '../Did.utils.js'
 
 /**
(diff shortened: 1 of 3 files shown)


packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
+    return {
+      details: fullDid,
+      metadata: {
+        deactivated: false,
+      },
+    }
+  }
+
+  // If the full DID has been deleted, return the info in the resolution metadata.
+  const isFullDidDeleted = await queryDidDeletionStatus(did)
```
tjwelde, line 43: are we not only interested in this, if the `type` is `full`?

tjwelde: ah, there was a duplication. I see. I think it would help to extend the comment, that even if a light did was migrated and then deleted, that we return the info in the metadata.

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
-        publicKey: key.publicKey,
-        type: key.type,
-      }
-    }
-    default:
-      throw new SDKErrors.UnsupportedDidError(didUri)
+    throw new SDKErrors.InvalidDidFormatError(keyUri)
+  }
+
+  const resolved = await resolve(keyUri)
```
tjwelde, line 96: so the `resolve` function ignores the key fragment part of the key uri?
No matter if yes or no, I would still just resolve the `did` uri here, since you already got it above and represent more what you actually are  doing (resolving the did document, not just the key)

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -168,47 +132,46 @@ export async function resolveKey(
...
+  if (!resolved) {
+    return null
+  }
+
+  const {
+    details,
+    metadata: { canonicalId },
+  } = resolved
+  const controller = canonicalId || did
+  const latestDid = canonicalId ? await Did.query(canonicalId) : details
```
tjwelde, line 163: again, that seems to me just unnecessary. The upgraded full did even having the same service endpoints with the same ids and so on, is rather a niche case, which we would not have to support I think.

packages/did/src/DidResolver/DidResolver.ts
arty-name: Done in https://github.com/KILTprotocol/sdk-js/commit/a1257df987a4cf856c6e3ffe35673b75b2a2ff11

packages/did/src/DidResolver/DidResolver.ts
arty-name: Done in https://github.com/KILTprotocol/sdk-js/commit/a1257df987a4cf856c6e3ffe35673b75b2a2ff11

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
-      // Metadata will simply contain `deactivated: false`.
-      return {
-        details,
-        metadata: {
-          deactivated: false,
-        },
-      }
+  }
+
+  if (type === 'full' && !fullDid) {
```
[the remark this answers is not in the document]
arty-name: Done in https://github.com/KILTprotocol/sdk-js/commit/a1257df987a4cf856c6e3ffe35673b75b2a2ff11

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -168,47 +132,46 @@ export async function resolveKey(
...
-      }
+    return {
+      ...endpoint,
+      id: serviceUri,
     }
-    default:
-      throw new SDKErrors.UnsupportedDidError(did)
+  }
+
+  const resolved = await resolve(serviceUri)
```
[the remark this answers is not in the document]
arty-name: Done in https://github.com/KILTprotocol/sdk-js/commit/a1257df987a4cf856c6e3ffe35673b75b2a2ff11

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
+  if (!resolved) {
+    return null
+  }
+
+  const {
+    details,
+    metadata: { canonicalId },
+  } = resolved
+  const controller = canonicalId || did
+  const latestDid = canonicalId ? await Did.query(canonicalId) : details
```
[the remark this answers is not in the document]
arty-name: It’s null now

packages/did/src/DidResolver/DidResolver.ts
arty-name: I have updated the PR to match the outcomes of the discussion

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
+  const resolved = await resolve(keyUri)
+  if (!resolved) {
+    return null
+  }
+
+  const {
+    details,
+    metadata: { canonicalId },
+  } = resolved
+  const controller = canonicalId || did
```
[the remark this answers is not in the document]
arty-name: Won’t be an issue anymore

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -33,129 +28,98 @@ import { getFullDidUri, parseDidUri } from '../Did.utils.js'
...
+  if (type === 'full' && !fullDid) {
+    return null
+  }
+
+  const details = Did.parseDetailsFromLightDid(did, false)
+
+  // If a full DID with same subject is present, return the resolution metadata accordingly.
+  if (fullDid) {
+    return {
+      details,
```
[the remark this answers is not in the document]
arty-name: The new `compliantResolve()` does that, the old `resolve()` does not return the document anymore

packages/did/src/DidResolver/DidResolver.ts
```diff
@@ -35,129 +31,133 @@ export async function resolve(
...
+  return {
+    document: lightDocument,
+    metadata: {
+      deactivated: false,
+    },
+  }
+}
+
+type CompliantDidResolutionResult = Omit<DidResolutionResult, 'document'> & {
+  document?: DidDocument | { uri: DidUri }
```
rflechtner, line 83: if that is meant to be fully DID compliant, then `uri` must be `id`

arty-name: It’s not meant to return a `CompliantDidDocument`

rflechtner: phew, do you think we should have a function that's called `compliantResolve` but does not return a compliant did document?

arty-name: I’m open to naming suggestions :shrug: Returning a compliant document would require pulling in DidExporter, and I hesitate to do that.

tjwelde (approved):
LGTM
but can you please update the comment on the `document` in `DidResolutionResult` in `types/DidResolver`, to reflect, that it can be undefined, when light did was upgraded?

packages/did/src/DidResolver/DidResolver.ts
arty-name: I came up with `strictResolve` and everyone likes it

PR #612, opened 2022-09-01
feat: strict boolean checks

arty-name:
## fixes KILTProtocol/ticket#2109

These are the changes we will need to have the strict boolean checks. 

The rule has caught a couple of real error cases. Some other requirements are a bit idiosyncratic.

## Checklist:


.eslintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
+    project: './tsconfig.build.json'
```
arty-name, line 23: Grants ESLint access to the TS types information

packages/core/src/attestation/Attestation.ts
```diff
@@ -45,7 +45,7 @@ export function verifyDataStructure(input: IAttestation): void {
   }
   DataUtils.validateHash(input.claimHash, 'Claim')
 
-  if (typeof input.delegationId !== 'string' && !input.delegationId === null) {
+  if (typeof input.delegationId !== 'string' && input.delegationId !== null) {
```
arty-name, line 48: This is a real error

packages/core/src/claim/Claim.ts
```diff
@@ -289,7 +289,8 @@ export function fromCTypeAndClaimContents(
   claimContents: IClaim['contents'],
   claimOwner: DidUri
 ): IClaim {
-  if (ctypeInput.schema) {
+  // TODO: is schema really optional? Yes = fix type. No = use isICType here.
+  if ('schema' in ctypeInput) {
```
arty-name, line 293: Need a decision here

packages/core/src/credential/Credential.ts
```diff
@@ -201,13 +201,13 @@ export function verifyDataIntegrity(input: ICredential): boolean {
...
-  const verificationResult = Claim.verifyDisclosedAttributes(input.claim, {
+  const { errors, verified } = Claim.verifyDisclosedAttributes(input.claim, {
     nonces: input.claimNonceMap,
     hashes: input.claimHashes,
   })
   // TODO: how do we want to deal with multiple errors during claim verification?
-  if (!verificationResult.verified)
-    throw verificationResult.errors[0] || new SDKErrors.ClaimUnverifiableError()
+  if (!verified)
+    throw errors.length > 0 ? errors[0] : new SDKErrors.ClaimUnverifiableError()
```
arty-name, line 210: There’s no simple way to let TS know that no value exists at some index of array. Empty arrays are also difficult for TS. See [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess).

packages/core/src/credential/Credential.ts
```diff
@@ -255,11 +255,11 @@ export function verifyDataStructure(input: ICredential): void {
     throw new SDKErrors.ClaimNonceMapMalformedError()
   }
 
-  if (!input.claimHashes) {
+  if (!('claimHashes' in input)) {
     throw new SDKErrors.DataStructureError('claim hashes not provided')
   }
 
-  if (typeof input.delegationId !== 'string' && !input.delegationId === null) {
+  if (typeof input.delegationId !== 'string' && input.delegationId !== null) {
```
arty-name, line 262: Again the same error

packages/core/src/delegation/DelegationNode.spec.ts
```diff
@@ -26,9 +26,9 @@ let nodes: Record<string, DelegationNode> = {}
 
 jest.mock('./DelegationNode.chain', () => ({
   getChildren: jest.fn(async (node: DelegationNode) =>
-    node.childrenIds.map((id) => nodes[id] || null)
+    node.childrenIds.map((id) => (id in nodes ? nodes[id] : null))
```
arty-name, line 29: Here `nodes` is a Record of string to object, so TS does not expect `nodes[id]` to be undefined. Modifying the type of `nodes` to be nullable results in a ton of unchecked access errors in this test, so I opted for the smaller set of fixes.

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -130,7 +130,7 @@ export async function authorizeExtrinsic(
   }
 
   const [signingKey] = getKeysForExtrinsic(did, extrinsic)
-  if (!signingKey) {
+  if (signingKey === undefined) {
```
arty-name, line 133: Again no notion of empty arrays. I think we should rather avoid them.

packages/vc-export/src/vc-js/suites/KiltAbstractSuite.ts
```diff
@@ -83,7 +83,7 @@ export abstract class KiltAbstractSuite extends suites.LinkedDataProof {
   }): Promise<boolean> {
     const { proof } = options
     const compact = await this.compactProof(proof, options)
-    const type = compact['@type'] || compact.type
+    const type = compact['@type'] ?? compact.type
```
arty-name, line 86: In all cases of amorphous JSON we get a type which can be a numeric value of `0`, and should use nullish coalescence

packages/vc-export/src/verificationUtils.ts
```diff
@@ -86,18 +86,19 @@ export async function verifySelfSignedProof(
...
     }
-    const dereferenced = documentLoader
-      ? await documentLoader(verificationMethod)
-      : undefined
-    if (!dereferenced?.document) {
+    if (typeof documentLoader !== 'function') {
+      throw new Error('did you select an appropriate document loader?')
+    }
+    const dereferenced = await documentLoader(verificationMethod)
+    if (!('document' in dereferenced)) {
```
arty-name, line 93: This whole block is weird

packages/chain-helpers/src/blockchain/SubscriptionPromise.ts
```diff
@@ -39,13 +39,14 @@ export function makeSubscriptionPromise<SubscriptionType>(
   const subscription: (value: SubscriptionType) => void =
     typeof rejectOn === 'function'
       ? (value) => {
-          if (rejectOn(value)) reject(value)
-          if (resolveOn(value)) resolve(value)
+          // eslint-disable-next-line no-extra-boolean-cast
```
arty-name, line 42: Should we disable this rule globally?

.eslintrc.js
tjwelde: otherwise it doesn't work?

arty-name: Yeah, the rule doesn’t even attempt to work without it, it’s a hard requirement.

packages/chain-helpers/src/blockchain/SubscriptionPromise.ts
arty-name: Found out that the reason is our poor type

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -139,11 +139,9 @@ export async function submitSignedTx(
 
   function handleDisconnect(): void {
     const result = new SubmittableResult({
-      events: latestResult.events || [],
+      events: latestResult.events,
```
tjwelde, line 142: That change looks not correct to me. `latestResult` can be declared, but not instantiated, before reaching `handleDisconnect`. Same with the status line below.

packages/core/src/claim/Claim.ts
tjwelde: The type definition of `ICType` does not define schema to be optional and that is good and correct.
Thinking about JS people though, it is a good idea to use our typeguards to protect against mistakes though.

packages/core/src/credential/Credential.ts
```diff
@@ -226,19 +226,19 @@ export function verifyDataIntegrity(input: ICredential): boolean {
...
   if (!input.claim.owner) {
     throw new SDKErrors.OwnerMissingError()
   }
-  if (!input.legitimations && !Array.isArray(input.legitimations)) {
+  if (!Array.isArray(input.legitimations)) {
     throw new SDKErrors.LegitimationsMissingError()
   }
 
-  if (!input.claimNonceMap) {
+  if (!('claimNonceMap' in input)) {
```
tjwelde, line 241: Not sure about all these `in` checks. `in` returns true, if the property is there, but set to undefined, which is possible in some of our types.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#using_in_with_deleted_or_undefined_properties

What are your thoughts on this?

packages/vc-export/src/verificationUtils.ts
tjwelde: there are a LOT of conditionals

packages/did/src/DidDetails/FullDidDetails.ts
tjwelde: don't really understand what you mean

tjwelde: we clarified it. TS is not complaining, that `signingKey` could be undefined.

.eslintrc.js
```diff
@@ -73,6 +74,11 @@ module.exports = {
       },
     ],
     '@typescript-eslint/ban-ts-comment': 'warn',
+    '@typescript-eslint/strict-boolean-expressions': ['error', {
+      allowNumber: false,
```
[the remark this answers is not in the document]
tjwelde: it is

packages/core/src/credential/Credential.ts
tjwelde: Function expects ICredential, not e.g. Partial<ICredential>, so it is technically correct to ignore the potential undefined properties. There is a bigger question on if we should change the verify functions to rather be type guards, or at least expect an input with potential undefined properties.

.eslintrc.js
```diff
@@ -73,6 +74,11 @@ module.exports = {
...
+      allowNullableString: true,
```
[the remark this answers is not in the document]
tjwelde: maybe we tackle that later

packages/core/src/claim/Claim.ts
arty-name: Added the type guard in https://github.com/KILTprotocol/sdk-js/pull/612/commits/e2abc8dc05b0343ea70e1959aa2a21578d445271

tjwelde (approved):
LGTM

PR #613, opened 2022-09-01
feat!: remove 40+ basic chain wrappers

arty-name:
## fixes KILTProtocol/ticket#2040

This is what the usage of SDK could look like if we remove the basic wrapper functions.

## ToDo
- [ ] Needs a documentation PR with mapping from sdk call to api call

## Checklist:

- [x] I have verified that the code works
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)

PR #615, opened 2022-09-05
feat: update the API types

arty-name:

The PR #613 would benefit from dropping the support of old delegations format, and this needs the latest types of the blockchain.

## Checklist:


arty-name:
Cannot work without other changes in #617

PR #616, opened 2022-09-05
feat!: remove all compression & decompression

tjwelde:
## fixes KILTProtocol/ticket#2165
We recommend using msgpack + gzip instead

## How to test:
- Tests run through

## Checklist:


arty-name (approved):
Could we drop jsonabc now?

tjwelde:
Will check

rflechtner (approved):
LGTM

rflechtner:
> Could we drop jsonabc now?

Don't think so, we need it to make sure Types serialise/hash consistently

tjwelde:
It is still used indirectly in Quote and CType. To investigate, if those can be removed is a separate PR, I would say.

PR #617, opened 2022-09-05
feat: no support for the old delegations format

arty-name:

The PR https://github.com/KILTprotocol/sdk-js/pull/613 would benefit from dropping the support of old delegations format.

Supersedes #615 

## Checklist:


packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -67,42 +52,26 @@ export interface AuthorizationId extends Enum {
...
   readonly attester: AccountId
-  readonly delegationId: Option<DelegationNodeId>
   readonly revoked: bool
   readonly deposit: Deposit
-}
-
-interface AttestationDetailsV2
-  extends Omit<AttestationDetailsV1, 'delegationId'> {
   readonly authorizationId: Option<AuthorizationId>
 }
```
rflechtner, line 61: Please use `AttestationAttestationsAttestationDetails`

packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -39,26 +37,13 @@ export async function getStoreTx(
...
+  return api.tx.attestation.add(
+    claimHash,
+    cTypeHash,
+    delegationId
+      ? ({ delegation: { subjectNodeId: delegationId } } as any) // maxChecks parameter is unused on the chain side and therefore omitted
+      : undefined
+  )
 }
 
 export interface AuthorizationId extends Enum {
```
rflechtner, line 49: this should also be replaced with the autogenerated type

packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -194,25 +150,12 @@ export async function getRemoveTx(
...
-    default:
-      throw new SDKErrors.CodecMismatchError(
-        'Failed to encode call: unknown authorization type'
-      )
-  }
+  return api.tx.attestation.remove(
+    claimHash,
+    maxParentChecks
+      ? ({ delegation: { maxChecks: maxParentChecks } } as any) // subjectNodeId parameter is unused on the chain side and therefore omitted
+      : undefined
```
rflechtner, line 157: `as any` should not be necessary any more

packages/core/src/attestation/Attestation.chain.ts
```diff
@@ -159,25 +128,12 @@ export async function getRevokeTx(
...
-      )
-    default:
-      throw new SDKErrors.CodecMismatchError(
-        'Failed to encode call: unknown authorization type'
-      )
-  }
+  return api.tx.attestation.revoke(
+    claimHash,
+    maxParentChecks
+      ? ({ delegation: { maxChecks: maxParentChecks } } as any) // subjectNodeId parameter is unused on the chain side and therefore omitted
```
rflechtner, line 134: `as any` should not be necessary any more - this should match the autogenerated types

rflechtner (changes requested):
there are a few things left that were introduced bc the autogenerated types reflected the old chain storage version

rflechtner (approved):
LGTM!

PR #618, opened 2022-09-05
feat!: presentations

rflechtner:
## fixes KILTProtocol/ticket#2193

This removes the optional `claimerSignature` from the `ICredential` interface and instead makes it mandatory on an interface that extends `ICredential`, which we call `ICredentialPresentation`.
In short, this change helps communicating in which cases a credential is expected to be signed, and when the signature should be added.
Another change helps to make expected flows and best practices clearer: Adding a signature is done via `createPresentation`, which is expected to be called anew for every instance of presenting a credential. 

## How to test:

Tests are included.

## Checklist:


packages/types/src/Credential.ts
```diff
@@ -21,8 +21,14 @@ export interface ICredential {
...
   delegationId: IDelegationNode['id'] | null
   legitimations: ICredential[]
   rootHash: Hash
 }
+
+export interface ICredentialPresentation extends ICredential {
+  claimerSignature: DidSignature & { challenge?: string }
+}
+
+export type CredentialOrPresentation = ICredential &
```
tjwelde, line 33: A normal type union doesn't work in this case?

rflechtner: weirdly a union of the two does not result in `{...credentialProps, claimerSignature?}`. Don't ask me why

packages/types/src/Credential.ts
```diff
@@ -21,8 +21,14 @@ export interface ICredential {
...
+  Partial<ICredentialPresentation>
```
arty-name, line 34: Would something like that be more clear?
```suggestion

type ClaimerSignature = DidSignature & { challenge?: string }

export interface ICredentialPresentation extends ICredential {
  claimerSignature: ClaimerSignature
}

export interface CredentialOrPresentation extends ICredential {
  claimerSignature?: ClaimerSignature
}
```

arty-name: Alternatively
```suggestion
interface PresentationProperties {
  claimerSignature: DidSignature & { challenge?: string }
}

export type ICredentialPresentation = ICredential & Required<PresentationProperties>

export type CredentialOrPresentation = ICredential & Partial<PresentationProperties>
```

arty-name (commented):
Looks like this preserves the data backwards compatibility, right?

I think we should keep the `Credential.verify()` as the go-to function the developers should use in the most common case. If verifying the presentations is what we expect to be the standard thing, we should rename `verifyPresentation()` to `verify()`.

packages/types/src/Credential.ts
rflechtner: `CredentialOrPresentation` is the result of the one breaking change that the split into credential and presentation could introduce: the message body types and the expectations associated with those. While it is pretty clear that for `submit-credential` we do expect the credential to be signed (i.e. we expect a presentation) it is less clear for `request-attestation`, because we didn't really define whether or not a request for attestation needs to be signed or not. So I see two possible solutions here: 
a) We make a decision here and say that requests for attestations come in the form of a presentation (or a credential). With regards to backwards compatibility this could lead to users of the new sdk having different expectations than users of the old sdk; but even then, this would not be dramatic as one side would take a property to be optional, while the other takes it as mandatory - or simply ignores it.
b) We do not make a decision here but rename `CredentialOrPresentation` to `ICredentialRequest` to make it explicit that the request can come signed or unsigned.

arty-name: I’m fine with both options.

I made the code suggestions because I found `ICredential & Partial<ICredentialPresentation>` a bit confusing, since `ICredentialPresentation` extends `ICredential`.

packages/messaging/src/Message.ts
```diff
@@ -156,9 +156,12 @@ export function verifyMessageBody(body: MessageBody): void {
       break
     }
     case 'submit-credential': {
-      body.content.forEach((credential) =>
+      body.content.forEach((credential) => {
```
ntn-x2, line 159: I would at least call the parameter for the forEach `presentation`, no?

packages/messaging/src/Message.spec.ts
```diff
@@ -1308,14 +1334,12 @@ describe('Error checking / Verification', () => {
     expect(() => Message.verifyMessageBody(rejectTermsBody)).toThrowError(
       SDKErrors.CTypeHashMissingError
     )
-    requestAttestationBody.content.credential.claimerSignature = {
+    submitCredentialBody.content[0].claimerSignature = {
```
tjwelde, line 1337: This is because `requestAttestationBody` wouldn't need a signature now, correct?

ntn-x2 (commented):
Overall I love the distinction, but I am just not sure whether it is enough or if we should add some namespacing to make it even clearer.

tjwelde (changes requested):

packages/messaging/src/Message.ts
rflechtner: right

packages/core/src/credential/Credential.ts
```diff
@@ -167,14 +143,14 @@ export async function sign(
   }: {
     challenge?: string
   } = {}
-): Promise<void> {
-  const { signature, keyUri: signatureKeyId } = await signPayload(
+): Promise<ICredentialPresentation> {
```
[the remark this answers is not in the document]
rflechtner: we have createPresentation. I was thinking of not exporting `sign` at all tbh

packages/core/src/credential/Credential.ts
```diff
@@ -432,6 +402,22 @@ export function isICredential(input: unknown): input is ICredential {
...
 }
 
+/**
+ * Type Guard to determine input being of type [[ICredentialPresentation]].
+ *
+ * @param input - An [[ICredential]], [[ICredentialPresentation]], or other object.
+ *
+ * @returns  Boolean whether input is of type ICredentialPresentation.
+ */
+export function isICredentialPresentation(
```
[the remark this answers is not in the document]
rflechtner: ok

packages/core/src/credential/Credential.ts
```diff
@@ -399,25 +360,34 @@ export async function verify(
...
+ *
+ * Upon presentation of a credential, a verifier would call this [[verify]] function.
+ *
+ * @param presentation - The object to check.
+ * @param options - Additional parameter for more verification steps.
+ * @param options.ctype - CType which the included claim should be checked against.
+ * @param options.challenge -  The expected value of the challenge. Verification will fail in case of a mismatch.
+ * @param options.didResolve - The function used to resolve the claimer's identity. Defaults to [[resolve]].
+ */
+export async function verify(
```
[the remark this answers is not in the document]
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/618#pullrequestreview-1098752970

Guess this can only be decided by a spoon fight to the death.

packages/messaging/src/Message.spec.ts
rflechtner: exactly, this property is now unknown

packages/core/src/credential/Credential.ts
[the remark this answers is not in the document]
tjwelde: could even be inlined in `createPresentation`, since it is so small

packages/core/src/credential/Credential.ts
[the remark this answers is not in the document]
tjwelde: I think it is not a strong argument, since we rename a lot of other stuff, too. The whole credential is not even the old credential anymore.
Important is, that we don't break externally facing things and serializable data.

PR #622, opened 2022-09-06
refactor!: do not auto-manage api connection 

rflechtner:
## fixes KILTProtocol/ticket#1766

This removes the `BlockchainApiConnection` module which so far has managed the `ApiPromise` instance through which the sdk interfaces with a Fullnode on the Kilt blockchain.
Users would have to create an `ApiPromise` instance themselves and pass it to the `ConfigService`, typically when calling `Kilt.init()`.

The `BlockchainApiConnection` module's most prominent function, `getConnectionOrConnect`, is thus replaced with `ConfigService.get('api')`, which would throw if `ConfigService.set({api})` had not been called beforehand.

As an additional chore, this PR also removes [`isRecoverableTxError`](https://github.com/KILTprotocol/sdk-js/commit/0355d4edd3bad4cfeaef0e483d1b8f69fd3af911) which seems to be a relic of times when we still had resigning capabilities.

## How to test:

Integrationtests are passing, this should be enough; an additional check could be to test this in one of the BTE applications. 

## Checklist:


packages/config/src/ConfigService.ts
```diff
@@ -97,6 +94,28 @@ export function set<K extends Partial<configOpts>>(opts: K): void {
...
 }
 
+/**
+ * Set the value for a configuration to its default (which may be `undefined`).
+ *
+ * @param key Key identifying the configuration option.
+ */
+export function unset<K extends keyof configOpts>(key: K): void {
+  if (Object.hasOwn(defaultConfig, key)) {
+    configuration[key] = defaultConfig
```
arty-name, line 104:
```suggestion
    configuration[key] = defaultConfig[key]
```

packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -49,7 +49,7 @@ let api: ApiPromise
 
 beforeAll(async () => {
   await initializeApi()
-  api = await BlockchainApiConnection.getConnectionOrConnect()
+  api = ConfigService.get('api')
```
arty-name, line 52: Should it be like in the other test?
```suggestion
  api = await initializeApi()
```

packages/utils/src/SDKErrors.ts
```diff
@@ -174,9 +174,21 @@ export class IdentityMismatchError extends SDKError {
   }
 }
 
-export class WsAddressNotSetError extends SDKError {
-  constructor() {
-    super('Node address to connect to not configured!')
+export class FullnodeConnectionNotSetError extends SDKError {
```
arty-name, line 177: Shall we capitalize the word?
```suggestion
export class FullNodeConnectionNotSetError extends SDKError {
```

tests/bundle-test.ts
```diff
@@ -7,6 +7,9 @@
 
 /// <reference lib="dom" />
 
+import type { KeypairType } from '@polkadot/util-crypto/types'
+import { WsProvider, ApiPromise } from '@polkadot/api'
```
arty-name, line 11: I think this is the reason for the test to fail. We access everything kilt from window.kilt. I don’t know actually how we can access polkadot stuff in this case. Should we include it in the bundle somehow?

packages/config/src/ConfigService.ts
rflechtner: damn, good catch!

packages/core/src/__integrationtests__/Did.spec.ts
rflechtner: why not. initially changed it in the other bc it was breaking for some reason

packages/utils/src/SDKErrors.ts
rflechtner: could do, I wasn't sure if full node is one word or not

tests/bundle-test.ts
rflechtner: it's one of the reasons. you also can't export apparently. I think we still need a connection builder then to make the bundle usable. or we re-export ApiPromise from the sdk package

packages/core/src/kilt/Kilt.ts
```diff
@@ -50,4 +40,31 @@ export async function init<K extends Partial<ConfigService.configOpts>>(
   config(configs || {})
   await cryptoWaitReady()
 }
-export const { disconnect } = BlockchainApiConnection
+
+/**
+ * Connects to the KILT Blockchain using the api instance set with `init()`.
+ *
+ * @param wsEndpoint
```
arty-name, line 47: I think we were calling this `address` before, should we keep the name?

tests/bundle-test.ts
```diff
@@ -19,10 +21,11 @@ import type {
   SigningAlgorithms,
   SigningData,
 } from '@kiltprotocol/types'
-import type { KeypairType } from '@polkadot/util-crypto/types'
 
 const { kilt } = window
 
+const WS_PORT = 9944
```
arty-name, line 27: If we’re not reusing the value by exporting and importing, does it even make sense to introduce this const?

packages/core/src/kilt/Kilt.ts
rflechtner: mhhhh, maybe... it was a pretty bad name though given that we also have cryptographic addresses

tests/bundle-test.ts
rflechtner: I guess not

packages/core/src/kilt/Kilt.ts
arty-name: Good point. We also have DID service endpoints, however. 

I think we can have `Uri` in the name because it is always a URI. And it kind of defines the blockchain you connect to, so `blockchainUri` would make sense to me. A `blockchainNodeUri` if you’d like. I would avoid using "full node" here though, as these words are not really clear to an external developer. (Could there also be a way to connect to a not-full node?).

Which also leads me to think that maybe we should then align the name of `FullNode…Error` with that.

rflechtner: It is also always a URL, isn't it? And more specifically, it's the URL where a Blockchain node exposes an RPC endpoint. Is JSON-RPC a concept people know of? Then we could call it `blockchainRpcUrl` or something. Or `blockchainRpcWsUri` to make it clearer that we expect a ws/wss url here

arty-name: Yeah, I guess `Url` is more fitting than `Uri` here :shrug: 

The RPC is definitely known, I like `blockchainRpcWsUrl`.

arty-name: Will `FullNodeConnectionNotSetError` then become `BlockchainRpcConnectionNotSetError`? Actually I avoid using "not" in names, so maybe even `BlockchainApiMissingError`?

packages/utils/src/SDKErrors.ts
```diff
@@ -174,9 +174,21 @@ export class IdentityMismatchError extends SDKError {
   }
 }
 
-export class WsAddressNotSetError extends SDKError {
-  constructor() {
-    super('Node address to connect to not configured!')
+export class FullNodeConnectionNotSetError extends SDKError {
+  constructor(options?: ErrorOptions) {
+    super(
+      'A function tried to access the ApiPromise instance before it was set. Did you forget to call `Kilt.init({ api })`?',
```
arty-name, line 180: I prefer messages to be to the point, what do you think?
```suggestion
      'The blockchain API is not set. Did you forget to call `Kilt.connect(…)` or `Kilt.init(…)`?',
```

packages/core/src/kilt/Kilt.ts
rflechtner: sounds good

packages/utils/src/SDKErrors.ts
```diff
@@ -174,9 +174,21 @@ export class IdentityMismatchError extends SDKError {
...
+      options
+    )
+  }
+}
+
+export class SubscriptionsNotSupportedError extends SDKError {
+  constructor(options?: ErrorOptions) {
+    super(
+      'The api instance was used in a context that relies on state or event subscriptions, but the provider does not support them',
```
arty-name, line 189: I’d like to have a shorter message and to point the developer in the right direction here.
```suggestion
      'The blockchain API should support state or event subscriptions, for example `WsProvider`.',
```

arty-name (approved):

PR #623, opened 2022-09-06
feat!: replace checkValidity with query + assertions

tjwelde:
## fixes KILTProtocol/ticket#2106
This PR removes the `Attestation.checkValidity` function.
Instead developers have to query themselves and just check for null and if it is revoked or not.

## How to test:
- Tests

## Checklist:


packages/core/src/__integrationtests__/Attestation.spec.ts
```diff
@@ -173,7 +173,9 @@ describe('When there is an attester, claimer and ctype drivers license', () => {
       tokenHolder.address
     )
     await submitExtrinsic(authorizedStoreTx, tokenHolder)
-    expect(await Attestation.checkValidity(attestation.claimHash)).toBe(true)
+    const storedAttestation = await Attestation.query(attestation.claimHash)
+    expect(storedAttestation).not.toBeNull()
```
arty-name, line 177: Do you think we need this double check? Would the next line fail when undefined !== false?

arty-name (approved):

packages/core/src/__integrationtests__/Attestation.spec.ts
rflechtner: agreed, seems like `(await Attestation.query(attestation.claimHash)?.revoked === false)` could be our recommended check to see whether an attestation is valid

rflechtner (approved):
LGTM

packages/core/src/__integrationtests__/Attestation.spec.ts
tjwelde: yeah, was thinking the same. Just added it for completeness.

PR #624, opened 2022-09-07
feat: simplify sign and encrypt callbacks

tjwelde:
## fixes KILTProtocol/ticket#2183 & fixes KILTProtocol/ticket#2184

## TODO
- [x] better naming for the callbacks
- [x] can we combine two sign callbacks again?
- [x] fix integration tests
- [x] move `SignWithoutDidCallback` inline to `getStoreTx`

## Checklist:


packages/types/src/CryptoCallbacks.ts
```diff
@@ -6,107 +6,154 @@
  */
 
 import type { SignerPayloadJSON } from '@polkadot/types/types'
+import {
```
rflechtner, line 9: are these not type imports?

packages/types/src/CryptoCallbacks.ts
```diff
@@ -6,107 +6,154 @@
...
    */
-  alg: A
+  data: Uint8Array
   /**
-   * Public key as u8a identifying the keypair to use (in combination with the alg).
+   * Info for extensions to display to user.
+   * Available when signing extrinsics.
    */
-  publicKey: Uint8Array
+  meta?: Partial<SignerPayloadJSON>
```
rflechtner, line 34: what do you think, should we rename this parameter? I think it's called meta on the polkadot-js signer interface but our callback doesn't really have anything to do with it anyway, I suppose. So a more descriptive name probably won't hurt

packages/types/src/CryptoCallbacks.ts
```diff
@@ -6,107 +6,154 @@
...
+ * A callback function to sign data.
+ */
+export type SignCallback = (
+  signData: SignRequestData
+) => Promise<SignResponseData>
+
+/**
+ * A callback function to sign data without an existing DID.
+ */
+export type SignWithoutDidCallback = (
```
rflechtner, line 74: I love the interfaces, but the names are a little backwards, no? I'd expect the callback that needs a DID for signing to be called something like `DidSignCallback` and the one that does not need a DID something like `KeyPairSignCallback` or just `SignCallback`.

We also talked about moving the without Did type to live with the did stuff if it has just a single use case (did creation). Did you decide against that?

rflechtner (commented):
I love the interfaces! Would argue for different names though and generally argue against making the `SignWithoutDidCallback` a 'thing' by defining it on CryptoCallbacks, while its use case is only ever going to be the signature of one single really specific function (did creation). Especially bc Did creation is likely to change soon and then that callback type is entirely unused. I'd treat it as a did package implementation detail instead and only define the `DidSignCallback` as a stable interface.

packages/types/src/CryptoCallbacks.ts
tjwelde: ah ok. I only know it as meta. any suggestions?

packages/types/src/CryptoCallbacks.ts
tjwelde: no, it is in the todo list of this PR

packages/types/src/CryptoCallbacks.ts
tjwelde: they are

packages/types/src/CryptoCallbacks.ts
tjwelde: we were asking ourselves, if we need it even. Will have a look.

packages/types/src/CryptoCallbacks.ts
rflechtner: If our only signCallback interface is about DIDs, I guess it's fine if we don't explicitly mention DIDs in its name

packages/did/src/DidDetails/DidDetails.ts
```diff
@@ -70,33 +65,25 @@ export function getEndpoint(
...
-      `Failed to find verification key with ID "${keyId}" on DID "${did.uri}"`
-    )
-  }
-  const alg = signatureAlgForKeyType[key.type as VerificationKeyType]
-  const { data: signature } = await sign({
-    publicKey: key.publicKey,
-    alg,
+  const { data: signature, keyUri } = await sign({
     data: Crypto.coToUInt8(payload),
+    keyRelationship: 'authentication',
```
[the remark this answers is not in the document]
tjwelde: oh, I was under the impression, that we do, because of the previous type for the did:
```
Partial<DidDocument> & Pick<DidDocument, 'authentication' | 'uri'>
```
This was asking for a did document, which has an uri and authentication key.

I can add the keyRelationship as an optional parameter, defaulting to `authentication`, though, if this functions is supposed to be used with other key relationships, too.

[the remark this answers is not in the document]
rflechtner: I think it makes sense. I can't say for sure though, because when it comes to non-extrinsic signing I just don't know which key we use for what

packages/did/src/DidDetails/DidDetails.ts
```diff
@@ -70,33 +66,27 @@ export function getEndpoint(
...
-  if (!key || !verificationKeyTypes.includes(key.type)) {
-    throw new SDKErrors.DidError(
-      `Failed to find verification key with ID "${keyId}" on DID "${did.uri}"`
-    )
-  }
-  const alg = signatureAlgForKeyType[key.type as VerificationKeyType]
-  const { data: signature } = await sign({
-    publicKey: key.publicKey,
-    alg,
+  const { data: signature, keyUri } = await sign({
```
arty-name, line 82: Could we change things in such a way, so that here we simply do `return sign({…`?

tjwelde: only, if we expect a hex string from the signing, afaics

tjwelde: will be followed up on here: https://github.com/KILTprotocol/ticket/issues/2210

PR #625, opened 2022-09-08
chore!: type guards should not throw

rflechtner:
## fixes KILTProtocol/ticket#2187

Refactors functions that were typed as type guards such that they do not throw, but always return a boolean.
This is the expected behaviour of a type guard.
Also enforces a naming convention where type guards must begin with `is...`.

Also removed 2 errors that have become or were already obsolete (unused in the sdk).

## How to test:

Tests are included.

## Checklist:

### DIFF
--- a/packages/did/src/Did.utils.ts
+++ b/packages/did/src/Did.utils.ts
@@ -165,29 +165,48 @@ export type EncodedKey = EncodedVerificationKey | EncodedEncryptionKey
 
 export type EncodedSignature = EncodedVerificationKey
 
+export function isKiltDidUri(
+  input: unknown,
+  expectType: 'ResourceUri'
+): input is DidResourceUri
+export function isKiltDidUri(input: unknown, expectType: 'Did'): input is DidUri
+export function isKiltDidUri(input: unknown): input is DidUri | DidResourceUri
+
 /**
- * Type guard assuring that a string (or other input) is a valid KILT DID uri.
+ * Type guard assuring that a string (or other input) is a valid KILT DID uri with or without a URI fragment.
  *
  * @param input Arbitrary input.
- * @param allowFragment Whether the uri is allowed to have a fragment (following '#').
- * @returns True if validation has passed.
+ * @param expectType `ResourceUri` if the URI is expected to have a fragment (following '#'), `Did` if it is expected not to have one. Default allows both.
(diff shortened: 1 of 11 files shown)


packages/did/src/Did.utils.ts
```diff
@@ -165,29 +165,48 @@ export type EncodedKey = EncodedVerificationKey | EncodedEncryptionKey
 
 export type EncodedSignature = EncodedVerificationKey
 
+export function isKiltDidUri(
+  input: unknown,
+  expectType: 'ResourceUri'
+): input is DidResourceUri
+export function isKiltDidUri(input: unknown, expectType: 'Did'): input is DidUri
+export function isKiltDidUri(input: unknown): input is DidUri | DidResourceUri
```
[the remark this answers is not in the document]
arty-name: I don’t know unfortunately

packages/did/src/Did.utils.ts
```diff
@@ -165,29 +165,48 @@ export type EncodedKey = EncodedVerificationKey | EncodedEncryptionKey
...
+ * Type guard assuring that a string (or other input) is a valid KILT DID uri with or without a URI fragment.
  *
  * @param input Arbitrary input.
- * @param allowFragment Whether the uri is allowed to have a fragment (following '#').
- * @returns True if validation has passed.
+ * @param expectType `ResourceUri` if the URI is expected to have a fragment (following '#'), `Did` if it is expected not to have one. Default allows both.
+ * @returns True if validation has passed, false otherwise.
  */
-export function validateKiltDidUri(
+export function isKiltDidUri(
```
arty-name, line 182: I think we should keep validate* as a short utility function that calls the type guard and throws when it fails.

rflechtner: ok why not

rflechtner: question though: the previous implementation was weird: you could either check for a "is DidUri" or for "is (DidUri | DidResourceUri)", even though the use case for the latter seems to be non-existent in the sdk and we would always want "is DidResourceUri" instead. Should I keep the old function signature regardless or use the signature of `isKiltDidUri`?

arty-name: I find `is (DidUri | DidResourceUri)` confusing and would not keep it unless we really need to

PR #627, opened 2022-09-08
feat: default tx subscription resolution config

rflechtner:
## fixes KILTProtocol/ticket#2185

Adding a config service option to set a default for the resolveOn parameter on `submitTx`, which will be used whenever this parameter is not set.
Also inlining the `parseSubscriptionOption` function which had little to no purpose at this point

## How to test:

Tests

## Checklist:

### DIFF
--- a/packages/chain-helpers/src/blockchain/Blockchain.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.ts
@@ -98,7 +98,9 @@ export function parseSubscriptionOptions(
   opts?: Partial<SubscriptionPromise.Options>
 ): SubscriptionPromise.Options {
   const {
-    resolveOn = IS_FINALIZED,
+    resolveOn = ConfigService.isSet('submitTxResolveOn')
+      ? ConfigService.get('submitTxResolveOn')
+      : IS_FINALIZED,
     rejectOn = (result: ISubmittableResult) =>
       EXTRINSIC_FAILED(result) || IS_ERROR(result),
     timeout,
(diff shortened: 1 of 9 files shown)


tjwelde (approved):
LGTM

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -88,53 +88,45 @@ export function EXTRINSIC_FAILED(result: ISubmittableResult): boolean {
...
-      EXTRINSIC_FAILED(result) || IS_ERROR(result),
-    timeout,
-  } = { ...opts }
-
-  return {
-    resolveOn,
-    rejectOn,
-    timeout,
-  }
+function defaultResolveOn(): SubscriptionPromise.ResultEvaluator {
```
ntn-x2, line 91: Could this be a getter? So we could access it like `Blockchain.resolveOn`? Would it make sense? You could also take it with `const { resolveOn } = Blockchain`, if needed.

tests/bundle-test.ts
```diff
@@ -36,7 +36,7 @@ const {
   BalanceUtils,
 } = kilt
 
-const resolveOn = Blockchain.IS_IN_BLOCK
+kilt.config({ submitTxResolveOn: Blockchain.IS_IN_BLOCK })
```
ntn-x2, line 39: Love it!

ntn-x2 (approved):
Just a minor point, but otherwise I love it!

PR #628, opened 2022-09-09
feat: derive KeyRelationship for batched calls

weichweich:
Batched calls are also callable by DIDs. We should also derive the KeyRelationship for those as well, the same way it's done on chain.

## Checklist:


packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +28,32 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
...
+
+  // get the VerificationKeyRelationship of a batched call
+  if (
+    section === 'utility' &&
+    (method === 'batch' || method === 'batchAll' || method === 'forceBatch') &&
+    call.args[0].toRawType() === 'Vec<Call>'
+  ) {
+    // map all calls to their VerificationKeyRelationship and deduplicate the items
+    const keys = new Set(
+      (call.args[0] as any as Array<CallBase<A>>).map(
```
weichweich, line 44: That might be a bit controversial? I guess the correct way would be to use ´registry.createType`? AFAIK for that we would need this function to be `async`.

tjwelde:
Is this about having potentially a batch call as input to `authorizeBatch`?

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +28,32 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
   web3Names: 'authentication',
 }
 
-export function getKeyRelationshipForExtrinsic(
-  extrinsic: Extrinsic
+export function getKeyRelationshipForExtrinsic<A extends AnyTuple = AnyTuple>(
+  call: CallBase<A>
```
tjwelde, line 32: Can we still use `Extrinsic` as parameter? We normally use that for the user-facing API

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +28,32 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
...
+  if (
+    section === 'utility' &&
+    (method === 'batch' || method === 'batchAll' || method === 'forceBatch') &&
+    call.args[0].toRawType() === 'Vec<Call>'
+  ) {
+    // map all calls to their VerificationKeyRelationship and deduplicate the items
+    const keys = new Set(
+      (call.args[0] as any as Array<CallBase<A>>).map(
+        (innerExtrinsic: CallBase<A>) => {
+          return getKeyRelationshipForExtrinsic(innerExtrinsic)
```
tjwelde, line 46: A bit convoluted to get one key relationship and making sure, that the rest of the array does not contain another one.
I would suggest just getting the key relationship from the first item and than using `Array.every` (or `.some`, depending on which way you want to test) to test for uniqueness.

weichweich:
I stumbled across this while exploring the `authorizeBatch` and a bug in sporran which throws an error because this function returns null. But that's only how I found this.

This is about having a batched DID call at all. You can't derive the `KeyRelationship` for a batched call. So you can't put a batch inside a DidCall using the SDK.

packages/did/src/DidDetails/FullDidDetails.utils.ts
weichweich: I changed it, because the inner call to the function doesn't have an `extrinsic` but a `CallBase`. I added another function to retain the expected interface.

packages/did/src/DidDetails/FullDidDetails.utils.ts
weichweich: Yes I guess I was a bit focused on my newly acquired `Set` skills. 😁 I would propose to use `reduce` since that's what we do, we reduce the array to a single value. I guess it's not as efficient as `every`/`some` since it's checking every value and doesn't short circuit, but I thought that might be irrelevant.

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +29,24 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
...
+  const { section, method } = call
+
+  // get the VerificationKeyRelationship of a batched call
+  if (
+    section === 'utility' &&
+    (method === 'batch' || method === 'batchAll' || method === 'forceBatch') &&
+    call.args[0].toRawType() === 'Vec<Call>'
+  ) {
+    // map all calls to their VerificationKeyRelationship and deduplicate the items
+    return (call.args[0] as any as Array<CallBase<A>>)
```
tjwelde, line 44: it is preferred, to do 
```suggestion
    return (call.args[0] as unknown as Array<CallBase<A>>)
```

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +29,24 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
   web3Names: 'authentication',
 }
 
-export function getKeyRelationshipForExtrinsic(
-  extrinsic: Extrinsic
+function getKeyRelationshipForCall<A extends AnyTuple = AnyTuple>(
```
arty-name, line 32: Since the polkadot uses name `method` for that, I’d prefer this function to be called `getKeyRelationshipForMethod`. 

I’d also prefer to drop the old one completely.

Finally, single-letter types are the new single-letter variables. Could we replace `A` with a sensible name?

packages/did/src/DidDetails/FullDidDetails.spec.ts
```diff
@@ -234,17 +234,17 @@ describe('When creating an instance from the chain', () => {
...
-            batchFunction: augmentedApi.tx.utility.batchAll,
-            extrinsics: [extrinsic, extrinsic],
-            sign,
-            submitter: keypair.address,
-          })
-        ).rejects.toMatchInlineSnapshot(
-          '[DidBuilderError: Can only batch extrinsics that require a DID signature]'
-        )
+        const batchFunction =
+          jest.fn() as unknown as typeof mockedApi.tx.utility.batchAll
```
arty-name, line 238: I thought mockedApi.tx.utility.batchAll is already a jest.fn()

packages/did/src/DidDetails/FullDidDetails.utils.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        null
+      )
+    )
+    expect(keyType).toBe('assertionMethod')
+  })
+  it('Should return correct KeyRelationship for batched call', () => {
+    const keyType = getKeyRelationshipForExtrinsic(
+      mockApi.tx.utility.batch([
+        mockApi.tx.attestation.add(
+          '0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
```
arty-name, line 32: Would it be more elegant to use `new UInt8Array(32)` here?

packages/did/src/DidDetails/FullDidDetails.utils.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+          '0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
+          null
+        ),
+      ])
+    )
+    expect(keyType).toBe('assertionMethod')
+  })
+  it('Should return correct KeyRelationship for batchAll call', () => {
+    const keyType = getKeyRelationshipForExtrinsic(
```
arty-name, line 46: This is keyRelationship, not keyType, right?

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +29,24 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
...
+  call: CallBase<A>
 ): VerificationKeyRelationship | undefined {
-  const { section, method } = extrinsic.method
+  const { section, method } = call
+
+  // get the VerificationKeyRelationship of a batched call
+  if (
+    section === 'utility' &&
+    (method === 'batch' || method === 'batchAll' || method === 'forceBatch') &&
```
arty-name, line 40: How about this?
```suggestion
    (['batch', 'batchAll', 'forceBatch'].includes(method)) &&
```

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +29,24 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
...
+    call.args[0].toRawType() === 'Vec<Call>'
+  ) {
+    // map all calls to their VerificationKeyRelationship and deduplicate the items
+    return (call.args[0] as unknown as Array<CallBase<A>>)
+      .map((innerCall: CallBase<A>) => {
+        return getKeyRelationshipForCall(innerCall)
+      })
```
arty-name, line 47: Is this more elegant?
```suggestion
      .map(getKeyRelationshipForCall)
```

packages/did/src/DidDetails/FullDidDetails.spec.ts
arty-name: Could we cast it to this type in mockedApis then?

packages/did/src/DidDetails/FullDidDetails.utils.ts
weichweich: > Since the polkadot uses name method for that, I’d prefer this function to be called getKeyRelationshipForMethod.

It expects a `CallBase` which extends the `IMethod`. So i guess both are valid. I'm not sure where the `Method` naming comes from since I never saw that naming on the chain side. There it's always a `Call`. 🤷 

> I’d also prefer to drop the old one completely.

You refer to `getKeyRelationshipForExtrinsic`? I would agree with Timo that we should keep it consistent. If all our calls expect `Extrinsics` this one should too.

packages/did/src/DidDetails/FullDidDetails.utils.spec.ts
weichweich: [5019b50](https://github.com/KILTprotocol/sdk-js/pull/628/commits/5019b50aace1a9b4f8b5de1ca13a505b2cfbc942)

packages/did/src/DidDetails/FullDidDetails.utils.spec.ts
weichweich: [5019b50](https://github.com/KILTprotocol/sdk-js/pull/628/commits/5019b50aace1a9b4f8b5de1ca13a505b2cfbc942)

packages/did/src/DidDetails/FullDidDetails.spec.ts
weichweich: `jest.fn()` needs to have the type of `mockedApi.tx.utility.batchAll` so that `jest.fn()` can be used as a batchFunction. I'm not sure if we can change anything here.

arty-name: Oh, now I see this can’t be done there

packages/did/src/DidDetails/FullDidDetails.utils.ts
arty-name: > You refer to getKeyRelationshipForExtrinsic? I would agree with Timo that we should keep it consistent. If all our calls expect Extrinsics this one should too.

I mean as an application developer I would be confused about which method to use: getKeyRelationshipForExtrinsic or getKeyRelationshipForMethod, if both are available. This is my preference though, so if you guys both think it should remain, let’s keep it.

packages/did/src/DidDetails/FullDidDetails.utils.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+        mockApi.tx.attestation.add(
+          new Uint8Array(32),
+          new Uint8Array(32),
+          null
+        ),
+      ])
+    )
+    expect(keyRelationship).toBe('assertionMethod')
+  })
+  it('Should return correct KeyRelationship for forcedBatch call', () => {
```
arty-name, line 58:
```suggestion
  it('Should return correct KeyRelationship for forceBatch call', () => {
```

packages/did/src/DidDetails/FullDidDetails.utils.ts
weichweich: > as an application developer I would be confused

You are right about that. Therefore one of them is exported while the other is not. The `getKeyRelationshipForMethod` is only there because I call it recursively for batch calls. And a batch call doesn't contain `Extrinsic`s but `CallBase`s

packages/did/src/DidDetails/FullDidDetails.utils.ts
```diff
@@ -28,10 +29,22 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
   web3Names: 'authentication',
 }
 
-export function getKeyRelationshipForExtrinsic(
-  extrinsic: Extrinsic
+function getKeyRelationshipForMethod<Arguments extends AnyTuple = AnyTuple>(
+  call: CallBase<Arguments>
 ): VerificationKeyRelationship | undefined {
```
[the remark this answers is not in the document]
arty-name: And then below we can do:
```ts
    return (call.args[0] as unknown as Array<Extrinsic['method']>)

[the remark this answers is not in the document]
weichweich: good idea! [ed537c3](https://github.com/KILTprotocol/sdk-js/pull/628/commits/ed537c35ceb432f21d0e99f594f90e044d42740f)

arty-name (approved):
Looks good to me! I wonder if @tjwelde has more comments…

packages/did/src/DidDetails/FullDidDetails.utils.ts
tjwelde: yes, better than throwing on duplicate.

tjwelde (approved):
LGTM

PR #629, opened 2022-09-12
feat: key pair helpers

rflechtner:
## fixes KILTProtocol/ticket#2192

Adds a number of helper functions to make key pair generation easier and more succinct.

## How to test:

currently untested

## Checklist:


packages/utils/src/Keypairs.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param type {Verification|Encryption} Key type.
+ * @param mnemonicOrSeed A BIP39 mnemonic or 32 byte hex string.
+ * @param derivationPath Key derivation path beginning with `//` or `/', e.g. `//hard/soft`.
+ * @returns An object containing public and secret keys.
+ */
+export function createKeypair(
+  type: VerificationKeyType | EncryptionKeyType,
+  mnemonicOrSeed: string | HexString,
+  derivationPath: string
```
tjwelde, line 62: Let's make this optional

packages/utils/src/Keypairs.ts
```diff
@@ -0,0 +1,116 @@
...
+): Keypair {
+  if (!Object.prototype.hasOwnProperty.call(PairFromSeed, type)) {
+    throw new UnsupportedKeyError(type)
+  }
+  const seed = isHex(mnemonicOrSeed)
+    ? hexToU8a(mnemonicOrSeed)
+    : mnemonicToMiniSecret(mnemonicOrSeed)
+  let pair = PairFromSeed[type](seed)
+  if (derivationPath) {
+    const { path } = keyExtractPath(derivationPath)
```
tjwelde, line 72: I think the derivation path should be applied to the seed

packages/utils/src/Keypairs.ts
```diff
@@ -0,0 +1,116 @@
...
+    pair = keyFromPath(pair, path, type === 'x25519' ? 'ed25519' : type)
+    if (type === 'x25519') {
+      pair = PairFromSeed[type](pair.secretKey)
+    }
+  }
+  return pair
+}
+
+const keyring = new Keyring({ ss58Format: 38 })
```
tjwelde, line 81: hmm, initialising such crypto heavy stuff on file load could bite us in the behind. I would just generate it on-the-fly for now

tjwelde (changes requested):
I think we should also re-export the mnemonic generation function here.

packages/utils/src/Keypairs.ts
rflechtner: oh true

packages/utils/src/Keypairs.ts
rflechtner: ok

rflechtner:
> I think we should also re-export the mnemonic generation function here.

see line 33 ;)

packages/utils/src/Keypairs.ts
tjwelde: sorry, I think I misread the code. It `is` applied to the keypair, which was generated from seed in the next line.

packages/utils/src/Keypairs.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+  return pair
+}
+
+let kr: Keyring
+const keyring = {
+  get instance() {
+    if (typeof keyring === 'undefined') {
+      kr = new Keyring({ ss58Format: 38 })
```
tjwelde, line 85: I don't think we save that much, by caching it

packages/utils/src/Keypairs.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Creates a KeyringPair instance which can be used for signing.
+ *
+ * @param pair A verification key pair as created by [[createKeypair]].
+ * @param type Key type of the pair.
+ * @returns KeyringPair instance defaulting to Kilt address encoding.
+ */
+export function createSigningKeyRingPair(
```
tjwelde, line 98:
```suggestion
export function createSigningKeyringPair(
```

arty-name:
I had [something like this in mind](https://github.com/KILTprotocol/sdk-js/commit/c5908378bda4b72a099d221dc8b918bb5141c6ca#diff-896ef2d7bcf3685314ea20d7d8ccfbe1212f1944086f644105c7597cdbc681e0)

rflechtner:
> I had [something like this in mind](https://github.com/KILTprotocol/sdk-js/commit/c5908378bda4b72a099d221dc8b918bb5141c6ca#diff-896ef2d7bcf3685314ea20d7d8ccfbe1212f1944086f644105c7597cdbc681e0)

That's also what I started with, but then I wondered whether the polkadot key ring pair really works as a base type for us bc it only handles half of the keys we use. The way sporran securely stores its keys for example is not by using the keyring's encodePcks8 - it stores the seed and re-derives a key pair from it every time it is used. Ideally, we would derive a key pair once and just save the keys, right? So that's why I came up with this strategy of having functions to create regular public/private key pairs and a function to turn them to keyring pairs when needed (e.g. for signing).

arty-name:
I see your point now.

As far as I know when Timo and I came up with this point it was to address common basic annoyances. This is also what I did in my example, and what you have started with. 

Your suggestion, while usable in an advanced application, seems to be more along the lines "forget everything you knew about the keys management, let’s do it differently". I guess it addresses and important point, but it will also need usage examples for that, and also examples of how it scratches the itches we’re having currently.

arty-name:
Superseded by #639

PR #630, opened 2022-09-14
fix: selective disclosure

rflechtner:
## rc hotfix

Selective disclosure turned out to be broken in the latest rc.
Additionally to fixing this, the docstring of `Kilt.connect()` was not particularly informative about what it actually does.

### DIFF
--- a/packages/core/src/attestation/Attestation.ts
+++ b/packages/core/src/attestation/Attestation.ts
@@ -140,9 +140,15 @@ export function verifyAgainstCredential(
   attestation: IAttestation,
   credential: ICredential
 ): boolean {
-  return (
-    credential.claim.cTypeHash === attestation.cTypeHash &&
-    credential.rootHash === attestation.claimHash &&
+  try {
+    if (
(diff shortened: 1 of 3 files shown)


tjwelde (approved):
LGTM
maybe we can change the title to reflect, that it just fixes, that functions either throw or do boolean

packages/core/src/attestation/Attestation.ts
```diff
@@ -140,9 +140,15 @@ export function verifyAgainstCredential(
   attestation: IAttestation,
   credential: ICredential
 ): boolean {
```
tjwelde, line 142: It would be more consistent though, if all the verify functions do the same, e.g. they all throw.

rflechtner: I looked into it and you're right, this is a larger issue. We still have inconsistent behavior with regards to throw vs return boolean, to much to simply fix this here. I reverted the commit and pulled it into a separate PR: https://github.com/KILTprotocol/sdk-js/pull/632

PR #632, opened 2022-09-16
refactor!: verify* throws

rflechtner:
## fixes KILTProtocol/ticket#2207

It could be nice if functions that are called `verify*` would follow a consistent pattern with regards to their error and return value behaviour.
In this PR, I explored what this could look like.

## How to test:

Tests are included

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -133,12 +133,12 @@ export function verifyRootHash(input: ICredential): boolean {
 }
 
 /**
- * Verifies the data of the [[Credential]] object; used to check that the data was not tampered with, by checking the data against hashes.
+ * Verifies the data of the [[Credential]] object; used to check that the data was not tampered with,
+ * by checking the data against hashes. Throws if invalid.
  *
  * @param input - The [[Credential]] for which to verify data.
- * @returns Whether the data is valid.
  */
-export function verifyDataIntegrity(input: ICredential): boolean {
+export function verifyDataIntegrity(input: ICredential): void {
   // check claim hash
   if (!verifyRootHash(input)) {
     throw new SDKErrors.RootHashUnverifiableError()
@@ -155,12 +155,8 @@ export function verifyDataIntegrity(input: ICredential): boolean {
 
   // check legitimations
   input.legitimations.forEach((legitimation) => {
-    if (!verifyDataIntegrity(legitimation)) {
-      throw new SDKErrors.LegitimationsUnverifiableError()
-    }
+    verifyDataIntegrity(legitimation)
   })
-
-  return true
 }
 
 /**
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -180,7 +180,7 @@ export function verifyDataStructure(input: ICredential): void {
   if (typeof input.claimNonceMap !== 'object')
     throw new SDKErrors.ClaimNonceMapMalformedError()
   Object.entries(input.claimNonceMap).forEach(([digest, nonce]) => {
-    DataUtils.validateHash(digest, 'statement digest')
+    DataUtils.verifyIsHex(digest, 32)
     if (!digest || typeof nonce !== 'string' || !nonce)
       throw new SDKErrors.ClaimNonceMapMalformedError()
   })
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -178,7 +178,7 @@ export function verifyDataStructure(input: ICredential): void {
   if (typeof input.claimNonceMap !== 'object')
     throw new SDKErrors.ClaimNonceMapMalformedError()
   Object.entries(input.claimNonceMap).forEach(([digest, nonce]) => {
-    DataUtils.verifyIsHex(digest, 32)
+    DataUtils.verifyIsHex(digest, 256)
     if (!digest || typeof nonce !== 'string' || !nonce)
       throw new SDKErrors.ClaimNonceMapMalformedError()
   })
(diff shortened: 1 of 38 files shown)


packages/core/src/credential/Credential.ts
```diff
@@ -126,41 +126,32 @@ export function makeSigningData(
...
-  // TODO: how do we want to deal with multiple errors during claim verification?
-  if (!verified)
-    throw errors.length > 0 ? errors[0] : new SDKErrors.ClaimUnverifiableError()
 
   // check legitimations
   input.legitimations.forEach((legitimation) => {
-    if (!verifyDataIntegrity(legitimation)) {
-      throw new SDKErrors.LegitimationsUnverifiableError()
-    }
+    verifyDataIntegrity(legitimation)
```
arty-name, line 153: Could it be `input.legitimations.forEach(verifyDataIntegrity)`?

packages/core/src/ctype/CType.ts
```diff
@@ -96,24 +95,28 @@ export function getIdForSchema(
...
   const validator = new JsonSchema.Validator(schema, '7', false)
   if (schema.$id !== CTypeModel.$id) {
     validator.addSchema(CTypeModel)
   }
   const result = validator.validate(object)
-  if (!result.valid && messages) {
-    result.errors.forEach((error) => {
-      messages.push(error.error)
-    })
+  if (!result.valid) {
```
arty-name, line 109: Do you think destructuring and an early return will make it more elegant?

packages/core/src/ctype/CType.ts
```diff
@@ -190,27 +178,29 @@ export function verifyDataStructure(input: ICType): void {
...
   const validator = new JsonSchema.Validator(cType, '7', false)
   nestedCTypes.forEach((ctype) => {
     validator.addSchema(ctype)
   })
   validator.addSchema(CTypeModel)
   const result = validator.validate(claimContents)
-  if (!result.valid && messages) {
-    result.errors.forEach((error) => {
-      messages.push(error.error)
+  if (!result.valid) {
```
arty-name, line 194: Same here

packages/did/src/Did.signature.ts
```diff
@@ -12,78 +12,44 @@ import {
...
-  did: DidDocument
+  didResolve?: DidResolve
 }
 
-export type VerificationResult = {
-  verified: boolean
-  reason?: string
-  did?: DidDocument
-  key?: DidVerificationKey
+// Used solely for retro-compatibility with previously-generated DID signatures.
```
arty-name, line 31: Is there the `@deprecated` tag for this purpose?

packages/utils/src/DataUtils.ts
```diff
@@ -8,66 +8,57 @@
...
+  input: unknown,
+  byteLength = 32,
+  prefixed = true
+): void {
+  if (typeof input !== 'string') {
+    throw new SDKErrors.HashTypeError()
   }
-  if (!verify(data, signature, signer)) {
-    throw new SDKErrors.SignatureUnverifiableError()
+  const hexStringPattern = new RegExp(
```
arty-name, line 57: Surprised that polkadot doesn’t have anything for that…

arty-name (approved):
Very nice! Happy to see that!

packages/core/src/credential/Credential.ts
rflechtner: oh yes it could!

packages/core/src/ctype/CType.ts
rflechtner: that would work, yes. why not.

packages/did/src/Did.signature.ts
rflechtner: I don't think that's necessary bc we don't export this type

packages/utils/src/DataUtils.ts
```diff
@@ -8,66 +8,57 @@
...
+    `^${prefixed ? '0x' : ''}[A-F0-9]{${Math.round(byteLength) * 2}}$`,
```
tjwelde, line 58: Went overkill with it again :D But I guess now you can verify any hex there is.
Fine for me

tjwelde (approved):
LGTM

packages/utils/src/DataUtils.ts
rflechtner: ohhhhhhh, damn should have thought of that :D

packages/messaging/src/Message.ts
```diff
@@ -92,10 +92,7 @@ export function verifyMessageBody(body: MessageBody): void {
         Credential.verifyDataStructure(credential)
       )
       if (body.content.delegationId) {
-        DataUtils.validateHash(
-          body.content.delegationId,
-          'Submit terms delegation id hash invalid'
-        )
+        DataUtils.verifyIsHex(body.content.delegationId, 32)
```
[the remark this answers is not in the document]
rflechtner: yeah maybe we don't need to check that here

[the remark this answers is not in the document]
rflechtner: would you not do length checks at all though? not even on the different verifyDataStructure functions?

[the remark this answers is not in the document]
tjwelde: My comment was only about removing the argument. The function itself will check the length, because we added the end character `$`

tests/bundle-test.ts
```diff
@@ -268,22 +268,8 @@ async function runAll() {
...
   await Blockchain.signAndSubmitTx(cTypeStoreTx, payer)
 
-  const stored = await CType.verifyStored(DriversLicense)
-  if (stored) {
-    console.info('CType successfully stored on chain')
-  } else {
-    throw new Error('CType not stored')
-  }
-
-  const result = await CType.verifyOwner({
```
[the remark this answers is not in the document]
tjwelde: as we discussed, it can be queried and no one cares about it.

packages/utils/src/DataUtils.ts
tjwelde: as we discussed, we will use isHex from polkadot/utils.

PR #633, opened 2022-09-16
feat: Did.uriFromChain and encode replaces type cast

arty-name:
This PR replaces the type casting of KILT accounts returned from blockchain with encoding them. Many cases don’t even need that, rather they want to get DidUri from blockchain data using `Did.uriFromChain()`.

Inspired by the [review comment](https://github.com/KILTprotocol/sdk-js/pull/613#discussion_r970700877) 

## Checklist:


rflechtner (approved):
Love it!

PR #634, opened 2022-09-16
feat: avoid .hash().isEmpty

arty-name:
This PR uses a more clear and readable call instead of another one, performance benefits of which have apparently failed to materialize.

Inspired by a [review comment](https://github.com/KILTprotocol/sdk-js/pull/613#discussion_r970741095)

## Checklist:


rflechtner (approved):
We don't have an 'is account linked' wrapper any more where this would have to be updated as well, right? In that case, LGTM!

PR #637, opened 2022-09-20
feat: add DID RPC decorations

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/1953 and supersedes https://github.com/KILTprotocol/sdk-js/pull/590.

## What it does

It uses the revamped https://github.com/KILTprotocol/type-definitions repo to decorate the `api` object with custom RPC and runtime calls and relative types.
There are few manual tweaks which I used to overcome some issues in the automated generation process.

### How to test

Somehow import the SDK in a typescript project, and then simply either create the api object with `Kilt.connect()`, or create an api object yourself and `import '@kiltprotocol/augment-api` at the top of the script.


arty-name (commented):
I am not qualified to make this call, I guess :) 

I see no issues in this PR :+1:

packages/augment-api/package.json
```diff
@@ -36,14 +36,17 @@
   "bugs": "https://github.com/KILTprotocol/sdk-js/issues",
   "homepage": "https://github.com/KILTprotocol/sdk-js#readme",
   "devDependencies": {
-    "@polkadot/api": "^9.0.0",
-    "@polkadot/typegen": "^9.0.0",
```
rflechtner, line 40: I wouldn't bump these in isolation. what's the reason for that?

packages/augment-api/src/interfaces/extraDefs/definitions.ts
```diff
@@ -1,5 +1,9 @@
 /* eslint-disable */
-// add any additional definitions (rpc etc) here
+
+import { latest, runtime } from '@kiltprotocol/type-definitions'
+
+// Only types and runtime calls can be exported from here.
```
rflechtner, line 5: what's the reason for having to split it in two files?

packages/augment-api/tsconfig.build.json
```diff
@@ -4,7 +4,18 @@
     "module": "CommonJS",
     "outDir": "./lib/cjs",
     "skipLibCheck": true,
-    "noUnusedLocals": false
+    "noUnusedLocals": false,
+    "paths": {
+      "@kiltprotocol/augment-api/*": [
+        "./src/*"
+      ],
```
rflechtner, line 11: Yeah that's the trickery I was talking about... have you tried adding `"@kiltprotocol/augment-api/interfaces/*"` as well to deal with the import issues?

rflechtner (approved):
Seems like you're also at your wits' end with how imports from the api-augment package within the api-augment package are supposed to work. I would love of course if we found a solution for that, but I guess right now we don't have much of a choice other than merging as is.

packages/augment-api/package.json
ntn-x2: We are relying on some features, for instance the `AccountId33` type, that were only released in [9.3.1](https://github.com/polkadot-js/api/releases/tag/v9.3.1).

ntn-x2: Also the runtime calls stuff has been improved in 9.2.x and 9.3.x

packages/augment-api/src/interfaces/extraDefs/definitions.ts
ntn-x2: So the RPC endpoints must be decorated in a folder that has the same name as the module you are trying to augment. So for `api.rpc.did` stuff, they have to be declared in a `did` folder. The rest, meaning types and runtime APIs, can be declared in the same folder, with no naming requirements. It's explained in [this section](https://polkadot.js.org/docs/api/examples/promise/typegen#types-setup)

packages/augment-api/tsconfig.build.json
ntn-x2: Technically `@kiltprotocol/augment-api/*` also catches any subpaths. And VScode stopped complaining with this tweak, but the build process from the project root was still failing. Maybe this is something that should rather be declared in the root?

packages/augment-api/src/interfaces/augment-api-rpc.ts
```diff
@@ -5,6 +5,9 @@
 // this is required to allow for ambient/previous definitions
 import '@polkadot/rpc-core/types/jsonrpc';
 
+// FIXME: Error when using the autogenerated "from '@kiltprotocol/augment-api/extraDefs'"
+import type { DidLinkedInfo } from './extraDefs';
```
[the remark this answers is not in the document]
ntn-x2: Not sure to be honest. But I would say that for the time being we could keep it as it is, and investigate it more when we have more time.

packages/augment-api/tsconfig.build.json
rflechtner: I imagine so, bc it could just fail then in later build steps

packages/augment-api/package.json
rflechtner: can we bump all the v9 deps then?

ntn-x2: Done in https://github.com/KILTprotocol/sdk-js/pull/637/commits/5b808272becd8f3a41680fd995cb85d6aeaa5f54.

packages/augment-api/tsconfig.build.json
ntn-x2: Tried it, and it did not work. Maybe we can get back to it once you are also back.

PR #638, opened 2022-09-21
feat!: export ConfigService from the SDK

arty-name:

Make the ConfigService directly available when importing from the sdk-js.
Removes the redundant `Kilt.config()` function.

## Checklist:

### DIFF
--- a/packages/core/src/kilt/Kilt.ts
+++ b/packages/core/src/kilt/Kilt.ts
@@ -19,7 +19,7 @@ import { ConfigService } from '@kiltprotocol/config'
 import { latest, rpc, runtime } from '@kiltprotocol/type-definitions'
 
 /**
- * Prepares crypto modules (required e.g. For identity creation) and calls ConfigService.set().
+ * Prepares crypto modules (required e.g. for identity creation) and calls ConfigService.set().
  *
  * @param configs Arguments to pass on to ConfigService.set().
  * @returns Promise that must be awaited to assure crypto is ready.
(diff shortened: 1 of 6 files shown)


tjwelde:
I think it makes sense to also export the whole config service.
Do you think we should keep the `Kilt.config()` API? https://github.com/KILTprotocol/sdk-js/blob/develop/packages/core/src/kilt/Kilt.ts#L26

arty-name:
> I think it makes sense to also export the whole config service.

What do you mean? There’s nothing else in the config package, only the ConfigService which I re-export.

> Do you think we should keep the Kilt.config() API?

Nice catch! Removed it in https://github.com/KILTprotocol/sdk-js/pull/638/commits/f9d883d1acc862c7930252a9792e7ab1d7681e68

tjwelde:
> > I think it makes sense to also export the whole config service.
> 
> What do you mean? There’s nothing else in the config package, only the ConfigService which I re-export.
Sorry, I didnt write it well. I was just agreeing that this PR makes sense.

packages/core/src/kilt/Kilt.ts
```diff
@@ -19,26 +19,15 @@ import { ConfigService } from '@kiltprotocol/config'
 import { latest, rpc, runtime } from '@kiltprotocol/type-definitions'
 
 /**
- * Allows setting global configuration such as the log level and the polkadot ApiPromise instance used throughout the sdk.
+ * Prepares crypto modules (required e.g. For identity creation) and calls ConfigService.set().
```
ntn-x2, line 22:
```suggestion
 * Prepares crypto modules (required e.g. for identity creation) and calls ConfigService.set().
```

ntn-x2 (approved):
LGTM!

PR #641, opened 2022-09-21
feat: remove Did.signPayload

tjwelde:
## fixes KILTProtocol/ticket#2210
This PR removes the `signPayload` helper function, since it does not help with much anymore.

## How to test:
- Tests
- Look, if users of the SDK would still be able to go through their use cases (especially regarding Quote and the two messages)

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -17,12 +17,7 @@
  * @packageDocumentation
  */
 
-import {
-  isDidSignature,
-  verifyDidSignature,
-  resolve,
-  signPayload,
-} from '@kiltprotocol/did'
+import { isDidSignature, verifyDidSignature, resolve } from '@kiltprotocol/did'
 import type {
   DidResolve,
   Hash,
@@ -35,6 +30,7 @@ import type {
   SignCallback,
 } from '@kiltprotocol/types'
 import { Crypto, DataUtils, SDKErrors } from '@kiltprotocol/utils'
+import { u8aToHex } from '@polkadot/util'
 import * as Claim from '../claim/index.js'
 import { hashClaimContents } from '../claim/index.js'
 import { verifyClaimAgainstSchema } from '../ctype/index.js'
@@ -415,11 +411,14 @@ export async function createPresentation({
     excludedClaimProperties
   )
 
-  const { signature, keyUri } = await signPayload(
-    credential.claim.owner,
-    makeSigningData(presentation, challenge),
-    signCallback
-  )
+  const { signature, keyUri } = await signCallback({
+    data: makeSigningData(presentation, challenge),
+    did: credential.claim.owner,
+    keyRelationship: 'authentication',
+  })
 
-  return { ...credential, claimerSignature: { signature, keyUri, challenge } }
+  return {
+    ...credential,
+    claimerSignature: { signature: u8aToHex(signature), keyUri, challenge },
+  }
 }
--- a/packages/core/src/quote/Quote.ts
+++ b/packages/core/src/quote/Quote.ts
@@ -26,7 +26,6 @@ import type {
 } from '@kiltprotocol/types'
 import { Crypto, JsonSchema, SDKErrors } from '@kiltprotocol/utils'
 import { resolve, verifyDidSignature } from '@kiltprotocol/did'
-import * as Did from '@kiltprotocol/did'
 import { QuoteSchema } from './QuoteSchema.js'
 
 /**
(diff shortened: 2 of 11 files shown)


tjwelde:
I am wondering, if the `makeDidSignature`, I put into the  TestUtils, would still be useful for people. What do you think?

packages/did/src/Did.chain.ts
```diff
@@ -404,13 +403,13 @@ export async function generateDidAuthenticatedTx({
  */
 export function didSignatureToChain(
   key: DidVerificationKey,
-  signature: Pick<DidSignature, 'signature'>
+  signature: Uint8Array
```
arty-name, line 406: This function didSignatureToChain is used in a single place, do we still need it?

tests/bundle-test.ts
```diff
@@ -50,7 +50,11 @@ function makeSignCallback(
         )
       }
       const signature = keypair.sign(data, { withType: false })
-      return { data: signature, keyUri: `${didDocument.uri}${keyId}`, keyType }
+      return {
+        signature: signature,
+        keyUri: `${didDocument.uri}${keyId}`,
```
arty-name, line 55:
```suggestion
        signature,
```

arty-name (commented):
I still need to look deeper into makeDidSignature()

tests/bundle-test.ts
arty-name: Oops, of course let’s keep the keyUri :)

packages/core/src/quote/Quote.ts
```diff
@@ -74,16 +73,16 @@ export async function createAttesterSignedQuote(
     throw new SDKErrors.QuoteUnverifiableError()
   }
 
-  const signature = await Did.signPayload(
-    quoteInput.attesterDid,
-    Crypto.hashObjectAsStr(quoteInput),
-    sign
-  )
+  const signature = await sign({
+    data: Crypto.coToUInt8(Crypto.hashObjectAsStr(quoteInput)),
```
arty-name, line 77: Could we do this here and elsewhere?
```suggestion
    data: Crypto.encodeObjectAsStr(quoteInput),
```

tjwelde: so not hashing? maybe?

tjwelde:
> `keyRelationship: 'authentication',` everywhere feels like annoying useless boilerplate btw

Yeah, I know what you mean, but I would still like to give the users the possibility to make a more generic callback and just tell them, what keytype the function they used expects. As long as the function does that internally, I don't see a big problem.

packages/did/src/Did.chain.ts
tjwelde: Not sure. Will users, who work with raw api calls need it?

packages/core/src/quote/Quote.ts
arty-name: Oh wait, I misread the code… Need to keep hashing,  it’s just that hashObjectAsStr converts to hex only for us to convert it back to UIntArray immediately. How about adding `hashObjectAsHex`? Maybe even removing `hashObjectAsStr`?

arty-name:
My preferred approach would be to remove `makeDidSignature`, update `verifyDidSignature` to only accept UInt8Arrays, and add a function `Did.signatureToJSON` that only does `return { signature: Crypto.u8aToHex(signature), keyUri }`. We’ll be using it in quite a few places then.

arty-name (approved):

ntn-x2 (approved):
For what it's worth, LGTM!

packages/core/src/quote/Quote.ts
tjwelde: I was also thinking about having such a function. It would be `hashObjectAsU8a` though

packages/core/src/credential/Credential.ts
```diff
@@ -418,7 +418,7 @@ export async function createPresentation({
   })
 
   return {
-    ...credential,
+    ...presentation,
```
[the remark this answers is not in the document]
rflechtner: oh never mind

packages/core/src/quote/Quote.ts
rflechtner: how about splitting into `sortAndStringify` and `Crypto.hash` / `Crypto.hashString` instead of having another two functions to cover these combinations?

rflechtner (changes requested):
Documentation needs to be adjusted, otherwise we're good to go

rflechtner:
> I am wondering, if the `makeDidSignature`, I put into the TestUtils, would still be useful for people. What do you think?

It pretty much does what the removed signPayload did, so no, I wouldn't export it. Would be another unnecessary wrapper in my eyes. It's trivial to implement if people feel that would make their code more concise.

rflechtner:
> My preferred approach would be to remove `makeDidSignature`, update `verifyDidSignature` to only accept UInt8Arrays, and add a function `Did.signatureToJSON` that only does `return { signature: Crypto.u8aToHex(signature), keyUri }`. We’ll be using it in quite a few places then.

I could get on board with that. Also though that using this wrapper function everywhere in the tests sends mixed signals.

packages/core/src/quote/Quote.ts
tjwelde: I like it

PR #642, opened 2022-09-22
test: breaking changes

tjwelde:
## fixes KILTProtocol/ticket#2081

Assert persistence of the following

- [ ]     All message types with their contents
  - [ ] | 'error'
  - [ ] | 'reject'
  - [x] | 'request-terms'
  - [x] | 'submit-terms'
  - [ ] | 'reject-terms'
  - [x] | 'request-attestation'  
  - [x] | 'submit-attestation'  
  - [ ] | 'reject-attestation'  
  - [ ] | 'request-payment'  
  - [ ] | 'confirm-payment'  
  - [x] | 'request-credential'  
  - [x] | 'submit-credential'  
  - [ ] | 'accept-credential'  
  - [ ] | 'reject-credential'  
  - [ ] | 'request-accept-delegation'  
  - [ ] | 'submit-accept-delegation'  
  - [ ] | 'reject-accept-delegation'  
  - [ ] | 'inform-create-delegation'
- [x]     Message envelope
  - [x] Created from contents
  - [x] Created from encrypted message
- [x]     Encrypted message
- [x]     Light did uri (with encryption key and service endpoints)
  - [x] Encode
  - [x] Decode
- [ ]     full did uri
- [x]     credential objects _(indirectly though, via being included in messages)_

## How to test:
If necessary, provide information about dependencies (specific configuration, branches, database dumps, etc.)

- Step 1
- Step 2
- etc.

## Checklist:


packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+  describe('Light DID', () => {
+    it('does not break the light did uri generation', () => {
+      const { did } = makeLightDidFromSeed(
+        '0x127f2375faf3472c2f94ffcdd5424590b27294631f2cb8041407e501bc97c44c'
+      )
+
+      expect(did.uri).toMatchInlineSnapshot(
+        `"did:kilt:light:004quk8nu1MLvzdoT4fE6SJsLS4fFpyvuGz7sQpMF7ZAWTDoF5:z1msTRicERqs59nwMvp3yzMRBhUYGmkum7ehY7rtKQc8HzfEx4b4eyRhrc37ZShT3oG7E89x89vaG9W4hRxPS23EAFnCSeVbVRrKGJmFQvYhjgKSMmrGC7gSxgHe1a3g41uamhD49AEi13YVMkgeHpyEQJBy7N7gGyW7jTWFcwzAnws4wSazBVG1qHmVJrhmusoJoTfKTPKXkExKyur8Z341EkcRkHteY8dV3VjLXHnfhRW2yU9oM2cRm5ozgaufxrXsQBx33ygTW2wvrfzzXsYw4Bs6Vf2tC3ipBTDcKyCk6G88LYnzBosRM15W3KmDRciJ2iPjqiQkhYm77EQyaw"`
+      )
+    })
```
rflechtner, line 83: I think we should test light DID decoding as well, as we had an issue with this breaking just now. We found out that the sdk no longer decodes some light DID details (services, encryption keys) created with older versions.

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+      const encrypted = await Message.encrypt(
+        message,
+        encrypt(aliceDid),
+        `${bobDid.uri}#encryption`,
+        {
+          resolveKey: makeResolveKey(bobDid),
+        }
+      )
+
+      expect(encrypted).toMatchSnapshot()
```
rflechtner, line 119: decrypting the message to make sure we reproduce the original message would be good

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+
+      const requestTerms: MessageBody = {
+        type: 'request-terms',
+        content: {
+          cTypeHash: CType.idToHash(cType.$id),
+        },
+      }
+      expect(requestTerms).toMatchSnapshot('request-terms')
+
+      const claim = Claim.fromCTypeAndClaimContents(
```
rflechtner, line 150: should be a snapshot in itself

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        type: 'submit-terms',
+        content: {
+          claim,
+          legitimations: [],
+        },
+      }
+      expect(submitTerms).toMatchSnapshot('submit-terms')
+
+      claim.owner = user.did.uri
+      const credential = Credential.fromClaim(claim, { legitimations: [] })
```
rflechtner, line 165: should be a snapshot in itself

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+      const requestAttestation: MessageBody = {
+        type: 'request-attestation',
+        content: { credential },
+      }
+      expect(requestAttestation).toMatchSnapshot('request-attestation')
+
+      const attestation = Attestation.fromCredentialAndDid(
```
rflechtner, line 172: could also be a snapshot

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+              requiredProperties: ['name'],
+              trustedAttesters: [attester.did.uri],
+            },
+          ],
+          challenge,
+        },
+      }
+      expect(requestCredential).toMatchSnapshot('request-credential')
+
+      const presentation = await Credential.createPresentation({
```
rflechtner, line 200: should also get a snapshot

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -80,14 +80,16 @@ describe('Breaking Changes', () => {
       expect(did.uri).toMatchInlineSnapshot(
         `"did:kilt:light:004quk8nu1MLvzdoT4fE6SJsLS4fFpyvuGz7sQpMF7ZAWTDoF5:z1msTRicERqs59nwMvp3yzMRBhUYGmkum7ehY7rtKQc8HzfEx4b4eyRhrc37ZShT3oG7E89x89vaG9W4hRxPS23EAFnCSeVbVRrKGJmFQvYhjgKSMmrGC7gSxgHe1a3g41uamhD49AEi13YVMkgeHpyEQJBy7N7gGyW7jTWFcwzAnws4wSazBVG1qHmVJrhmusoJoTfKTPKXkExKyur8Z341EkcRkHteY8dV3VjLXHnfhRW2yU9oM2cRm5ozgaufxrXsQBx33ygTW2wvrfzzXsYw4Bs6Vf2tC3ipBTDcKyCk6G88LYnzBosRM15W3KmDRciJ2iPjqiQkhYm77EQyaw"`
       )
+
+      expect(Did.parseDocumentFromLightDid(did.uri)).toMatchSnapshot()
```
rflechtner, line 84: my idea was to use a hard-coded did here to make sure that even if we update the snapshot above bc we change the way dids are _en_coded, we still make sure we can _de_code the original form

packages/testing/src/BreakingChanges.spec.ts
```diff
@@ -117,6 +119,12 @@ describe('Breaking Changes', () => {
       )
 
       expect(encrypted).toMatchSnapshot()
+
+      const decrypted = await Message.decrypt(encrypted, decrypt, {
+        resolveKey: makeResolveKey(aliceDid),
+      })
+
+      expect(decrypted).toMatchSnapshot()
```
rflechtner, line 127: wouldn't it be better to compare against `message` to make sure they are the same instead of making a second snapshot which you'd have to compare manually to the first?
```suggestion
      expect(decrypted).toMatchObject(message)
```

arty-name: Good idea! :+1:

packages/testing/src/BreakingChanges.spec.ts
arty-name: Sure, we can do that

rflechtner (approved):
let's merge this as-is

PR #646, opened 2022-09-26
feat: added KiltEncryptionKeypair

arty-name:
This PR adds a missing type for the encryption keypair: `KiltEncryptionKeypair`

## Checklist:

### DIFF
--- a/packages/testing/src/TestUtils.ts
+++ b/packages/testing/src/TestUtils.ts
@@ -7,16 +7,16 @@
 
 import { blake2AsHex, blake2AsU8a } from '@polkadot/util-crypto'
 
-import {
+import type {
   DecryptCallback,
   DidDocument,
   DidKey,
   DidServiceEndpoint,
   DidVerificationKey,
   EncryptCallback,
-  EncryptionKeyType,
   KeyRelationship,
   KeyringPair,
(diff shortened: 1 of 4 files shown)


tjwelde (approved):
LGTM

packages/testing/src/TestUtils.ts
```diff
@@ -73,10 +71,7 @@ export function makeEncryptCallback({
  */
 export function makeDecryptCallback({
   secretKey,
-}: {
-  secretKey: Uint8Array
-  type: 'x25519'
-}): DecryptCallback {
+}: KiltEncryptionKeypair): DecryptCallback {
```
ntn-x2, line 74: Decrypting does not need the public key here, right? At least that is not what the function was expecting before.

ntn-x2 (approved):
LGTM, besides a small nitpick

packages/testing/src/TestUtils.ts
arty-name: Yeah, but this here are just a bunch of testing utilities, it’s not too important here I think?

ntn-x2: Oh yeah, did not check what file it was 👍

PR #648, opened 2022-09-26
feat!: throw instead of return null

arty-name:
This PR replaces `return null` with `throw new Error()` in the few remaining functions

## Checklist:

### DIFF
--- a/packages/core/src/delegation/DelegationNode.ts
+++ b/packages/core/src/delegation/DelegationNode.ts
@@ -173,15 +173,6 @@ export class DelegationNode implements IDelegationNode {
   }
 
   /**
-   * Fetches the parent node of this delegation node.
-   *
-   * @returns Promise containing the parent as [[DelegationNode]] or [null].
-   */
-  public async getParent(): Promise<DelegationNode | null> {
-    return this.parentId ? query(this.parentId) : null
-  }
-
-  /**
    * Fetches the children nodes of this delegation node.
    *
    * @returns Promise containing the children as an array of [[DelegationNode]], which is empty if there are no children.
@@ -348,15 +339,22 @@ export class DelegationNode implements IDelegationNode {
         node: this,
       }
     }
-    const parent = await this.getParent()
-    if (parent) {
+    if (!this.parentId) {
+      return {
+        steps: 0,
+        node: null,
+      }
+    }
+    try {
+      const parent = await query(this.parentId)
       const result = await parent.findAncestorOwnedBy(did)
       result.steps += 1
       return result
-    }
-    return {
-      steps: 0,
-      node: null,
+    } catch {
+      return {
+        steps: 0,
+        node: null,
+      }
     }
   }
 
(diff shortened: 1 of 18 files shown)


packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -38,7 +38,7 @@ import { parse } from '../Did.utils.js'
  *
  * @returns The fetched [[DidDocument]], or null if DID does not exist.
```
tjwelde, line 39:
```suggestion
 * @returns The fetched [[DidDocument]], or throws Error if DID does not exist.
```

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -163,17 +163,12 @@ export function connectedAccountsFromChain(
  * @param linkedAccount The account to use for the lookup.
  * @returns The Web3 name linked to the given account, or `null` otherwise.
```
tjwelde, line 164:
```suggestion
 * @returns The Web3 name linked to the given account, or throws Error otherwise.
```

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -163,17 +163,12 @@ export function connectedAccountsFromChain(
...
  */
-export async function queryWeb3Name(
-  linkedAccount: Address
-): Promise<Web3Name | null> {
+export async function queryWeb3Name(linkedAccount: Address): Promise<Web3Name> {
```
tjwelde, line 166: I feel like `queries` should still return null, if nothing could be queried,

packages/did/src/DidDetails/FullDidDetails.ts
arty-name: Done!

packages/did/src/DidLinks/AccountLinks.chain.ts
arty-name: Done!

packages/did/src/DidLinks/AccountLinks.chain.ts
arty-name: Renamed all query to fetch :)

packages/core/src/delegation/DelegationNode.ts
```diff
@@ -167,35 +167,22 @@ export class DelegationNode implements IDelegationNode {
...
     }
     return this.hierarchyDetails
   }
 
-  /**
-   * Fetches the parent node of this delegation node.
-   *
-   * @returns Promise containing the parent as [[DelegationNode]] or [null].
-   */
-  public async getParent(): Promise<DelegationNode | null> {
```
tjwelde, line 185: This might still be relevant for externals

arty-name: Restored it in https://github.com/KILTprotocol/sdk-js/pull/648/commits/6360b90256f97d4207d2bee9764acad3c763d9d3

tjwelde (approved):
LGTM

PR #649, opened 2022-09-27
fix: missing JSDoc annotations and other lints

arty-name:
This PR cleans up some leftovers reported by eslint.

## Checklist:

### DIFF
--- a/packages/core/src/kilt/Kilt.ts
+++ b/packages/core/src/kilt/Kilt.ts
@@ -19,7 +19,7 @@ import { ConfigService } from '@kiltprotocol/config'
 import { latest, rpc, runtime } from '@kiltprotocol/type-definitions'
 
 /**
- * Prepares crypto modules (required e.g. for identity creation) and calls ConfigService.set().
+ * Prepares crypto modules (required for identity creation and others) and calls ConfigService.set().
  *
  * @param configs Arguments to pass on to ConfigService.set().
  * @returns Promise that must be awaited to assure crypto is ready.
--- a/packages/did/src/Did.utils.ts
+++ b/packages/did/src/Did.utils.ts
@@ -135,29 +135,35 @@ export function validateUri(
     throw new TypeError(`DID string expected, got ${typeof input}`)
   }
   const { address, fragment } = parse(input as DidUri)
-  switch (expectType) {
-    // for backwards compatibility with previous implementations, `false` maps to `Did` while `true` maps to `undefined`.
-    // @ts-ignore
-    case false:
-    case 'Did':
-      if (fragment)
-        throw new SDKErrors.DidError(
-          'Expected a Kilt DidUri but got a DidResourceUri (containing a #fragment)'
-        )
-      break
-    case 'ResourceUri':
-      if (!fragment)
-        throw new SDKErrors.DidError(
-          'Expected a Kilt DidResourceUri (containing a #fragment) but got a DidUri'
-        )
-      break
-    default:
-      break
+
+  if (
+    fragment &&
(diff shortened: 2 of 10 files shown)


packages/core/src/kilt/Kilt.ts
```diff
@@ -19,7 +19,7 @@ import { ConfigService } from '@kiltprotocol/config'
 import { latest, rpc, runtime } from '@kiltprotocol/type-definitions'
 
 /**
- * Prepares crypto modules (required e.g. for identity creation) and calls ConfigService.set().
+ * Prepares crypto modules (required for identity creation and others) and calls ConfigService.set().
```
arty-name, line 22: eslint is not smart enough to see this is not a sentence start

packages/did/src/Did.utils.ts
```diff
@@ -135,29 +135,35 @@ export function validateUri(
...
-        )
-      break
-    default:
-      break
+
+  if (
+    fragment &&
+    (expectType === 'Did' ||
+      // for backwards compatibility with previous implementations, `false` maps to `Did` while `true` maps to `undefined`.
+      (typeof expectType === 'boolean' && expectType === false))
```
arty-name, line 143: The switch statements cause lots of issues, like here only a ts-ignore could help to silence the error, while a common if statement can be written without overrides.

packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -129,23 +134,6 @@ function increaseNonce(currentNonce: BN, increment = 1): BN {
...
-/**
- * Returns all the DID keys that could be used to sign the provided extrinsic for submission.
- * This function should never be used directly by SDK users, who should rather call [[Did.authorizeExtrinsic]].
- *
- * @param did The DID data.
- * @param extrinsic The unsigned extrinsic to perform the lookup.
- *
- * @returns All the keys under the full DID that could be used to generate valid signatures to submit the provided extrinsic.
- */
-export function getKeysForExtrinsic(
```
arty-name, line 141: Removed as unused now

packages/utils/src/Crypto.ts
```diff
@@ -351,21 +351,41 @@ export function hashStatements(
...
 }
 
-export function makeEncryptionKeyFromSeed(
+/**
+ * Generate from a seed a x25519 keypair to be used as DID encryption key.
+ *
+ * @param seed The keypair seed, only optional in the tests.
+ * @returns The keypair.
+ */
+export function makeEncryptionKeypairFromSeed(
```
arty-name, line 388: Renamed to match the type name and the other `make*()` functions

packages/did/src/Did.utils.ts
```diff
@@ -135,29 +135,35 @@ export function validateUri(
...
+    throw new SDKErrors.DidError(
+      'Expected a Kilt DidResourceUri (containing a #fragment) but got a DidUri'
+    )
   }
 
   DataUtils.verifyKiltAddress(address)
 }
 
+/**
+ * Internal: derive the address part of the DID when it is created from authentication key.
```
tjwelde, line 160: Can use [@internal](https://tsdoc.org/pages/tags/internal/) for internal functions. This will even remove it from d.ts files.

arty-name: ESLint: Invalid JSDoc tag name "internal".(jsdoc/check-tag-names)

arty-name: While this is a nice thing to have, I’d rather have it introduced and consistently applied to the codebase in a separate PR.

PR #650, opened 2022-09-27
chore: rename DidBuilderError

arty-name:
This PR addresses a TODO remnant from #581 and renames a legacy error.

## Checklist:

### DIFF
--- a/packages/did/src/Did.utils.ts
+++ b/packages/did/src/Did.utils.ts
@@ -162,18 +162,14 @@ export function getAddressByKey({
   publicKey,
   type,
 }: Pick<DidVerificationKey, 'publicKey' | 'type'>): KiltAddress {
-  switch (type) {
-    case 'ed25519':
-    case 'sr25519':
-      return encodeAddress(publicKey, ss58Format)
-    case 'ecdsa': {
-      // Taken from https://github.com/polkadot-js/common/blob/master/packages/keyring/src/pair/index.ts#L44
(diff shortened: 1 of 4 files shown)


packages/did/src/Did.utils.ts
```diff
@@ -162,18 +162,14 @@ export function getAddressByKey({
...
-    case 'sr25519':
-      return encodeAddress(publicKey, ss58Format)
-    case 'ecdsa': {
-      // Taken from https://github.com/polkadot-js/common/blob/master/packages/keyring/src/pair/index.ts#L44
-      const pk = publicKey.length > 32 ? blake2AsU8a(publicKey) : publicKey
-      return encodeAddress(pk, ss58Format)
-    }
-    default:
-      throw new SDKErrors.DidBuilderError(`Unsupported key type "${type}"`)
+  if (type === 'ed25519' || type === 'sr25519') {
```
arty-name, line 165: The function is internal anyway and doesn’t have to check for unexpected broken types. 

So we don’t need the default case in the switch (have I said the switch statement is problematic?) and don’t have to throw this error here.

tjwelde (approved):
LGTM

packages/did/src/Did.utils.ts
tjwelde: Is it checked somewhere else, if the type might be broken?

arty-name: I guess some user-facing functions do that, but likely not consistently. Still I think the checks should rather be implemented there instead of mistrusting TS all over the codebase.

PR #651, opened 2022-09-29
feat!: verify* functions take `didResolveKey` parameter

arty-name:

This PR addresses the remaining outcomes of the code review I did before the refactoring.

One point was to remove code duplication which had the ripple effect across the `verify*()` functions and significantly increased the size of this PR.

## Checklist:


packages/did/src/Did.signature.ts
```diff
@@ -75,13 +75,16 @@ export async function verifyDidSignature({
...
     throw new SDKErrors.SignatureMalformedError(
-      `Signature key URI "${signature.keyUri}" invalid`
+      `Signature key URI "${inputUri}" invalid`
     )
 
   const keyId: UriFragment = fragment
   const keyUri: DidResourceUri = inputUri
 
+  // Cannot simply use resolveKey here as we need to check the key relationship.
+  // Shall we add this feature in the resolveKey and remove the duplicated code below?
```
arty-name, line 85: An example of this would be in 598c249d. Simplifies quite a few other things, I’d say. Also makes the `resolve()` effectively unused.

rflechtner: I agree this would be a common use case. We should either cater for it in resolveKey, or have this feature in `Did.getKey()`

rflechtner (approved):
LGTM

PR #656, opened 2022-10-06
refactor!: simplify CType

rflechtner:
## fixes KILTProtocol/ticket#2232

Redenomination of the `ICType` to what has formerly been the `ICTypeSchema`. This means that everything that used to require an `ICType` object - including an owner, hash and schema property - now requires just the value of the `schema` property instead. CTypes are therefore just a specific type of json schema, described by the `CTypeModel` metaschema.

Although this makes the value of the `$id` property the new canonical identifier of the CType, I have refrained from refactoring most other object to use it instead of the ctype hash, as this would significantly alter the api surface and reduce backwards compatibility.

### Changes to Data structures

- __CType__: Existing CType data can be consumed after this change by simply extracting the value of the `schema` property.
- __CTypeMetadata__: Breaking change to the data structure: instead of a `ctypeHash` property, the CTypeMetadata is expected to have a `cTypeId` property. Its value can be derived from `ctypeHash` by prefixing it with `kilt:ctype:`.
- All other data structures that reference a CType still do so using the ctype hash, which is contained in its id. This has not been changed at this point so as not to harm backwards compatibility. 

### Changes to interfaces

- We currently do not have any message type for exchanging ctype data, which is why messaging is __not affected__. All message types that referenced ctypes still do so using the ctype hash.
- Only the CType schema has ever been submitted to the blockchain, so that __no changes__ are necessary here.

## How to test:
Unit & integration tests.

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/Ctypes.spec.ts
+++ b/packages/core/src/__integrationtests__/Ctypes.spec.ts
@@ -72,9 +72,7 @@ describe('When there is an CtypeCreator and a verifier', () => {
     await submitTx(authorizedStoreTx, paymentAccount)
 
     expect(
-      CType.fromChain(
-        await api.query.ctype.ctypes(CType.getCTypeHashFromId(ctype.$id))
-      )
+      CType.fromChain(await api.query.ctype.ctypes(CType.idToHash(ctype.$id)))
     ).toBe(ctypeCreator.uri)
     await expect(CType.verifyStored(ctype)).resolves.not.toThrow()
   }, 40_000)
@@ -102,9 +100,7 @@ describe('When there is an CtypeCreator and a verifier', () => {
     ).rejects.toMatchObject({ section: 'ctype', name: 'CTypeAlreadyExists' })
 
     expect(
-      CType.fromChain(
-        await api.query.ctype.ctypes(CType.getCTypeHashFromId(ctype.$id))
-      )
+      CType.fromChain(await api.query.ctype.ctypes(CType.idToHash(ctype.$id)))
     ).toBe(ctypeCreator.uri)
   }, 45_000)
 
@@ -115,8 +111,7 @@ describe('When there is an CtypeCreator and a verifier', () => {
 
     await expect(CType.verifyStored(iAmNotThere)).rejects.toThrow()
     expect(
-      (await api.query.ctype.ctypes(CType.getCTypeHashFromId(iAmNotThere.$id)))
-        .isNone
+      (await api.query.ctype.ctypes(CType.idToHash(iAmNotThere.$id))).isNone
     ).toBe(true)
 
     const fakeHash = Crypto.hashStr('abcdefg')
--- a/packages/core/src/__integrationtests__/Ctypes.spec.ts
+++ b/packages/core/src/__integrationtests__/Ctypes.spec.ts
@@ -35,15 +35,9 @@ describe('When there is an CtypeCreator and a verifier', () => {
 
   function makeCType(): ICType {
     ctypeCounter += 1
-    return {
-      $id: `kilt:ctype:0x${ctypeCounter}`,
-      $schema: 'http://kilt-protocol.org/draft-01/ctype#',
-      title: `ctype${ctypeCounter}`,
-      properties: {
-        name: { type: 'string' },
-      },
-      type: 'object',
-    }
+    return CType.fromProperties(`ctype${ctypeCounter}`, {
+      name: { type: 'string' },
+    })
   }
 
   beforeAll(async () => {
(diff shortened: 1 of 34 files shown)


packages/core/src/ctype/CType.ts
```diff
@@ -38,18 +33,13 @@ import {
...
   // otherwise the resulting hash will be different, as the hash on chain would contain the CType ID,
   // which is itself a hash of the CType schema.
-  const schemaWithoutId: Partial<ICType['schema']> =
-    '$id' in ctypeSchema
-      ? (ctypeSchema as ICType['schema'])
-      : (ctypeSchema as CTypeSchemaWithoutId)
-  const shallowCopy = { ...schemaWithoutId }
-  delete shallowCopy.$id
-  return shallowCopy
+  const { $id: _, ...schemaWithoutId } = ctypeSchema as ICType
```
arty-name, line 42: Would that simpler version work?
```suggestion
  const { $id, ...schemaWithoutId } = ctypeSchema as ICType
```

rflechtner: would also work

packages/core/src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -108,42 +101,26 @@ describe('When there is an CtypeCreator and a verifier', () => {
...
-        $schema: 'http://kilt-protocol.org/draft-01/ctype#',
-        title: 'ctype2',
-        properties: {
-          game: { type: 'string' },
-        },
-        type: 'object',
-      },
-      ctypeCreator.uri
-    )
+    const iAmNotThere: ICType = CType.fromProperties('ctype2', {
```
arty-name, line 112: Do we need to explicitly mention the type here?

packages/core/src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -77,12 +71,11 @@ describe('When there is an CtypeCreator and a verifier', () => {
...
 
-    expect(CType.fromChain(await api.query.ctype.ctypes(ctype.hash))).toBe(
-      ctypeCreator.uri
-    )
-    await expect(CType.verifyStored(ctype)).resolves.not.toThrow()
-
-    ctype.owner = ctypeCreator.uri
+    expect(
+      CType.fromChain(
+        await api.query.ctype.ctypes(CType.getCTypeHashFromId(ctype.$id))
```
arty-name, line 76: Even if it would be just an alias, I think it would make for a consistent code to use `CType.toChain()` here

packages/core/src/credential/Credential.spec.ts
```diff
@@ -37,16 +36,11 @@ import * as Attestation from '../attestation'
 import * as Claim from '../claim'
 import * as CType from '../ctype'
 import * as Credential from './Credential'
+import { getCTypeHashFromId } from '../ctype'
```
arty-name, line 39: Could we use `CType.getCTypeHashFromId` in this file instead of the new import?

packages/core/src/ctype/CType.spec.ts
```diff
@@ -198,7 +132,7 @@ describe('blank ctypes', () => {
 
 describe('CType verification', () => {
   const ctypeInput = {
-    $id: 'kilt:ctype:0x1',
+    $id: 'kilt:ctype:0xdeadbeef',
```
arty-name, line 135: As a side note, in my code I try to avoid words with dark meaning :) I was shocked seeing someone putting `0xdeaddad1` somewhere.

packages/core/src/ctype/CType.ts
```diff
@@ -71,21 +56,34 @@ export function getHashForSchema(
  * @param hash CType hash as hex string.
  * @returns Schema id uri.
  */
-export function getIdForCTypeHash(
-  hash: ICType['hash']
-): ICType['schema']['$id'] {
+export function getIdForCTypeHash(hash: CTypeHash): ICType['$id'] {
```
arty-name, line 59: Since this is likely to be used as `CType.getIdForCTypeHash()` I’d prefer to rename it to `getIdForHash`

packages/core/src/ctype/CType.ts
```diff
@@ -71,21 +56,34 @@ export function getHashForSchema(
...
   return `kilt:ctype:${hash}`
 }
 
+/**
+ * Extracts the CType hash from a CType $id.
+ *
+ * @param id A CType id of the form 'kilt:ctype:0x[0-9a-f]'.
+ * @returns The CType hash as a zero-prefixed string of hex digits.
+ */
+export function getCTypeHashFromId(id: ICType['$id']): CTypeHash {
```
arty-name, line 69: Since this is likely to be used as `CType.getCTypeHashFromId()` I’d prefer to rename it to `getHashFromId`. Actually if we’re using `*ToChain()` functions, maybe this could be `idToHash()`? The other one could be `hashToId()` then.

arty-name (approved):

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: no, not necessarily

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: we do have `CType.toChain` however for serialising the CType for submission. So we'd have to call id `CType.idToChain`

packages/core/src/credential/Credential.spec.ts
rflechtner: oh I missed that one

packages/core/src/ctype/CType.spec.ts
rflechtner: there is not really a reason to change it from what it was, I think that's just because I had removed that id altogether and then added one again

packages/core/src/__integrationtests__/Ctypes.spec.ts
arty-name: Sounds good!

packages/core/src/ctype/CType.ts
rflechtner: it's definitely shorter

PR #659, opened 2022-10-10
fix: update type definitions

ntn-x2:
Update type definitions to match what is currently deployed on the `develop` branch.

When merging into `master` before releasing, the type definitions must be updated to match what the Spiritnet `master` branch expects. I did not find another way to address this.


arty-name (approved):

rflechtner (approved):
LGTM

PR #662, opened 2022-10-14
[DO NOT MERGE] feat: add new potential type definitions

ntn-x2:
Add runtime definitions as implemented in https://github.com/KILTprotocol/kilt-node/pull/423.

## Checklist
- [ ] Apply any changes if the node PR is updated
- [ ] Generate new TS type definitions to reflect the new runtime APIs


rflechtner (approved):
Nice, that looks real good!

ntn-x2:
@rflechtner I will wait for the 1.8.0 `latest` image to be pushed, to verify that it still works with that. We could then release with this PR in as the new runtime calls are not exposed via the SDK API, but only internally in the resolver and in the integration tests. What do you think?

PR #664, opened 2022-10-17
chore: remove strictResolve

rflechtner:
## fixes KILTProtocol/ticket#2249

Removes the `strictResolve` function, which has not yet been released and whose usefulness is unclear (see ticket).

## Checklist:


arty-name (approved):
I assume there are no more mentions of strictResolve in the codebase

PR #665, opened 2022-10-17
feat: DID spec compliant implementation of resolve 

rflechtner:
## fixes KILTProtocol/ticket#2283

Adds an implementation of the `resolve` function described by the resolution section of the DID specifications (https://www.w3.org/TR/did-core/#resolution) - rn I called it `compliantResolve`, because `DID.resolve` is taken in the sdk.

Main differences: 
- Returns an actual DID document, i.e. the result of `exportToDidDocument`.
- Returns a minimal DID document if a DID has been deleted or upgraded, instead of returning no document.
- Adds a `resolutionMetadata` structure, which can indicate resolution errors.
- Does not throw/return `null` for invalid/non-existent DIDs, but uses the `resolutionMetadata` error reporting.

To be able to add the `alsoKnownAs` field to the DidDocument, I also changed the signature of `resolve` to include the web3name in the returned data, if available.

## How to test:

Unit tests are included.

## Checklist:


ntn-x2 (approved):
Love it!

PR #666, opened 2022-10-18
refactor!: make verifyDidSignature accept Uint8Array

rflechtner:
## fixes KILTProtocol/ticket#2236

As per the ticket, this changes verifyDidSignature such that it takes signatures as a byte array instead of a hex string.
It also removes the makeDidSignature helper.

## How to test:

Unit tests have been adapted.

## Checklist:

### DIFF
--- a/packages/core/src/quote/Quote.ts
+++ b/packages/core/src/quote/Quote.ts
@@ -23,6 +23,7 @@ import type {
   SignCallback,
   DidResolveKey,
   DidUri,
+  DidResourceUri,
 } from '@kiltprotocol/types'
 import { Crypto, JsonSchema, SDKErrors } from '@kiltprotocol/utils'
 import { resolveKey, verifyDidSignature } from '@kiltprotocol/did'
@@ -103,7 +104,11 @@ export async function verifyAttesterSignedQuote(
 ): Promise<void> {
   const { attesterSignature, ...basicQuote } = quote
   await verifyDidSignature({
-    signature: attesterSignature,
+    signature: Crypto.coToUInt8(attesterSignature.signature),
+    keyUri:
+      attesterSignature.keyUri ??
+      // accept the old did signature format where keyUri was keyId
+      (attesterSignature as unknown as { keyId: DidResourceUri }).keyId,
     message: Crypto.hashStr(Crypto.encodeObjectAsStr(basicQuote)),
     expectedVerificationMethod: 'authentication',
     didResolveKey,
@@ -140,7 +145,11 @@ export async function createQuoteAgreement(
   const { attesterSignature, ...basicQuote } = attesterSignedQuote
 
   await verifyDidSignature({
-    signature: attesterSignature,
+    signature: Crypto.coToUInt8(attesterSignature.signature),
+    keyUri:
+      attesterSignature.keyUri ??
+      // accept the old did signature format where keyUri was keyId
+      (attesterSignature as unknown as { keyId: DidResourceUri }).keyId,
     message: Crypto.hashStr(Crypto.encodeObjectAsStr(basicQuote)),
     expectedVerificationMethod: 'authentication',
     didResolveKey,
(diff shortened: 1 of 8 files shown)


rflechtner:
@arty-name is that along the lines of what you had in mind?

packages/core/src/quote/Quote.ts
```diff
@@ -103,7 +104,11 @@ export async function verifyAttesterSignedQuote(
 ): Promise<void> {
   const { attesterSignature, ...basicQuote } = quote
   await verifyDidSignature({
-    signature: attesterSignature,
+    signature: Crypto.coToUInt8(attesterSignature.signature),
+    keyUri:
+      attesterSignature.keyUri ??
+      // accept the old did signature format where keyUri was keyId
+      (attesterSignature as unknown as { keyId: DidResourceUri }).keyId,
```
rflechtner, line 111: I reimplemented the backwards compatibility for the old DidSignature interface in its consumers. We could also drop that though, it's been a while and we have tons of breaking changes anyway?

packages/did/src/Did.signature.ts
```diff
@@ -54,32 +57,28 @@ function verifyDidSignatureDataStructure(
...
     )
 
-  const { publicKey } = await didResolveKey(
-    inputUri,
-    expectedVerificationMethod
-  )
+  const { publicKey } = await didResolveKey(keyUri, expectedVerificationMethod)
 
-  Crypto.verify(message, signature.signature, u8aToHex(publicKey))
+  Crypto.verify(message, signature, u8aToHex(publicKey))
```
ntn-x2, line 81: Why doesn't `Crypto.verify()` also take a `Uint8Array`? I thought we wanted to get rid of "useless" conversions between HEX string and byte arrays.

ntn-x2: At the end a raw public key is represented by bytes, not by a HEX string.

packages/did/src/Did.signature.ts
```diff
@@ -99,3 +98,18 @@ export function isDidSignature(
...
+ * @param input Signature data returned from the [[SignCallback]].
+ * @param input.signature Signature bytes.
+ * @param input.keyUri DID URI of the key used for signing.
+ * @returns A [[DidSignature]] object where signature is hex-encoded.
+ */
+export function signatureToJson({
+  signature,
+  keyUri,
+}: SignResponseData): DidSignature {
+  return { signature: Crypto.u8aToHex(signature), keyUri }
```
ntn-x2, line 114: Also here, I think the raw stuff should all be byte arrays, and hex conversion should only ever be needed when serialising to, e.g., print.

ntn-x2 (commented):
Can we maybe remove any leftover of byte array to HEX conversions? I feel we still do it quite few times and it does not seem to be necessary.

packages/did/src/Did.signature.ts
ntn-x2: I know it's out of scope for this PR, but this could potentially be another breaking change.

rflechtner: you're right, that could be fixed too

packages/did/src/Did.signature.ts
rflechtner: I think we're on the same page here, but this function is built explicitly for this serialisation use case. Our interfaces like credentials, quotes and so on are data formats that are designed with serialisation in mind, so I don't think Uint8Arrays work there.

ntn-x2: Alright, then sounds good 👍

packages/core/src/quote/Quote.ts
arty-name: I think last time we had to put in specifically in the SDK because this function was called automatically before the application had any chance to update the data, like during decryption of a message. I am not sure this is still the case, I think Timo decoupled the decryption from verification. Still this could be a problem for long-living data objects, like on DIDsign. It could be much cheaper to just keep this in.

arty-name (approved):

packages/core/src/quote/Quote.ts
rflechtner: moving this to the deserialiser is a good compromise I think

ntn-x2 (approved):

PR #670, opened 2022-10-26
fix: DID API types

wischli:
## fixes KILTProtocol/ticket#2235
* Fixes types for `RawDidLinkedInfo`
* Fixes runtime spec names. Unfortunately, `Peregrine` does not have a custom spec name 🙈 With the current ones, the type changes were not applied in my local tests.

## Checklist:

### DIFF
--- a/packages/type-definitions/src/index.ts
+++ b/packages/type-definitions/src/index.ts
@@ -113,27 +113,15 @@ const defaultTypesBundle: OverrideVersionedType[] = [
 // Current runtime version: 10730
 export const typesBundle: OverrideBundleType = {
   chain: {
-    'KILT Spiritnet': {
+    'kilt-spiritnet': {
       runtime: {
         ...didCalls,
         ...parachainStakingCalls,
       },
       types: defaultTypesBundle,
     },
-    'KILT Peregrine': {
-      runtime: {
-        ...didCalls,
-        ...parachainStakingCalls,
-      },
-      types: defaultTypesBundle,
-    },
-    'KILT Mashnet': {
-      runtime: {
-        ...didCalls,
-      },
-      types: defaultTypesBundle,
-    },
-    Development: {
+    // Includes Standalone and Peregrine runtimes
+    'mashnet-node': {
       runtime: {
         ...didCalls,
         ...parachainStakingCalls,
--- a/packages/type-definitions/src/types_10720.ts
+++ b/packages/type-definitions/src/types_10720.ts
@@ -14,8 +14,8 @@ export const types10720: RegistryTypes = {
     identifier: 'AccountId32',
     accounts: 'Vec<DidApiAccountId>',
     w3n: 'Option<Text>',
-    serviceEndpoints: 'Vec<RawServiceEndpoints>',
-    details: 'RawDidDetails',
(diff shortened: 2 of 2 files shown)


packages/type-definitions/src/types_10720.ts
```diff
@@ -14,8 +14,8 @@ export const types10720: RegistryTypes = {
     identifier: 'AccountId32',
     accounts: 'Vec<DidApiAccountId>',
     w3n: 'Option<Text>',
-    serviceEndpoints: 'Vec<RawServiceEndpoints>',
-    details: 'RawDidDetails',
+    serviceEndpoints: 'Vec<DidServiceEndpointsDidEndpoint>',
+    details: 'DidDidDetails',
   },
   RawServiceEndpoints: {
```
wischli, line 20: @ntn-x2 I wasn't sure whether I can remove these as well because they are still relevant for the RPCs. Maybe I should adjust the types in a new `types_10750.ts` file and remove all `Raw*` stuff in there?

packages/type-definitions/src/index.ts
```diff
@@ -113,27 +113,15 @@ const defaultTypesBundle: OverrideVersionedType[] = [
 // Current runtime version: 10730
 export const typesBundle: OverrideBundleType = {
   chain: {
-    'KILT Spiritnet': {
+    'kilt-spiritnet': {
```
ntn-x2, line 116: Are you sure this works? `chain` refers to the chain name, while `spec` refers to the spec name. Here I wanted to use the chain name because `mashnet-node` spec includes both standalone and `peregrine` (which is not ideal),

packages/type-definitions/src/types_10720.ts
ntn-x2: We don't have RPC definitions anyway. So it's fine to remove also the types used for those.

ntn-x2: People willing to use the RPC stuff will most likely have to use an older version of the type definitions, which still contains both RPCs and their type definitions.

ntn-x2 (approved):
LGTM! @rflechtner can you double check these quickly before merging? Do you spot anything that would break something?

ntn-x2:
> Should we add an integration test?

I am sure those will run fine, as they only target the latest versions. Perhaps there might be something else breaking for people using the old types with the new version, although I can't seem to spot anything suspicious.
Perhaps some sanity checks using the type definitions we will release as part of the SDK could be done as part of https://github.com/KILTprotocol/ticket/issues/2226. So we can fix anything broken in one patch release, also for very old types.

rflechtner:
Do we need to update api augmentation?

wischli:
> Do we need to update api augmentation?

Should not be required. I tested with the current SDK version and the proposed changes in a custom script and it worked fine for me.

ntn-x2:
Yeah I think what we would get is some deleted TS interfaces, which I guess would not hurt. They are never returned nor taken anywhere, so it is not a must. But if you have @rflechtner some time, we could re-run them either in this PR or before releasing. I have no strong opinion and for me this PR could also be merged as is assuming we don't find anything that breaks.

rflechtner (approved):
LGTM now, let's wait for the tests to pass though

PR #674, opened 2022-11-01
feat: compare owner and signer of credential on signature verification

rflechtner:
## fixes KILTProtocol/ticket#1773

Checking whether the signer of a credential is (related to) the owner has is so far not happening as part of `verifyPresentation`.
This adds this feature to `verifyDidSignature`, which is called by `verifyPresentation`.

> __Note:__ This cherry-picks a commit (01bf6fd1a98d5a75a18d90491c97b220ca94bb68) originally included with  https://github.com/KILTprotocol/sdk-js/pull/675.

## How to test:

Unit tests related to that issue did not exist yet. I added these.

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.spec.ts
+++ b/packages/core/src/credential/Credential.spec.ts
@@ -71,8 +71,6 @@ describe('Credential', () => {
     legitimation = buildCredential(identityAlice, {}, [])
   })
 
-  it.todo('signing and verification')
-
   it('verify credential', async () => {
     const credential = buildCredential(
       identityBob,
--- a/packages/core/src/credential/Credential.spec.ts
+++ b/packages/core/src/credential/Credential.spec.ts
@@ -481,10 +481,7 @@ describe('Credential', () => {
       signCallback: keyAlice.getSignCallback(identityAlice),
     })
     // but replace signer key reference with authentication key of light did
-    presentation.claimerSignature.keyUri = `${
-      Did.createLightDidDocument({ authentication: keyAlice.authentication })
-        .uri
-    }${identityAlice.authentication[0].id}`
+    presentation.claimerSignature.keyUri = `${identityDave.uri}${identityDave.authentication[0].id}`
 
     // signature would check out but mismatch should be detected
     await expect(
--- a/packages/core/src/credential/Credential.spec.ts
+++ b/packages/core/src/credential/Credential.spec.ts
@@ -362,13 +362,13 @@ describe('Credential', () => {
     identityCharlie = await createLocalDemoFullDidFromKeypair(
       keyCharlie.keypair
     )
-      ;[legitimation] = await buildPresentation(
-        identityAlice,
-        identityBob.uri,
-        {},
-        [],
-        keyAlice.getSignCallback(identityAlice)
-      )
+    ;[legitimation] = await buildPresentation(
+      identityAlice,
+      identityBob.uri,
+      {},
+      [],
+      keyAlice.getSignCallback(identityAlice)
+    )
   })
 
   it('verify credentials signed by a full DID', async () => {
@@ -392,7 +392,7 @@ describe('Credential', () => {
   })
   it('verify credentials signed by a light DID', async () => {
     const { getSignCallback, authentication } = makeSigningKeyTool('ed25519')
-    identityDave = await Did.createLightDidDocument({
+    identityDave = Did.createLightDidDocument({
       authentication,
     })
 
@@ -449,14 +449,24 @@ describe('Credential', () => {
       [legitimation]
     )
 
-    const presentation = await Credential.createPresentation({ credential, signCallback: getSignCallback(identityDave) })
+    const presentation = await Credential.createPresentation({
+      credential,
+      signCallback: getSignCallback(identityDave),
+    })
 
-    await expect(Credential.verifySignature(presentation, {
-      didResolveKey,
-    })).resolves.toThrow()
+    await expect(
+      Credential.verifySignature(presentation, {
+        didResolveKey,
+      })
+    ).rejects.toThrow(SDKErrors.DidSubjectMismatchError)
   })
 
   it('throws if signature is by corresponding light did', async () => {
+    // make mock resolver resolve corresponding light did by assigning it to dave identity
+    identityDave = Did.createLightDidDocument({
+      authentication: keyAlice.authentication,
+    })
+
     const credential = buildCredential(
       identityAlice.uri,
       {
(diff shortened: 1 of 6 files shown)


ntn-x2 (commented):
Can we maybe move the logic inside the `isSameSubject()` function? And modify it to perhaps take a `checkType` flag, which fails if the subject is the same but the DID is of a different type? Flag false would return `true` if the subject is the same regardless of the type.

Also I don't think we need directionality in the type check, right? If one is light and one is full, it is always an error, so that could simplify the verification logic a bit?

rflechtner:
> Can we maybe move the logic inside the `isSameSubject()` function? And modify it to perhaps take a `checkType` flag, which fails if the subject is the same but the DID is of a different type? Flag false would return `true` if the subject is the same regardless of the type.
> 
> Also I don't think we need directionality in the type check, right? If one is light and one is full, it is always an error, so that could simplify the verification logic a bit?
> 
> Like signing something with a light DID when a full DID exist is an error, and signing a DID with a full DID that does not exist is also an error.

I was looking into how this interacts with the `isSameSubject` function but the check we do here seems really specific to credentials. We have a situation here where we want to allow creds owned by a light did to be signed by the corresponding full did, but not the other way around. That implies directionality, which seems out of scope for `isSameSubject`. 
In fact this check is so specific to this use case, that I even decided against having it as a feature on did signature verification more generally.

packages/core/src/credential/Credential.spec.ts
```diff
@@ -433,6 +431,69 @@ describe('Credential', () => {
...
+
+    // sign presentation using Alice's authenication key
+    const presentation = await Credential.createPresentation({
+      credential,
+      signCallback: keyAlice.getSignCallback(identityAlice),
+    })
+    // but replace signer key reference with authentication key of light did
+    presentation.claimerSignature.keyUri = `${
+      Did.createLightDidDocument({ authentication: keyAlice.authentication })
+        .uri
```
arty-name, line 486:
```suggestion
      identityDave.uri
```

packages/core/src/credential/Credential.ts
```diff
@@ -232,6 +233,35 @@ export async function verifySignature(
...
+    } else {
+      // owner did is light did -> full- or light did with same subject may sign
+      if (ownerAddress !== signerAddress)
+        throw new SDKErrors.DidSubjectMismatchError(signer, owner)
+      // if signer is also a light did, require both to use the same key type
+      if (signerType === 'light' && ownerKeyType !== signerKeyType)
+        throw new SDKErrors.DidSubjectMismatchError(signer, owner)
+    }
+  }
+
```
arty-name, line 264: The logic is fine, however do you think something like that will be more elegant? 
```suggestion

  const owner = parse(input.claim.owner)
  const signer = parse(input.claimerSignature.keyUri)

  // check if credential owner matches signer
  if (owner.did !== signer.did) {
    // owner did is full did -> exact match required (signer cannot be light did)
    const typesMismatch = owner.type === 'full'

    // owner did is light did -> full- or light did with same subject may sign
    const subjectMismatch = owner.address !== signer.address

    // if signer is also a light did, require both to use the same key type
    const keyTypeMismatch =
      signer.type === 'light' &&
      owner.authKeyTypeEncoding !== signer.authKeyTypeEncoding

    if (typesMismatch || subjectMismatch || keyTypeMismatch) {
      throw new SDKErrors.DidSubjectMismatchError(signer.did, owner.did)
    }
  }
```

arty-name (approved):

packages/core/src/credential/Credential.ts
rflechtner: oh nice, that does look better indeed

packages/did/src/Did.signature.ts
```diff
@@ -59,23 +62,46 @@ function verifyDidSignatureDataStructure(
...
+    const expected = parse(expectedSigner)
+    // NECESSARY CONDITION: subjects and versions match
+    const subjectVersionMatch =
+      expected.address === signer.address && expected.version === signer.version
+    // EITHER: signer is a full did and we allow signatures by corresponding full did
+    const allowedUpgrade = allowUpgraded && signer.type === 'full'
+    // OR: both are light dids and their auth key type matches
+    const keyTypeMatch =
+      signer.type === 'light' &&
+      expected.type === 'light' &&
```
rflechtner, line 98: I'm not sure these two checks are necessary - a key type match could happen if both dids are full dids and do not have a key type indicator, but in that case identical addresses could only mean they are the same did?

rflechtner:
I ended up pulling the logic into `verifyDidSignature` because https://github.com/KILTprotocol/sdk-js/pull/675 requires something similar. What do you think, @ntn-x2 ?

packages/did/src/Did.signature.spec.ts
```diff
@@ -139,10 +140,11 @@ describe('light DID', () => {
     })
     keyUri += '1a'
     jest.mocked(resolveKey).mockRejectedValue(new Error('Key not found'))
-    await expect(() =>
+    await expect(
       verifyDidSignature({
         message: SIGNED_STRING,
         signature,
+        // @ts-ignore Appending the uri has changed its type
```
arty-name, line 147: Would it make more sense to rather cast the URI type above? I am not a fan of ts-ignore and ts-expect-error…

packages/did/src/Did.signature.ts
```diff
@@ -59,23 +62,46 @@ function verifyDidSignatureDataStructure(
...
+  expectedSigner,
+  allowUpgraded = false,
   expectedVerificationMethod,
   didResolveKey = resolveKey,
 }: DidSignatureVerificationInput): Promise<void> {
   // Verification fails if the signature key URI is not valid
-  const { fragment } = parse(keyUri)
-  if (!fragment)
+  const signer = parse(keyUri)
+  if (!signer.fragment)
```
arty-name, line 81: Not introduced in this PR, but: doesn’t the didResolveKey below perform the same check? It could be removed from here then…

packages/did/src/Did.signature.ts
```diff
@@ -59,23 +62,46 @@ function verifyDidSignatureDataStructure(
...
+    const subjectVersionMatch =
+      expected.address === signer.address && expected.version === signer.version
+    // EITHER: signer is a full did and we allow signatures by corresponding full did
+    const allowedUpgrade = allowUpgraded && signer.type === 'full'
+    // OR: both are light dids and their auth key type matches
+    const keyTypeMatch =
+      signer.type === 'light' &&
+      expected.type === 'light' &&
+      expected.authKeyTypeEncoding === signer.authKeyTypeEncoding
+    if (!(subjectVersionMatch && (allowedUpgrade || keyTypeMatch))) {
```
arty-name, line 100: I like this condition even better than the one I have proposed :+1:

packages/did/src/Did.signature.ts
arty-name: Indeed, it looks like checking for the light type can be skipped

ntn-x2 (approved):
I like it!

packages/did/src/Did.signature.ts
rflechtner: it does, I was assuming this was introduced to provide better error reporting?

arty-name: Could it be that it just was not removed when the previous duplicated code was replaced with resolveKey?

packages/did/src/Did.signature.spec.ts
arty-name: For me personally it would be preferable

packages/did/src/Did.signature.ts
rflechtner: possibly. also the error on resolveKey is misleading... I'll fix that

packages/did/src/Did.signature.spec.ts
arty-name: Oh yeah, the template string is even better!

packages/did/src/Did.signature.ts
arty-name: Nice!

packages/did/src/Did.signature.spec.ts
rflechtner: props to eslint :D

PR #675, opened 2022-11-01
feat: improve quote verification

rflechtner:
## fixes KILTProtocol/ticket#2188

- Sign `claimerDid` and credential `rootHash` into quote agreement
- Check that attester signature can be matched to attesterDid and claimer signature to claimerDid
- Make Quote agreement verification easier with new `verifyQuoteAgreement` function. 

> __Note:__ This shares a commit (01bf6fd1a98d5a75a18d90491c97b220ca94bb68) with https://github.com/KILTprotocol/sdk-js/pull/674.

## How to test:

Unit tests included

## Checklist:

### DIFF
--- a/packages/types/src/Quote.ts
+++ b/packages/types/src/Quote.ts
@@ -28,5 +28,6 @@ export interface IQuoteAttesterSigned extends IQuote {
 
 export interface IQuoteAgreement extends IQuoteAttesterSigned {
   rootHash: ICredential['rootHash']
+  claimerDid: DidUri
   claimerSignature: DidSignature
 }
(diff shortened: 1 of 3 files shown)


packages/types/src/Quote.ts
```diff
@@ -28,5 +28,6 @@ export interface IQuoteAttesterSigned extends IQuote {
 
 export interface IQuoteAgreement extends IQuoteAttesterSigned {
   rootHash: ICredential['rootHash']
+  claimerDid: DidUri
```
ntn-x2, line 31: Doesn't `DidSignature` already contains the key URI? Do we need also the `claimerDid`?

rflechtner: The point is that because the key URI is part of the signature, it is not signed into the credential... so you could change the claiming party after the fact, as long as the key exists on the other DID document. And the claimerDid is the one that is supposed to go into the credential, so forcing claimers to commit to an identity in the agreement seems to make sense

ntn-x2 (approved):

PR #676, opened 2022-11-03
feat: query attestation for credential

rflechtner:
## fixes KILTProtocol/ticket#2103

The ticket is flagged for discussion. If we wanted such a function (for querying an Attestation to a Credential), this is what I could imagine it looking like.

What do you think, @ntn-x2 & @arty-name, do we need this? In that case I would add tests and prepare for merge.

## How to test:

TBD

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -17,6 +17,7 @@
  * @packageDocumentation
  */
 
+import { ConfigService } from '@kiltprotocol/config'
 import {
   isDidSignature,
   verifyDidSignature,
@@ -36,6 +37,8 @@ import type {
   SignCallback,
 } from '@kiltprotocol/types'
 import { Crypto, DataUtils, SDKErrors } from '@kiltprotocol/utils'
+import { fromChain } from '../attestation/Attestation.chain.js'
+import { verifyAgainstCredential } from '../attestation/Attestation.js'
 import * as Claim from '../claim/index.js'
 import { hashClaimContents } from '../claim/index.js'
 import { verifyClaimAgainstSchema } from '../ctype/index.js'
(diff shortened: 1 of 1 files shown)


arty-name (commented):
First thing: I don’t see any check for `revoked !== true` anywhere

Besides that, I was under the impression, that simply querying the attestation by the rootHash of the credential was a reliable enough check, and didn’t need any additional comparisons of cType. 

I cannot really judge the necessity of this function as I don’t know its place in the application developer usage. If they receive a free-form credential object and simply feed it into this function to make sure the credential is currently valid, I think we can have it with more guards. Otherwise it’s necessity is questionable, I believe.

rflechtner:
Okay it seems that this ticket is unclear not only to me but also you guys wonder whether we need it - and have different ideas of what it says.
@arty-name according to the ticket we wanted a function to query an attestation, so yes that would return something - the attestation. It would also return the attestation if it is revoked, but IMO it should throw if the attestation contains data that does not fit the credential - which includes checking for the CType. This is actually an important step where attestations are delegated, bc delegation trees are only valid for a specific CType.
@ntn-x2 Seems like we have a similar understanding of what the function should do, but a different understanding of what it actually does - because as far as I am concerned, what you described is exactly what it does. Can you elaborate?

On the question of whether we need it, it is actually fairly straightforward to query an attestation for a credential as we all know (https://docs.kilt.io/docs/develop/workshop/verification#verify).
But all the checks you need to make during verification are not to clear; you'd need to compare CTypes, check the attester (even when refreshing the status!), and check revocation status. That may be a reason to have an extra function for that?

ntn-x2:
> @ntn-x2 Seems like we have a similar understanding of what the function should do, but a different understanding of what it actually does - because as far as I am concerned, what you described is exactly what it does. Can you elaborate?

So the ticket name is wrong, indeed. It does not fetch the attestation info, since that is already possible, but it verifies whether a credential is valid, including its attestation info. If that is the case, then it makes sense to have this auxiliary function, but that means that simply calling `Credential.verifyCredential()` is then not enough.

rflechtner:
> > @ntn-x2 Seems like we have a similar understanding of what the function should do, but a different understanding of what it actually does - because as far as I am concerned, what you described is exactly what it does. Can you elaborate?
> 
> So the ticket name is wrong, indeed. It does not fetch the attestation info, since that is already possible, but it verifies whether a credential is valid, including its attestation info. If that is the case, then it makes sense to have this auxiliary function, but that means that simply calling `Credential.verifyCredential()` is then not enough.

Well yes and no. Almost seems like we can't have a function that checks if a credential is valid because that would require the expected/accepted attester(s) as input. This goes a long way though and only requires you to check the revocation flag and compare the attester (owner) property.

arty-name:
> you'd need to compare CTypes, check the attester

Why is that?

rflechtner:
closing as not planned right now

PR #677, opened 2022-11-03
feat: public credentials and asset DIDs

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2092 and fixes https://github.com/KILTprotocol/ticket/issues/2175.

Original draft document (might not be entirely up to date with what's implemented here) -> https://hackmd.io/@EAX5fDsURyKJJ3GahxyZbA/Skoh424wq.

## What is this

This PR adds support for the long-awaited public credentials and asset DIDs.

A public credential behaves differently than a traditional KILT credential since it is publicly accessible by anyone that has access to a full archive node.
A credential is not stored inside the blockchain state, but a pointer to the block number in which it was created is stored instead.
When retrieving a credential, two operations must be performed: 
1. retrieve the details from the blockchain state
2. parse the block with the given number "hunting" for the right call, which contains the content of the credential (i.e., the actual claims)

### How are credentials supposed to be used and shared [SEE PR DISCUSSION BELOW FOR NEWER INFORMATION]

I envisioned that public credentials would never be shared directly between parties, because they are public. They would be used in one of the following two ways:

1. By **asset**: a party know the AssetDID of an asset and queries a KILT full node to obtain all the public credentials issued to it
2. By **identifier**: two parties can exchange credentials by simply sharing their IDs, which is sufficient to retrieve the full content by combining the query and RPC calls as mentioned above. There might be a better way to share credentials, such that the credential is shared as a whole, but I did not find a good way to modify the functions to make that work, as basically performing all the required checks involves fetching the actual credential from the blockchain and verifying it, which becomes redundant if the credential has already been shared.

Most of the code has been inspired by how credentials and attestations have been modelled in the SDK, since a public credential is kind of a combination of both.

### What about AssetDIDs

Since they are simply a wrapper over CAIP-19 identifier, which are simply "regexable", the new package does not do much. Eventually, we will develop an AssetDID resolver that can attach additional information to an AssetDID, such as the credentials linked to it, but I think that is out of scope from this PR, which represents already a good-enough PoC.


ntn-x2:
@arty-name @rflechtner I am not sure if I have missed something when modelling these new credentials. I have followed what has been done for normal credentials, but I see that they are more different than I thought. So the way I envisioned them is to exchange credential IDs rather then full credentials, but I am open to add the possibility to verify a full credential that has been received (by re-computing its ID, and query its details from the blockchain state; that should be enough to verify it since the ID is generated from the content).

ntn-x2:
@arty-name @rflechtner I am actually happy with the changes I introduced in https://github.com/KILTprotocol/sdk-js/pull/677/commits/ab209bb03b1c1cf4ce5c7a2af222dd581a5bea2d. 
I removed public functions that could be misused, and added comments about how to deal with public credentials, i.e., that they should always be fetched given their ID, or given the DID of the asset. I think this is the way to go, as I do not see any point in sending over the whole thing, which then has to be re-checked.

ntn-x2:
@arty-name @rflechtner ok last message I promise 😃 I gave it a third thought, and perhaps it makes sense to add a function that verifies a credential that has been received by another party. I added that in a `verifyCredential()` function in https://github.com/KILTprotocol/sdk-js/pull/677/commits/64e84a6a20853a5343fed71c9d82c1f11ecd4c56. I would like to have your opinion about it, before adding some unit tests for that.

packages/core/src/publicCredential/PublicCredential.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  assetDid: AssetDidUri,
+  contents: IClaimContents
+): IPublicCredential {
+  const claim: IAssetClaim = {
+    cTypeHash: CType.idToChain(testCType.$id),
+    contents,
+    subject: assetDid,
+  }
+  const credential = PublicCredential.fromClaim(claim)
+  const attester: DidUri = `did:kilt:${devAlice.address}`
```
arty-name, line 45: Is there a way to generate a DidUri which is not so coupled to the implementation details?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+import { HexString } from '@polkadot/util/types'
+import { ConfigService } from '@kiltprotocol/config'
+import { fromChain as didFromChain } from '@kiltprotocol/did'
+import { validateUri } from '@kiltprotocol/asset-did'
+import { SDKErrors } from '@kiltprotocol/utils'
+
+import { computeId } from './PublicCredential.js'
+
+export interface EncodedPublicCredential {
+  ctypeHash: CTypeHash
```
arty-name, line 37: Could that be `cTypeHash` or is it already on the chain like this?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+
+// Flatten any nested batch calls into a single list of calls.
+function flattenCalls(api: ApiPromise, call: Call): Call[] {
+  if (
+    api.tx.utility.batch.is(call) ||
+    api.tx.utility.batchAll.is(call) ||
+    api.tx.utility.forceBatch.is(call)
+  ) {
+    // Inductive case
+    return call.args[0].reduce(
```
arty-name, line 73: I believe a `flatMap` would be simpler here

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Transform a blockchain-formatted public credential into the original [[IPublicCredentialInput]].
+// It throws if what was written on the chain was garbage.
+function credentialInputFromChain(
+  credential: PublicCredentialsCredentialsCredential
+): IPublicCredentialInput {
+  validateUri(credential.subject.toUtf8())
+  return {
+    claims: cborDecode(hexToU8a(credential.claims.toHex())),
```
arty-name, line 89: Could claims be directly converted to u8a, skipping the hex step?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+  api: ApiPromise,
+  credentialId: HexString,
+  blockNumber: u64
+): Promise<Extrinsic | null> {
+  const { extrinsics } = await api.derive.chain.getBlockByNumber(blockNumber)
+  return (
+    extrinsics
+      // Filter out failed extrinsics
+      .filter(({ dispatchError }) => dispatchError === undefined)
+      // Filter out extrinsics that don't contain a `CredentialStored` event with the right credential ID
```
arty-name, line 107: I believe that self-documenting code is more expressive and more future-proof at the same time. For example, these two lines could be replaced with `const successfulExtrinsics = extrinsics.filter(({ dispatchError }) => dispatchError === undefined)`. Also, would a simple `!dispatchError` work?

arty-name: The same could be applied to the following steps in the chain as well

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+
+// Retrieve a given block and looks into it to find a public credential creation tx that matches the provided credential ID.
+async function retrievePublicCredentialCreationExtrinsicsFromBlock(
+  api: ApiPromise,
+  credentialId: HexString,
+  blockNumber: u64
+): Promise<Extrinsic | null> {
+  const { extrinsics } = await api.derive.chain.getBlockByNumber(blockNumber)
```
arty-name, line 102: Should they be flattened as well?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+  return {
+    claims: cborDecode(hexToU8a(credential.claims.toHex())),
+    cTypeHash: credential.ctypeHash.toHex(),
+    delegationId: credential.authorization.unwrapOr(undefined)?.toHex() ?? null,
+    subject: credential.subject.toUtf8() as AssetDidUri,
+  }
+}
+
+// Retrieve a given block and looks into it to find a public credential creation tx that matches the provided credential ID.
+async function retrievePublicCredentialCreationExtrinsicsFromBlock(
```
arty-name, line 97: Should it use singular `Extrinsic`?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+    blockNumber
+  )
+
+  if (extrinsic === null) {
+    throw new SDKErrors.PublicCredentialError(
+      `The block number as specified in the provided credential entry (${blockNumber}) does not have any extrinsic that includes a credential creation.`
+    )
+  }
+
+  if (!api.tx.did.submitDidCall.is(extrinsic)) {
```
arty-name, line 151: Do we need to do this check provided we have already checked `api.events.publicCredentials.CredentialStored.is(event)` above?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+    throw new SDKErrors.PublicCredentialError(
+      'Extrinsic should be a did.submitDidCall extrinsic'
+    )
+  }
+
+  const [extrinsicCalls, extrinsicDidOrigin] = [
+    flattenCalls(api, extrinsic.args[0].call),
+    didFromChain(extrinsic.args[0].did),
+  ]
+
```
arty-name, line 161: Would two simple const assignments work here as well?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+  let credentialInput: IPublicCredential | undefined
+
+  // Iterate over the calls in the extrinsic to find the right one, and re-create the issued public credential.
+  // If more than a call are present, it always consider the last one as the valid one.
+  extrinsicCalls.forEach((call) => {
```
arty-name, line 166: Grammar?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (api.tx.publicCredentials.add.is(call)) {
```
arty-name, line 167: I have several suggestions on this loop. First, early returns reduce nesting and improve code clarity. Second, if we’re only interested in the last one, we could reverse the array and look for the first one. Finally, I have a hunch that splitting the loop into several calls to `filter` and `map` might increase clarity and readability even further.

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+export async function credentialsFromChain(
+  publicCredentialEntries: Result<
+    Vec<ITuple<[Hash, PublicCredentialsCredentialsCredentialEntry]>>,
+    PublicCredentialError
+  >
+): Promise<IPublicCredential[]> {
+  if (publicCredentialEntries.isErr) {
+    throw new Error(publicCredentialEntries.asErr.toString())
+  }
```
arty-name, line 212: I believe the call of `toString()` is implicit here as well

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  const api = ConfigService.get('api')
+  const formattedCredentials: Array<
+    [HexString, Option<PublicCredentialsCredentialsCredentialEntry>]
```
arty-name, line 216: Doesn’t TS infer this type automatically?

packages/core/src/__integrationtests__/PublicCredentials.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+        name: `Certified NFT collection with id ${UUID.generate()}`,
+      },
+      cTypeHash: CType.getHashForSchema(nftNameCType),
+      delegationId: null,
+      subject: assetId,
+    }
+    const encodedPublicCredential = PublicCredential.toChain(latestCredential)
+    const storeTx = api.tx.publicCredentials.add(encodedPublicCredential)
+    const authorizedStoreTx = await Did.authorizeTx(
+      attester.uri,
```
arty-name, line 83: Will the naming become simpler if we inline the `storeTx` constant? The same applies to more transactions in the tests below.

packages/core/src/__integrationtests__/PublicCredentials.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+      storeTx,
+      attesterKey.getSignCallback(attester),
+      tokenHolder.address
+    )
+    await submitTx(authorizedStoreTx, tokenHolder)
+
+    const encodedAssetCredentials =
+      await api.call.publicCredentials.getCredentials(assetId, null)
+    const assetCredentials = await PublicCredential.credentialsFromChain(
```
arty-name, line 135: Does having `encodedAssetCredentials` as a constant help? We’re only using it once, and it’s quite clear what the function returns.

packages/core/src/__integrationtests__/PublicCredentials.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+    })
+    await submitTx(authorizedBatch, tokenHolder)
+
+    const encodedAssetCredentials =
+      await api.call.publicCredentials.getCredentials(assetId, null)
+    const assetCredentials = await PublicCredential.credentialsFromChain(
+      encodedAssetCredentials
+    )
+
+    // We only check that we return all the twenty credentials back.
```
arty-name, line 173: This is the reason I prefer self-documenting code to comments. It’s very easy to forget to update the comment when you’re updating the code. It’s harder to forget to update the code.

packages/core/src/__integrationtests__/PublicCredentials.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+    encodedAllAssetCredentials =
+      await api.call.publicCredentials.getCredentials(assetId, null)
+    const allAssetCredentialsAfterRevocation =
+      await PublicCredential.credentialsFromChain(encodedAllAssetCredentials)
+
+    // Verify it is now removed
+    expect(encodedAssetCredential.isNone).toBe(true)
+    // Verify the number of credentials has decreased by one
+    expect(allAssetCredentialsBeforeRevocation.length).toEqual(
+      allAssetCredentialsAfterRevocation.length + 1
```
arty-name, line 308: I was confused by mismatching order and "decreased" vs "+1". Could this be aligned?

packages/types/src/index.ts
```diff
@@ -16,6 +16,7 @@ export type { KeyringPair } from '@polkadot/keyring/types'
 export * as SubscriptionPromise from './SubscriptionPromise.js'
 
 export * from './Attestation.js'
+export * from './AssetDid.js'
```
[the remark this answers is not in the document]
ntn-x2: Like what? They are taking from the [AssetDID](https://github.com/KILTprotocol/spec-asset-did] spec which in turns takes them from the [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md) and [CAIP-19](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md) definitions.

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -0,0 +1,175 @@
...
+    )
+  }
+  if (revoked !== credential.revoked) {
+    throw new SDKErrors.PublicCredentialError(
+      `Revocation status in the credential (${credential.revoked}) different than what is stored on the blockchain (${credential.revoked}).`
+    )
+  }
+}
+
+export type Options = {
```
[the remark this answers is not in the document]
ntn-x2: I took it from the Credential stuff. What would you suggest?

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: Unfortunately it did not work, as u8A expects stuff to be SCALE-encoded, so calling `hexToU8a(credential.claims)` does not work.

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: Yes, because the event could be generated in multiple ways in the future, and this code makes sure that our assumptions are indeed correct. I think this is a viable approach, don't you think?

packages/types/src/index.ts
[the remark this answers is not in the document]
arty-name: Does this look okay to you? Remember, there are many other things in the SDK beyond this topic.
```ts
import type { ChainId } from '@kiltprotocol/types'

File: packages/core/src/publicCredential/PublicCredential.ts (hunk 19 above)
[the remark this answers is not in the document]
Comment (arty-name) [reply packages/core/src/publicCredential/PublicCredential.ts:151]: Same as in the other example above, I want to avoid the following:
```ts
import type { Options } from '@kiltprotocol/types'
```
Actually, I am generally against exporting types of arguments of functions, but when this is unavoidable, I try to include the function name in the type name.

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: Would this work?
```ts
cborDecode(credential.claims.toU8a()),

File: packages/core/src/publicCredential/PublicCredential.chain.ts (hunk 8 above)
Comment (arty-name) [reply packages/core/src/publicCredential/PublicCredential.chain.ts:151]: Okay, I was just making sure :)

File: packages/core/src/publicCredential/PublicCredential.chain.ts (hunk 11 above)
Comment (rflechtner) [reply packages/core/src/publicCredential/PublicCredential.chain.ts:167]: Maybe even `find` instead of `filter` given that it's built to return on the first match?

File: packages/types/src/index.ts (hunk 18 above)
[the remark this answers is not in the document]
Comment (ntn-x2) [reply packages/types/src/index.ts:19]: Would you rather namespace them? Or simply add a prefix for these types?

**Review by rflechtner — COMMENTED** [review]
Some high level questions first: 
- You were asking about ideas how public credentials could be verified without looking them up from the block of issuance. I think there could be a way where we combine credential data with attester data on chain while the extrinsic is being processed and trigger an event containing both the resulting object and its hash. If the hash becomes the credential id, then looking up the attestation should suffice if you have the combined object. Did we decide against using events with public credentials and if so why?
- Don't you think public credentials are better bundled up with asset dids than added to the `core` package? They seem like an isolated module that doesn't interact much with other functionality in `core` and has completely different flows and use cases than regular credentialing. Projects are unlikely to use regular credentials if they use public credentials and the other way around. So grouping them accordingly makes sense to me, allowing for leaner and more targeted imports

**Comment by ntn-x2** [comment]
> You were asking about ideas how public credentials could be verified without looking them up from the block of issuance. I think there could be a way where we combine credential data with attester data on chain while the extrinsic is being processed and trigger an event containing both the resulting object and its hash. If the hash becomes the credential id, then looking up the attestation should suffice if you have the combined object. Did we decide against using events with public credentials and if so why?

Isn't that what's been implemented in the `verifyCredential` function? We re-compute the ID of the provided credential and try to look it up in the block state. If we find a match, the credential exists. If we don't, the credential is wrong.

> Don't you think public credentials are better bundled up with asset dids than added to the core package? They seem like an isolated module that doesn't interact much with other functionality in core and has completely different flows and use cases than regular credentialing. Projects are unlikely to use regular credentials if they use public credentials and the other way around. So grouping them accordingly makes sense to me, allowing for leaner and more targeted imports

I would like @arty-name opinion about this, as well. For me, a public credential is still on the same level as a KILT credential, and I do not necessarily see them very coupled with asset DIDs. Specifically, I would see the package for asset DIDs as a separate one, which we could extend in the future by allowing for some sort of plugin-based components. So for me the question is whether we want to have public credentials in their own package, or leave them in core. I would not personally put them in the same package as asset DIDs.

File: packages/augment-api/src/interfaces/augment-api-runtime.ts [hunk 20]
```diff
@@ -117,6 +118,19 @@ declare module '@polkadot/api-base/types/calls' {
        **/
       offchainWorker: AugmentedCall<ApiType, (header: Header | { parentHash?: any; number?: any; stateRoot?: any; extrinsicsRoot?: any; digest?: any } | string | Uint8Array) => Observable<Null>>;
     };
+    /** 0xa47b7d544994c99b/1 */
+    publicCredentials: {
+      /**
+       * Return the public credential with the specified ID, if found.
+       **/
+      getCredential: AugmentedCall<ApiType, (credential_id: Hash | string | Uint8Array) => Observable<Option<PublicCredentialsCredentialsCredentialEntry>>>;
```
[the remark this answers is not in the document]
ntn-x2: We might be able to squeeze them into this release. How would you call them? `publicCredentials.getById()` and `publicCredentials.getAllForAsset()`? Open to naming suggestions 😊

packages/core/src/claim/Claim.ts
```diff
@@ -189,17 +196,23 @@ export function verifyDisclosedAttributes(
...
 
 /**
- * Checks whether the input meets all the required criteria of an [[IClaim]] object.
+ * Checks whether the input meets all the required criteria of an [[IClaim]] or [[IAssetClaim]] object.
  * Throws on invalid input.
  *
  * @param input The potentially only partial IClaim.
  */
-export function verifyDataStructure(input: IClaim | PartialClaim): void {
+export function verifyDataStructure(
```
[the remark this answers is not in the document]
ntn-x2: But we do rely on some common information such as CTypes, claims, and delegations. I don't see them THAT different, to be honest. Do you think we should repeat code that basically does the same thing for normal credentials in the public credentials?

packages/types/src/index.ts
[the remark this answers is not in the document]
ntn-x2: Ok probably here's my limited knowledge of TS hitting. We use them in the return type of the `parse` function. Does that make them required to be exported? If not, then we can only export the `AssetDid`.

arty-name:
>> Don't you think public credentials are better bundled up with asset dids than added to the core package?

> I would like @arty-name opinion about this

I think the outcome depends on what future the team foresees for these credentials. I don’t really know...

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (!('claims' in input)) {
+    throw new SDKErrors.ClaimMissingError()
+  } else {
+    verifyClaimDataStructure({
+      cTypeHash: input.cTypeHash,
+      contents: input.claims,
+      subject: input.subject,
+    })
+  }
+  if (!input.subject) {
```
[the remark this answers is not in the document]
ntn-x2: This was also taken from the credential verification function. I am not sure what the best one is, that would also work for JS code. Whatever you suggest, I will implement.

packages/core/src/claim/Claim.ts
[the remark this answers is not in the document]
rflechtner: It's kind of a red flag for me to accept two different data structures in a function called `verifyDataStructure` whose purpose is acting as a type guard of sorts. I know the types are close, but with the subject property being called differently in the two instances I would prefer to have separate guards for separate types.

packages/types/src/index.ts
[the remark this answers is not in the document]
arty-name: In that case, I’d try to move them all in the AssetDid.ts and only export `AssetDid`

packages/augment-api/src/interfaces/augment-api-runtime.ts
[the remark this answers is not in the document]
rflechtner: That's essentially what I had in mind, maybe `publicCredentials.getById()` and `publicCredentials.getBySubject()`?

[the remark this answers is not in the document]
ntn-x2: Sounds good to me! Will open a node PR for that.

packages/core/src/publicCredential/PublicCredential.ts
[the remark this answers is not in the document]
arty-name: Any of those work for JS code. TS might complain about some cases however. I believe it was the most accepting about the first option: `'something' in data`. This only checks for presence of this property and ignores the value, though. If you think that checking the value type is important here, do include the `typeof data.something ===` in the check, as in the last option.

rflechtner:
> > You were asking about ideas how public credentials could be verified without looking them up from the block of issuance. I think there could be a way where we combine credential data with attester data on chain while the extrinsic is being processed and trigger an event containing both the resulting object and its hash. If the hash becomes the credential id, then looking up the attestation should suffice if you have the combined object. Did we decide against using events with public credentials and if so why?
> 
> Isn't that what's been implemented in the `verifyCredential` function? We re-compute the ID of the provided credential and try to look it up in the block state. If we find a match, the credential exists. If we don't, the credential is wrong.

Ok yeah seems like everything but revocation status and block number is already included in the hash/id - which makes sense. It's all there then!

packages/types/src/index.ts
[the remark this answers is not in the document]
ntn-x2: Then we would have to move the `IPublicCredentialInput` type as well, because so far the `@kiltprotocol/types` package does not depend on any other package of ours. How do we deal with that?

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: What do you mean?

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: It is not, at least the compiler is complaining. `asErr` returns a `Codec` type, and it does not use `toString()` by default, AFAICT.

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2:
```
Argument of type '`0x${string}` | Option<PublicCredentialsCredentialsCredentialEntry>' is not assignable to parameter of type '`0x${string}`'.
  Type 'Option<PublicCredentialsCredentialsCredentialEntry>' is not assignable to type '`0x${string}`'.
```

ntn-x2: Apparently returning a tuple makes all elements of the tuple type unions...

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,230 @@
...
+): EncodedPublicCredential {
+  const { cTypeHash, claims, subject, delegationId } = publicCredential
+
+  const cborSerializedClaims = cborEncode(claims)
+
+  return {
+    ctypeHash: cTypeHash,
+    subject,
+    // FIXME: Using Uint8Array directly fails to encode and decode, I guess because the api object assumes the byte array is SCALE-encoded.
+    claims: u8aToHex(new Uint8Array(cborSerializedClaims)),
```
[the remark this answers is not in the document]
ntn-x2: This is what I get if I try to pass a U8a array directly:

```
createType(PublicCredentialsCredentialsCredential):: Struct: failed on claims: Bytes:: Bytes: required length less than remainder, expected at least 6442, found 107
```

My guess is that polkadot expects the type to be SCALE-encoded. So giving it as a HEX bypasses that and lets polkadot encode it in the right way. I think it's good enough. What do you think?

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: I didn’t notice that below you look at events, not calls, so it probably doesn’t matter that some calls could have been batch calls.

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: Okay, I see

packages/types/src/index.ts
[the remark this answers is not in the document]
ntn-x2: I prefixed them with the standard they refer to. I guess there's nothing more accurate than that 😄 https://github.com/KILTprotocol/sdk-js/pull/677/commits/63d4a0542d29bb702078ba1925fd24bf230e00e9#diff-e450783712ca158e57332108b7545b8eb5d7b47bf88ce541bcc60a0017097adfR11

yarn.lock
```diff
@@ -1563,6 +1573,21 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@polkadot/api-augment@npm:9.6.2":
```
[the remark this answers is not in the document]
ntn-x2: Apparently I could not get to remove duplicates unless I updated all polkadot dependencies to the latest version, which I did in https://github.com/KILTprotocol/sdk-js/pull/677/commits/4074343193c63dffcd470ccf89a89f429c915179.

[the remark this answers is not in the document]
ntn-x2: Good catch, by the way!

packages/core/src/publicCredential/PublicCredential.ts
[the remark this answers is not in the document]
ntn-x2: Done in https://github.com/KILTprotocol/sdk-js/pull/677/commits/e113cbbc643026f3dadf6e2857bff733dd69d924#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R87. I also copy-pasted and adapted the `verifyClaimsStructure` as Raphael suggested below in https://github.com/KILTprotocol/sdk-js/pull/677/commits/e113cbbc643026f3dadf6e2857bff733dd69d924#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R60.

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: `cborDecode(credential.claims.toU8a())` did not work, but apparently `cborDecode(credential.claims),` does, so I did that in https://github.com/KILTprotocol/sdk-js/pull/677/commits/63d4a0542d29bb702078ba1925fd24bf230e00e9#diff-3228260bd7a8a1c0fb1e0b94efbc0b692518737b171dda86329b57db740703bbR86

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: Very nice suggestion! I guess I tried to make it work since understanding how to parse a block has been quite frustrating. Then I simply forgot to make the code clean and efficient. I fixed all of that in https://github.com/KILTprotocol/sdk-js/pull/677/commits/e113cbbc643026f3dadf6e2857bff733dd69d924#diff-3228260bd7a8a1c0fb1e0b94efbc0b692518737b171dda86329b57db740703bbR100.

I do a `.reverse()` and a `.find()` in two different places where I just need to access the last occurrence of an element in the array. Love this suggestion!

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/677/commits/e113cbbc643026f3dadf6e2857bff733dd69d924#diff-3228260bd7a8a1c0fb1e0b94efbc0b692518737b171dda86329b57db740703bbR100

Best suggestion of this PR, nothing against the other ones tho 😄

packages/augment-api/src/interfaces/augment-api-runtime.ts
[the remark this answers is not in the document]
ntn-x2: PR has been merged: https://github.com/KILTprotocol/kilt-node/pull/433

Will update the integration tests once a new `latest-develop` image is pushed. Good suggestion, completely forgot about the naming.

packages/core/src/claim/Claim.ts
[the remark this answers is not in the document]
ntn-x2: I did it in https://github.com/KILTprotocol/sdk-js/pull/677/commits/e113cbbc643026f3dadf6e2857bff733dd69d924#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R60.

packages/core/src/publicCredential/PublicCredential.chain.ts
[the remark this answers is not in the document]
ntn-x2: We can somehow decode ok with `cborDecode(credential.claims)`, but encoding does not work, neither with `cborEncode(credential.claims` if we make `claims` a `Buffer`, nor by passing it as a byte array. It only works if we pass it as a HEX string.

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -133,22 +132,31 @@ export async function verifyCredential(
...
+  if (cType) {
+    verifyAgainstCType(credential, cType)
   }
-  if (blockNumber !== credential.blockNumber) {
+
+  const {
+    blockNumber: retrievedBlockNumber,
+    revoked: retrievedRevocationStatus,
+  } = encodedCredentialEntry.unwrap()
+  if (blockNumber !== retrievedBlockNumber.toBn()) {
```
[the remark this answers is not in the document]
ntn-x2: Didn't yet run all unit tests, but done in https://github.com/KILTprotocol/sdk-js/pull/677/commits/63d4a0542d29bb702078ba1925fd24bf230e00e9#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R166.

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: I see… With that knowledge, it looks like creating an intermediate data structure harms more than it helps. Would it become better when calling `credentialFromChain` directly in the same loop?

packages/asset-did/src/AssetDid.ts
```diff
@@ -0,0 +1,81 @@
...
+  return {
+    chainId,
+    chainNamespace,
+    chainReference,
+    assetId,
+    assetNamespace,
+    assetReference,
+    assetInstance,
+    uri: `did:asset:${chainId}.${assetId}`,
+  }
```
arty-name, line 66: Now that I see it, I think this can be even simpler
```suggestion
  return {
    ...(matches as Omit<IAssetDidParsingResult, 'uri'>),
    uri: `did:asset:${chainId}.${assetId}`,
  }
```

packages/core/src/claim/Claim.ts
```diff
@@ -189,7 +189,7 @@ export function verifyDisclosedAttributes(
 }
 
 /**
- * Checks whether the input meets all the required criteria of an [[IClaim]] object.
+ * Checks whether the input meets all the required criteria of an [[IClaim]] or [[IAssetClaim]] object.
```
arty-name, line 192: Guess this comment needs to be reverted as well

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: Looks good now!

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: Nice!

yarn.lock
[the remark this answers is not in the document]
arty-name: The only reason for me to look at the diffs of yarn.lock :)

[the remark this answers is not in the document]
arty-name: By the way, babel duplicates in the new diff seem innocent, however it would be worth it to make sure that the bundle size doesn’t change significantly with new dependencies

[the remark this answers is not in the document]
ntn-x2: Is there a fix for it? Neither `yarn --dedupe` not `npx yarn-dedupe` work.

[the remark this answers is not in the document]
arty-name: Let’s bother ourselves with that only if we have to :)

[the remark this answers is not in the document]
rflechtner: it would be `yarn dedupe` but if that doesn't work, you can also remove anything babel related from the lock file and run `yarn install`

packages/asset-did/src/AssetDid.ts
ntn-x2: We can't use `chainId` and `assetId` in this way 😁

ntn-x2: I only captured `chanId` and `assetId` and then used your suggestion here. Will push soon.

packages/core/src/publicCredential/PublicCredential.chain.ts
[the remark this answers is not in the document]
rflechtner: that's the most plausible explanation, yeah... guess it expects the first byte(s) to be a length indicator then.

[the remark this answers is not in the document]
ntn-x2: Correct. I would go as far as removing the FIXME, since I can't think of another way of making this work.

yarn.lock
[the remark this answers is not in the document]
ntn-x2: Ok `yarn dedupe` this time seems to have removed some duplicates -> https://github.com/KILTprotocol/sdk-js/pull/677/commits/be41879ba04a7c63603df394a1b1653d0d938a1f

packages/asset-did/src/AssetDid.ts
ntn-x2: Something like this: https://github.com/KILTprotocol/sdk-js/pull/677/commits/589c8bc7d8f0fbdf240c2824a70adb34102dbfb0#diff-78f6deb4434716917618adc7d5660fc049921f43f0f6e539ea4702316353c594R47

packages/core/src/claim/Claim.ts
ntn-x2: Nice! https://github.com/KILTprotocol/sdk-js/pull/677/commits/589c8bc7d8f0fbdf240c2824a70adb34102dbfb0#diff-0e2cdf2fffcea87f7c76a48effc1b12d7758d3f545a5975fba1c0c71440fbee8R192

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: I personally don't think this is any clearer, but I followed your suggestion regardless: https://github.com/KILTprotocol/sdk-js/pull/677/commits/589c8bc7d8f0fbdf240c2824a70adb34102dbfb0#diff-3228260bd7a8a1c0fb1e0b94efbc0b692518737b171dda86329b57db740703bbR208.

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -80,11 +80,11 @@ function verifyClaimsStructure(input: IAssetClaim | PartialAssetClaim): void {
 
 // Used internally only when building the [[IPublicCredentialInput]].
 function verifyDataStructure(input: IPublicCredentialInput): void {
-  // Taken from `Credential.verifyDataStructure()`
-  if (!('claims' in input)) {
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
arty-name, line 83: Right, I guess I remembered this guy when saying that TS is more accepting of the form `'xxx' in yyy

ntn-x2: It was returning `true` even when I did not pass it in the object. I tested that in the unit tests, and I do not know why. And also the warning is only shown for the `claims` property, somehow.

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: I think it has less noise this way :+1:

packages/augment-api/src/interfaces/augment-api-runtime.ts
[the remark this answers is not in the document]
ntn-x2: Done!

packages/core/src/publicCredential/PublicCredential.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/677/commits/f503afa1216bf43518598267e499a0f61f49019d

arty-name:
I just recalled, there’s a new guy on the block: `Array.prototype.findLast()`. It’s too early to use it, though, but maybe in a year.

ntn-x2:
@arty-name @rflechtner I consider this PR ready to be merged. There is something seemingly broken with CType verification, which will be tackled in a different PR. For now, the test for that check is skipped, and we can enable it again once we have a fix for the CTypes.
Please re-review the last changes, otherwise I would say we can proceed and merge, so that other people can start playing around with it. I won't be available to do any more major work after today, so I hope that anything left comes up today so I can address it.

packages/core/src/publicCredential/PublicCredential.chain.ts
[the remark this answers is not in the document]
ntn-x2: Done in https://github.com/KILTprotocol/sdk-js/pull/677/commits/1b5f46728cf5083e56e9f4e689a796a050586db6.

packages/asset-did/src/AssetDid.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @param assetDidUri An AssetDID uri as a string.
+* @returns Object containing information extracted from the AssetDID uri.
+ */
+export function parse(assetDidUri: AssetDidUri): IAssetDidParsingResult {
+  const matches = ASSET_DID_REGEX.exec(assetDidUri)?.groups
+  if (!matches) {
+    throw new SDKErrors.InvalidDidFormatError(assetDidUri)
+  }
+
+  const { chainId, assetId } = matches as Omit<IAssetDidParsingResult, 'did'>
```
arty-name, line 47: I know it was me who mistyped it, but it should be `uri`, not `did` :)

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -0,0 +1,218 @@
...
+function credentialInputFromChain(
+  credential: PublicCredentialsCredentialsCredential
+): IPublicCredentialInput {
+  validateUri(credential.subject.toUtf8())
+  return {
+    claims: cborDecode(credential.claims),
+    cTypeHash: credential.ctypeHash.toHex(),
+    delegationId: credential.authorization.unwrapOr(undefined)?.toHex() ?? null,
+    subject: credential.subject.toUtf8() as AssetDidUri,
+  }
```
arty-name, line 90: I think destructuring the properties from credential will make this code more readable

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * Calculates the ID of a [[IPublicCredentialInput]], to be used to retrieve the full content from the blockchain.
+ *
+ * The ID is formed by first concatenating the SCALE-encoded [[IPublicCredentialInput]] with the SCALE-encoded [[DidUri]] and then Blake2b hashing the result.
+ *
+ * @param credential The input credential object.
+ * @param attester The DID of the credential attester.
+ * @returns The credential ID.
+ */
+export function computeId(
```
arty-name, line 39: I don’t see many `compute*` function in the SDK, would the consistent name be `getIdFromCredential()`?

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -0,0 +1,206 @@
...
+  const scaleEncodedAttester = api
+    .createType<AccountId>('AccountId', Did.toChain(attester))
+    .toU8a()
+
+  return blake2AsHex(
+    Uint8Array.from([...scaleEncodedCredential, ...scaleEncodedAttester])
+  )
+}
+
+function verifyClaimsStructure(input: IAssetClaim | PartialAssetClaim): void {
```
arty-name, line 60: Should it remain "verify claimS structure"?

arty-name (approved):

packages/core/src/publicCredential/PublicCredential.ts
ntn-x2: But `get` sounds like it's in there and we retrieve it. What we do is we re-calculate it based on the other properties of the credentials. Compute might not be the best, but I don't think get is the right word either here.

arty-name: We already have `getHashForSchema()` which does the same. I did misremembered `for` and typed `from`, so my updated suggestions is `getIdForCredential()`

ntn-x2: Good point. I changed it then, but I still don't like the naming 😄

packages/asset-did/src/AssetDid.ts
ntn-x2: Ops... https://github.com/KILTprotocol/sdk-js/pull/677/commits/023ad1c369410f4be90a93e0c4363bf2b8759885#diff-78f6deb4434716917618adc7d5660fc049921f43f0f6e539ea4702316353c594R47

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: Like this? https://github.com/KILTprotocol/sdk-js/pull/677/commits/023ad1c369410f4be90a93e0c4363bf2b8759885#diff-3228260bd7a8a1c0fb1e0b94efbc0b692518737b171dda86329b57db740703bbR81

packages/core/src/publicCredential/PublicCredential.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/677/commits/023ad1c369410f4be90a93e0c4363bf2b8759885#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R39

packages/core/src/publicCredential/PublicCredential.ts
ntn-x2: Yeah makes sense: https://github.com/KILTprotocol/sdk-js/pull/677/commits/023ad1c369410f4be90a93e0c4363bf2b8759885#diff-0ed53062bda598489034fe7ca13c690382a38738d8e6ea23ebfc8b3a6a3b5880R60

packages/asset-did/src/AssetDid.ts
arty-name: And the line below too ;)

packages/core/src/publicCredential/PublicCredential.chain.ts
arty-name: Yes, and `subject.toUtf8()` is done twice

packages/core/src/publicCredential/PublicCredential.ts
arty-name: I also had a long discussion with Timo on a similar topic :) Consistency is important though.

arty-name:
Looking at the updated error message in the latest commit, I started thinking that the ID of the public credential is its hash of a kind. If we use this more specific word "hash" (as we do for CType), we won’t have to explain to other developers what are the IDs of public credentials, and how we generate them, and what a mismatching ID means. Because many developers already know these facts about hashes.

packages/core/src/publicCredential/PublicCredential.chain.ts
ntn-x2: Good point: https://github.com/KILTprotocol/sdk-js/pull/677/commits/31a33366db67f7deb50aecba5e2c9f61c0df3c5e

packages/asset-did/src/AssetDid.ts
ntn-x2: Ops 😄😄😄....... https://github.com/KILTprotocol/sdk-js/pull/677/commits/31a33366db67f7deb50aecba5e2c9f61c0df3c5e

packages/core/src/publicCredential/PublicCredential.ts
ntn-x2: I will resolve this conversation now, as we seem to have reached mutual understanding and an agreement.

rflechtner (approved):
There's a suggestion or two but I think we're done here!

packages/testing/src/mocks/mockedApi.ts
```diff
@@ -375,13 +379,18 @@ export function getMockedApi(): MockApiPromise {
...
+        credentialSubjects: jest.fn(),
+      },
     },
     runtimeMetadata: {
       asV11: {
         modules: [],
       },
     },
     registry: TYPE_REGISTRY,
+    createType: TYPE_REGISTRY.createType,
```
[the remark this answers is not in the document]
ntn-x2: I just removed it since we don't need it anymore anyway: https://github.com/KILTprotocol/sdk-js/pull/677/commits/9ebc427e6823cf30c0ea59ee4e8d6b59b1f50d7e

packages/core/src/publicCredential/PublicCredential.ts
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @param credential A [[IPublicCredential]] for the attester.
+ * @param cType A [[CType]] to verify the [[Claim]] structure.
+ */
+export function verifyAgainstCType(
+  credential: IPublicCredential,
+  cType: ICType
+): void {
+  verifyClaimAgainstSchema(credential.claims, cType)
+}
```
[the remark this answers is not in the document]
ntn-x2: It was to have consistency with normal credentials. Do you think we should remove them?

packages/core/src/claim/Claim.ts
```diff
@@ -198,7 +198,8 @@ export function verifyDataStructure(input: IClaim | PartialClaim): void {
   if (!input.cTypeHash) {
     throw new SDKErrors.CTypeHashMissingError()
   }
-  if (input.owner) {
+  if ('owner' in input) {
+    // input is IClaim
```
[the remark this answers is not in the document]
ntn-x2: I think we discussed with Tom we wanted to have the same logic that we have in the public credentials now? Do you think we should revert?

PR #678, opened 2022-11-04
chore: update develop metadata and migrate to new DID runtime API in integration tests

ntn-x2:
https://github.com/KILTprotocol/kilt-node/pull/425 entirely removed the old `api.call.didApi` syntax, hence integration tests against develop were failing. Since the `api.call.did` syntax is present on both `master` and `develop`, this PR should be enough.


arty-name (approved):
Nice code! Also the first time in a long time I see all the checks to pass!

PR #681, opened 2022-11-11
fix: add peregrine stg chain to type decorations

wischli:
## no ticket
* adds Peregrine Staging Chain name to type decorations

## Checklist:


rflechtner (approved):
Makes sense

PR #688, opened 2022-11-23
namespaces for type augmentation

rflechtner:
## fixes KILTProtocol/ticket#2224

Originally I was trying to fix type errors coming from types that are defined e.g. in `@polkadot/api-base` or `@polkadot/api-derive` and prevent us from disabling `skipLibCheck`. I made no real progress there but while working on it found moonbeams solution to augmenting their api which at least provides a nice template for how we can use augmentation with multiple chains.
Putting this here as a WIP to be picked up later. @ntn-x2 I thought you'd probably be interested.

## Checklist:


ntn-x2:
Cool! I also saw how moonbeam does it, and tried to replicate it, but I was somehow having a duplicate definition error, since two different packages where trying to declare the `@polkadot/types` namespace. If that is fixed in this PR (which I haven't yet checked, will do some time this week), then this definitely represents a step in the right direction.

rflechtner:
> Cool! I also saw how moonbeam does it, and tried to replicate it, but I was somehow having a duplicate definition error, since two different packages where trying to declare the `@polkadot/types` namespace. If that is fixed in this PR (which I haven't yet checked, will do some time this week), then this definitely represents a step in the right direction.

I don't think it's completely fixed unfortunately, at least it does through related errors when you disable skipLibCheck if I remember correctly. This still needs a bit of work, but I think some issues might be resolved by a separate line of typescript compilation improvements I was pushing forward.

ntn-x2:
@rflechtner I pushed a commit to make sure we don't default to spiritnet when importing `@kiltprotocol/augment-api`. It's my personal opinion that we should be explicit about this. Anyway, I don't think I have made much progress, since:

* The project compiles with `yarn build`, but fails to be checked with `yarn check`
* As soon as the peregrine namespace is imported in some files (e.g., public credentials), apparently it's not possible to remove those declarations even by manually importing `@kiltprotocol/augment-api/spiritnet` in a test script that simply imports the SDK

Dudleyneedham:
@rflechtner Can we close this now?

rflechtner:
Closing as superseded by https://github.com/KILTprotocol/types-augment

PR #689, opened 2022-11-24
refactor: reduce startup time of tests by using babel to transpile js

rflechtner:
Unit- and integration tests would sometimes get stuck in the startup phase for minutes, as they transpile polkadot-js dependencies that unfortunately are not always 100% clean cjs code - somehow esm export statements seem to sneak into the dependency tree, leading to all tests failing.

Using babel to transpile js files instead of tsc (which only transpiles ts now) appears to buy us significant performance gains, with startup times being reduced to 15-20 seconds on my system.

This comes at the cost of a slightly larger development install, as we now need more babel dependencies. 

## How to test:

Build sdk and run tests on your machine, report how that goes.

## Checklist:

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,10 @@
+{
+    "presets": [
+        [
+            "@babel/preset-env",
+            {
+                "modules": "commonjs"
+            }
+        ]
+    ]
+}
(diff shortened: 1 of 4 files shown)


ntn-x2 (approved):
FWIW, looks good to me!

PR #690, opened 2022-11-24
fix: 0.30.0-rc.1 issues

rflechtner:
fixing outdated README examples and fixing the export of types from the augment-api package.

## Checklist:


weichweich (commented):
Could we move this to docs.kilt.io and simply link to there from the readme? We could in that case also test the code examples. 😁

rflechtner:
That’s a good idea, but maybe not for a hotfix on the rc? Let’s do it afterwards

weichweich (approved):

PR #697, opened 2022-12-15
fix: block parsing logic for public credentials

ntn-x2:
Previous logic was assuming that the only extrinsic that could create a public credential was a `submit_did_call` extrinsic. This is not true, as `batch`, `batchAll`, and `forceBatch` can also create a public credential.

This is now taken into account in this PR, and an integration test that covers the new case has been added.

One problematic block, raised by BTE, was https://kilt-testnet.subscan.io/block/2128207.


rflechtner:
I'd have to take more time to give qualified feedback, but looks reasonable on first glance.
I have to say though it still seems _wild_ to me that we do this instead of, say, firing events with the credential in them. I did expect issues like this to come up when I first heard of it, and here we are 🥹
Are we sure this is the best way of doing this?

ntn-x2:
@rflechtner I understand and I agree it would make things easier, but remember that the credential is user-provided input. Yes, it is bound in the maximum size it can have, but the general feeling also on the forums is that events should be kept small and only used as index to something else in the state (e.g., https://substrate.stackexchange.com/questions/4296/large-event-payloads-unsafe).
In our case, we don't store the credential input in the state, hence this approach does not apply to us. Furthermore, events are written on storage and cleaned on the next block, so that means that larger event payload would have an even bigger impact on the weight of the extrinsic. I know this is probably not the cleanest way to access the information, but I think it's the best tradeoff. Perhaps, once back, @weichweich might want to chip in as well, since we discussed also in the past how we can use the block number as a pointer and it never came up to put the whole payload in an event body.

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -147,31 +168,40 @@ export async function credentialFromChain(
...
+
+  // If more than one call is present, it always considers the last one as the valid one, and takes its submitter.
   const lastRightCredentialCreationCall = callCredentialsContent
     .reverse()
-    .find((credentialInput) => {
-      const reconstructedId = getIdForCredential(
-        credentialInput,
-        extrinsicDidOrigin
-      )
+    .find(([c, s]) => {
```
arty-name, line 203: Do you think the multi-letter variables were more readable?

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -147,31 +168,40 @@ export async function credentialFromChain(
...
+    ? extrinsic.args[0].flatMap((batchCall) =>
+        extractDidCallsFromBatchCall(api, batchCall)
+      )
+    : [extrinsic]
 
-  const credentialCreationCalls = extrinsicCalls.filter(
-    (call): call is GenericCall<typeof api.tx.publicCredentials.add.args> =>
-      api.tx.publicCredentials.add.is(call)
-  )
+  // From the list of DID calls, only consider public_credentials::add calls, bundling each of them with their DID submitter.
```
arty-name, line 185: I was frivolous with comments during refactoring, please confirm that they still fit

packages/core/src/publicCredential/PublicCredential.chain.ts
```diff
@@ -180,10 +210,13 @@ export async function credentialFromChain(
       'Block should always contain the full credential, eventually.'
     )
   }
+
+  const [credentialInput, submitter] = lastRightCredentialCreationCall
```
arty-name, line 214: Would it be nice to have one mapping less in the object below?
```suggestion
  const [credentialInput, attester] = lastRightCredentialCreationCall
```

arty-name (approved):

ntn-x2:
@rflechtner since this is blocking BTE, I've decided to go on and merge it after getting a review from Tom. Anything else you wanted to discuss about this, we can do once we're both back, and address any further open points.

PR #700, opened 2023-01-09
fix: esm imports

rflechtner:
## fixes KILTProtocol/ticket#2342

Fixes module issues occurring when using the ES6 modules build of the sdk:
- ESM doesn't support json imports
- Import of jsonld-signatures, which is CJS-only, needs to follow special rules (named exports are not supported)

## How to test:

Use build in an ES6 modules project.

## Checklist:


ntn-x2 (approved):
👍

PR #702, opened 2023-01-24
feat: verification helpers

rflechtner:
## fixes KILTProtocol/ticket#xxx

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:

### DIFF
--- a/packages/core/src/attestation/Attestation.ts
+++ b/packages/core/src/attestation/Attestation.ts
@@ -138,7 +138,8 @@ export function verifyAgainstCredential(
 ): void {
   if (
     credential.claim.cTypeHash !== attestation.cTypeHash ||
-    credential.rootHash !== attestation.claimHash
+    credential.rootHash !== attestation.claimHash ||
+    credential.delegationId !== attestation.delegationId
   ) {
     throw new SDKErrors.CredentialUnverifiableError(
       'Attestation does not match credential'
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -17,6 +17,7 @@
  * @packageDocumentation
  */
 
+import { ConfigService } from '@kiltprotocol/config'
 import {
   isDidSignature,
   verifyDidSignature,
@@ -26,6 +27,7 @@ import {
 } from '@kiltprotocol/did'
 import type {
   DidResolveKey,
+  DidUri,
   Hash,
   IAttestation,
   IClaim,
@@ -36,9 +38,11 @@ import type {
   SignCallback,
 } from '@kiltprotocol/types'
 import { Crypto, DataUtils, SDKErrors } from '@kiltprotocol/utils'
+import * as Attestation from '../attestation/index.js'
 import * as Claim from '../claim/index.js'
 import { hashClaimContents } from '../claim/index.js'
 import { verifyClaimAgainstSchema } from '../ctype/index.js'
+import { DelegationNode } from '../delegation/DelegationNode.js'
 
 function getHashRoot(leaves: Uint8Array[]): Uint8Array {
   const result = Crypto.u8aConcat(...leaves)
@@ -293,6 +297,8 @@ type VerifyOptions = {
 
 /**
  * Verifies data structure & data integrity of a credential object.
+ * THIS DOES NOT VERIFY THAT A CREDENTIAL HAS BEEN ATTESTED, OR BY WHOM!
+ * For this, you need to call [[verifyAttestation]].
  *
  * @param credential - The object to check.
  * @param options - Additional parameter for more verification steps.
@@ -312,8 +318,9 @@ export async function verifyCredential(
 
 /**
  * Verifies data structure, data integrity and the claimer's signature of a credential presentation.
+ * THIS DOES NOT VERIFY THAT A CREDENTIAL HAS BEEN ATTESTED, OR BY WHOM!
  *
- * Upon presentation of a credential, a verifier would call this function.
+ * Upon presentation of a credential, a verifier would call this function, then call [[verifyAttestation]] to check the credential's attestation status.
  *
  * @param presentation - The object to check.
  * @param options - Additional parameter for more verification steps.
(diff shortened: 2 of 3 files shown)


packages/core/src/attestation/Attestation.ts
```diff
@@ -138,7 +138,8 @@ export function verifyAgainstCredential(
 ): void {
   if (
     credential.claim.cTypeHash !== attestation.cTypeHash ||
-    credential.rootHash !== attestation.claimHash
+    credential.rootHash !== attestation.claimHash ||
```
weichweich, line 141: I think it would be nice to report which entry didn't match. A dev using the SDK and debugging a potential issue might appreciate that information.

packages/core/src/credential/Credential.ts
```diff
@@ -332,6 +339,43 @@ export async function verifyPresentation(
...
+    credential.claim.cTypeHash !== attestation.cTypeHash ||
+    credential.delegationId !== attestation.delegationId
+  ) {
+    throw new SDKErrors.CredentialUnverifiableError(
+      'Attestation does not match credential'
+    )
+  }
+  if (attestation.revoked)
+    throw new SDKErrors.RevokedTypeError('Attestation revoked')
+  if (trustedAttesters.includes(attestation.owner)) return
```
weichweich, line 370: this should be an else. If we don't trust the attester, we would check if we trust the delegation root

packages/core/src/credential/Credential.ts
```diff
@@ -332,6 +339,43 @@ export async function verifyPresentation(
...
+  const attestation = Attestation.fromChain(maybeAttestation, rootHash)
+  if (
+    credential.claim.cTypeHash !== attestation.cTypeHash ||
+    credential.delegationId !== attestation.delegationId
+  ) {
+    throw new SDKErrors.CredentialUnverifiableError(
+      'Attestation does not match credential'
+    )
+  }
+  if (attestation.revoked)
```
weichweich, line 368: BTW I always prefer to add `{}`. This reduces the chance of potential issues where you would want to have something in the if but it's only indented and not really inside. There was a huge security issue because of this once (i think it was in some apple product). They also didn't use `{}` for ifs and because of a merge error a `goto verified` was added twice. The second one was always executed even though the check wasn't valid. -> security issue. 😨

packages/core/src/credential/Credential.ts
rflechtner: this is what's happening. no need for an else; it's an early return (return means success)

packages/core/src/credential/Credential.ts
rflechtner: uuuu damn that's nasty. will be adding {} everywhere

packages/core/src/credential/Credential.ts
weichweich: Ah yes. Should have studied the code in more depth. Sorry.

packages/core/src/attestation/Attestation.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/702/commits/818f32c03eb0445a0b8e37efa7bcd12d80148694

packages/core/src/credential/Credential.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/702/commits/0b7a7e468f206cda0720c5eebbf8cc26e0b8b78a

rflechtner:
I talked to Tom to get his feedback on this feature as both someone involved in the refactor and the primary user of the sdk rn. His stance on it in a nutshell:
> Ease-of-use was not the reason to refactor the SDK. We wanted to reduce its size for maintainability by a minimal team, and to expose external developers to polkadot APIs for skill reusability. If convenience becomes important again, we could really revert the whole refactoring.
> I’d say we should introduce functions only when it’s unreasonable to expect from the external developer that they understand the intricacies of something. Checking that attester is trusted is a no-brainer."

Personally, I see things similarly. I get that the process of fetching and matching an attestation to a credential should be improved, but once you have the attestation object, it's really not to much to ask to have people access and process its properties according to what they are trying to do. E.g. just displaying the credential and whether it's valid, or using the attester prop and match it to a whitelist, or for advanced use cases use the delegation id prop too (this is two function calls now, `Delegation.fetch` & `.findAncestorOwnedBy()`, but could be streamlined if necessary).

What do you think of improving on `Attestation.fromChain()` instead so that it matches attestation data to the credential? This way credential verification is 4 steps:
1. Verify credential data with `verifyCredential` or `verifyPresentation`
2. Query attestation
3. throw queried, scale encoded attestation option and credential into `Attestation.fromChain()`
4. Use attestation data according to your use case (fail if revoked or attester unknown, simply show data, whatever).

@Dudleyneedham do you want to weigh in here as well with your learnings from workshops and building demos?

ntn-x2:
I am of a similar opinion as well, but I think having a higher level function which wraps of the low-level building blocks, which we are still providing, would be good to have.
We have been tricked ourselves everywhere in the docs, where we were not checking the attestation on chain, which I would argue is at the core of what verifying a credential, our credential, means. Verifying against a trusted attester list, we always offloaded that to the app developer, so at least that was a conscious we took.

In this very specific case, maybe a common ground would be to provide a flag that would indicate whether to check also the on-chain attestation? So if users want to implement that themselves, they would pass `false` and would not query the thing twice. This is coming just from the top of my head, it might not be a good suggestion. But if I had to choose, at least for the `verifyCredential`, I would include all the required checks in there, and also provide the building blocks for people to do that themselves.

rflechtner:
> I am of a similar opinion as well, but I think having a higher level function which wraps of the low-level building blocks, which we are still providing, would be good to have. We have been tricked ourselves everywhere in the docs, where we were not checking the attestation on chain, which I would argue is at the core of what verifying a credential, our credential, means. Verifying against a trusted attester list, we always offloaded that to the app developer, so at least that was a conscious we took.
> 
> In this very specific case, maybe a common ground would be to provide a flag that would indicate whether to check also the on-chain attestation? So if users want to implement that themselves, they would pass `false` and would not query the thing twice. This is coming just from the top of my head, it might not be a good suggestion. But if I had to choose, at least for the `verifyCredential`, I would include all the required checks in there, and also provide the building blocks for people to do that themselves.

Without changing the function's return type, which is quite a drastic change I would say, there is no real middle ground between making `verifyCredential` a function that does (or can do) everything from revocation checks to matching the attester and walking through the delegation tree, and the polar opposite of not checking the attestation at all. If we query the attestation internally and we're not returning the data within from the function, we need to make sure that all possible checks on its contents can be performed. And I follow Tom here that building a function for any conceivable combination of these checks just does not make sense if these checks themselves a fairly trivial.

The more I think about it, the more I realise that I don't like the idea of changing how `verifyCredential` works because I see it precisely as a building block to be used in credential verification logic; my biggest gripe with it is its name, which suggests otherwise. But I'd rather rename it than introduce under the hood queries which I think just don't belong there.

rflechtner:
I played with api derives a bit yesterday and found that the following signature is possible:
```js
const {attester, revoked} = await api.derive.credential.verifyAttested( credential )
```
where this fetches the attestation and checks it against the credential as well, rejecting if there's an issue. The nice thing is that subscriptions are possible but tbh other than that it doesn't differ much from a function-based approach that calls api.query internally.

rflechtner:
A compromise would be to not have verifyCredential implement every possible combination of assertions on the credential issuers (and delegations), but still query the attestation under the hood. I opened a PR targeting this branch (https://github.com/KILTprotocol/sdk-js/pull/708) to show what that could look like.

rflechtner:
> I played with api derives a bit yesterday and found that the following signature is possible:
> 
> ```js
> const {attester, revoked} = await api.derive.credential.verifyAttested( credential )
> ```
> 
> where this fetches the attestation and checks it against the credential as well, rejecting if there's an issue. The nice thing is that subscriptions are possible but tbh other than that it doesn't differ much from a function-based approach that calls api.query internally.

And here an example of the derives approach: https://github.com/KILTprotocol/sdk-js/pull/709

rflechtner:
closing in favour of https://github.com/KILTprotocol/sdk-js/pull/708

PR #703, opened 2023-01-25
feat: backwards-compatible ctype data decoding

rflechtner:
Makes sure the sdk still decodes CType records from before the chain upgrade to 0.9.0 / 0.8.5.
Adjust tests so that CType fetching is skipped where not available.

## How to test:

We should no longer see the `Invalid empty address passed` errors in integration tests against `latest` (e.g., https://github.com/KILTprotocol/sdk-js/actions/runs/4003744234/jobs/6872161008).
~Fetching CType data will still fail with `Cannot fetch CType definitions on a chain that does not store the createdAt block`.~

## Checklist:

### DIFF
--- a/packages/core/src/__integrationtests__/Ctypes.spec.ts
+++ b/packages/core/src/__integrationtests__/Ctypes.spec.ts
@@ -23,8 +23,12 @@ import { disconnect } from '../kilt'
 import { createEndowedTestAccount, initializeApi, submitTx } from './utils'
 
 let api: ApiPromise
+// we skip fetching CTypes from chain for the old pallet version, where the necessary information was not yet on chain.
+let hasBlockNumbers: boolean
 beforeAll(async () => {
   api = await initializeApi()
+  // @ts-ignore not decorated for some reason
+  hasBlockNumbers = typeof api.tx.ctype.setBlockNumber !== 'undefined'
 }, 30_000)
 
 describe('When there is an CtypeCreator and a verifier', () => {
@@ -56,7 +60,9 @@ describe('When there is an CtypeCreator and a verifier', () => {
     )
     await expect(submitTx(authorizedStoreTx, keypair)).rejects.toThrowError()
     await expect(CType.verifyStored(ctype)).rejects.toThrow()
-    await expect(CType.fetchFromChain(ctype.$id)).rejects.toThrow()
+    if (hasBlockNumbers) {
+      await expect(CType.fetchFromChain(ctype.$id)).rejects.toThrow()
+    }
   }, 20_000)
 
   it('should be possible to create a claim type', async () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -23,8 +23,17 @@ import { disconnect } from '../kilt'
 import { createEndowedTestAccount, initializeApi, submitTx } from './utils'
 
 let api: ApiPromise
+// we skip fetching CTypes from chain for the old pallet version, where the necessary information was not yet on chain.
+let hasBlockNumbers: boolean
 beforeAll(async () => {
   api = await initializeApi()
+  // this extrinsic should only exist in the new pallet version
```
ntn-x2, line 30: So you're not using the storage version value anymore?

packages/core/src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -23,8 +23,17 @@ import { disconnect } from '../kilt'
...
+  hasBlockNumbers = typeof api.tx.ctype.setBlockNumber !== 'undefined'
+  if (!hasBlockNumbers) {
+    console.warn(
+      'detected pallet version 1, skipping CType fetching which is not yet supported'
```
ntn-x2, line 34: Version could be either 0 or 1 here, since 2 is the first valid one containing block number.

packages/core/src/__integrationtests__/Ctypes.spec.ts
```diff
@@ -23,8 +23,17 @@ import { disconnect } from '../kilt'
...
 
 let api: ApiPromise
+// we skip fetching CTypes from chain for the old pallet version, where the necessary information was not yet on chain.
+let hasBlockNumbers: boolean
 beforeAll(async () => {
   api = await initializeApi()
+  // this extrinsic should only exist in the new pallet version
+  hasBlockNumbers = typeof api.tx.ctype.setBlockNumber !== 'undefined'
+  if (!hasBlockNumbers) {
+    console.warn(
```
ntn-x2, line 33: This `console.warn` triggers a lint warning. Is there a workaround? Maybe simply skipping it?

ntn-x2 (approved):
LGTM! The only thing that's not 100% clear is why we don't use the storage version number, that we decided to be 2 for the valid storage structure.

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: I'm happy to use it once it's fixed!

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: or a linter exception :D

ntn-x2: Yeah I meant skipping the warning with a linter exception 😁

packages/core/src/__integrationtests__/Ctypes.spec.ts
ntn-x2: Ok, then as a placeholder, this is the PR that bumps the CType storage version to 2 for the new type: https://github.com/KILTprotocol/kilt-node/pull/458. Once it's merged into develop and an image is built, it can be used.

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: but 0 is an invalid version number technically, if I got that right?

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: Will put this on hold then until https://github.com/KILTprotocol/kilt-node/pull/458 is merged.

ntn-x2: It has been. A new `latest-develop` tag is out.

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: actually that's something we can disable for all tests I think (https://github.com/KILTprotocol/sdk-js/pull/703/commits/ba18a01b1c480533470839794d5e6e53f0c7a000)

packages/core/src/__integrationtests__/Ctypes.spec.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/703/commits/6fd18eb4f7f104cb4b21dda2ca42eaf0304c17e0

PR #705, opened 2023-01-31
feat: allow passing additional arguments to ApiPromise.create

rflechtner:
## fixes KILTProtocol/ticket#2385

Allows passing additional arguments for ApiPromise.create in `connect()` via a second parameter.
Furthermore, sets the default value of one of them (`noInitWarn`) based on the current log level.

## How to test:

Calling `connect` should produce more log output concerning types and augmentation if the log level had been set to 3 and below compared to setting it to 4 or 5.

## Checklist:

### DIFF
--- a/packages/core/src/kilt/Kilt.ts
+++ b/packages/core/src/kilt/Kilt.ts
@@ -37,7 +37,8 @@ export async function init<K extends Partial<ConfigService.configOpts>>(
  *
  * @param blockchainRpcWsUrl WebSocket URL of the RPC endpoint exposed by a node that is part of the Kilt blockchain network you wish to connect to.
  * @param apiOpts Additional parameters to be passed to ApiPromise.create().
- * @param apiOpts.noInitWarn Allows suppressing warnings related to runtime types and augmentation. Enabled by default if the global log level is 'error' or higher.
+ * @param apiOpts.noInitWarn Allows suppressing warnings related to runtime types and augmentation.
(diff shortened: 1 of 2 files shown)


packages/core/src/kilt/Kilt.ts
```diff
@@ -35,13 +36,23 @@ export async function init<K extends Partial<ConfigService.configOpts>>(
  * Connects to the KILT Blockchain and passes the initialized api instance to `init()`, making it available for functions in the sdk.
  *
  * @param blockchainRpcWsUrl WebSocket URL of the RPC endpoint exposed by a node that is part of the Kilt blockchain network you wish to connect to.
+ * @param apiOpts Additional parameters to be passed to ApiPromise.create().
+ * @param apiOpts.noInitWarn Allows suppressing warnings related to runtime types and augmentation. Enabled by default if the global log level is 'error' or higher.
```
arty-name, line 40: By synchronizing the levels I meant that these *warnings* should show when the log level in the SDK is set to *warn*. But I don’t really mind the current implementation.

arty-name (approved):

packages/core/src/kilt/Kilt.ts
rflechtner: that's what's happening - error level 3 is warn, anything above disables it.

arty-name: Thank you for the improved documentation!

PR #708, opened 2023-02-01
refactor!: fetch attestation as part of verifyCredential

rflechtner:
### also fixes https://github.com/KILTprotocol/ticket/issues/2305

This changes the function signature of `verifyCredential` and `verifyPresentation` so that all verification steps are covered, including querying the attestation. They now return the credential enriched with attester and revocation info. Assertions about both are __NOT__ made, but these depend a lot on your application logic and are trivial to do with the data returned from these functions.

I like the idea of having a data model for a `VerifiedCredential` bc it allows introducing another function, called `refreshRevocationStatus` here, which queries the attestation again and throws not only failing if the attestation has been deleted or has been recreated with incoherent data, but also if it has been recreated by a different attester. This avoids potential exploits where a verifier who only checks the revoked flag once a credential has been verified is tricked by a claimer who simply self-attests their credential after the original attester had deleted the attestation.

If preferred, I could return just revocation status and attester though, which would then mean `refreshRevocationStatus` would take both the credential and the last status as input parameters.

## Usage Example

```js
const didsWithAccess = new Set()
try {
  const verifiedCredential = await verifyPresentation(presentation, {
    ctype,
    challenge,
  })
  if(verifiedCredential.revoke === false) {
    // access granted
    didsWithAccess.add(presentation.claim.owner)
  }
  // periodically recheck if credential is still valid
  setInterval(() => {
    recheckRevocationStatus(verifiedCredential)
      .then(c => if(c.revoked !== false) throw new Error())
      .catch(
        () => {
          // access revoked
          didsWithAccess.delete(presentation.claim.owner)
        }
      )
  }, 60_000)
} catch {
  // access denied
}
```

## How to test:

Test coverage in unit tests and integration tests has been added.

## Checklist:

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -339,41 +339,81 @@ export async function verifyPresentation(
   })
 }
 
+export interface TrustPolicy {
+  /**
+   * If set to true, attestations issued directly by this identity are accepted.
+   */
+  isAttester?: boolean
+  /**
+   * If set to true, attestations linked to a delegation node which this identity has control over are accepted.
+   */
+  isDelegator?: boolean
+}
+export type TrustPolicies = Record<DidUri, TrustPolicy>
+
 /**
  * Queries the attestation record for a credential and matches their data. Fails if no attestation exists, if it is revoked, or if the attester is unknown.
  *
  * @param credential The [[ICredential]] whose attestation status should be checked.
- * @param trustedAttesters A list of one or more attesters who are accepted as issuers of the credential. If the actual attester is not in trustedAttesters, and this is not a delegated attestation, verification fails.
- * @param trustedDelegators An optional list of accepted delegators. If this is a delegated attestation and its issuer is not trusted, verification will still be successful if any of these identities is the owner of an ancestor node in the delegation tree.
+ * @param allowedAuthorities A map from DIDs to trust policies to be applied for that DID. If the credential's attestation cannot be linked to one of the identies in this set, verifiation will fail. If omitted, the issuer of the attestation will not be checked.
+ * @param allowRevoked If true, a revoked attestation will not fail verification.
+ * @returns Information on the attester and revocation status of the on-chain attestation, as well as info on which trust policy led to acceptance of the credential.
  */
-export async function verifyAttestation(
+export async function verifyAttested(
   credential: ICredential,
-  trustedAttesters: DidUri[],
-  trustedDelegators: DidUri[] = []
-): Promise<void> {
+  allowedAuthorities?: TrustPolicies,
+  allowRevoked = false
+): Promise<{
+  attester: DidUri
+  revoked: boolean
+  matchedTrustPolicy: Partial<Record<keyof TrustPolicy, DidUri>>
+}> {
   const api = ConfigService.get('api')
   const { rootHash } = credential
   const maybeAttestation = await api.query.attestation.attestations(rootHash)
   if (maybeAttestation.isNone)
     throw new SDKErrors.CredentialUnverifiableError('Attestation not found')
-  const attestation = Attestation.fromChain(maybeAttestation, rootHash)
+  const {
+    cTypeHash,
+    delegationId,
+    owner: attester,
+    revoked,
+  } = Attestation.fromChain(maybeAttestation, rootHash)
   if (
-    credential.claim.cTypeHash !== attestation.cTypeHash ||
-    credential.delegationId !== attestation.delegationId
+    credential.claim.cTypeHash !== cTypeHash ||
+    credential.delegationId !== delegationId
   ) {
     throw new SDKErrors.CredentialUnverifiableError(
       'Attestation does not match credential'
     )
   }
-  if (attestation.revoked)
+  if (revoked && allowRevoked !== true)
     throw new SDKErrors.RevokedTypeError('Attestation revoked')
-  if (trustedAttesters.includes(attestation.owner)) return
-  if (credential.delegationId && trustedDelegators?.length > 0) {
-    const delegation = await DelegationNode.fetch(credential.delegationId)
-    const { node } = await delegation.findAncestorOwnedBy(trustedDelegators)
-    if (node) return
+  if (typeof allowedAuthorities === 'undefined')
+    return { attester, revoked, matchedTrustPolicy: {} }
+  if (allowedAuthorities[attester]?.isAttester === true)
+    return { attester, revoked, matchedTrustPolicy: { isAttester: attester } }
+  if (credential.delegationId) {
+    const trustedDelegators: DidUri[] = []
+    Object.entries(allowedAuthorities).forEach(([did, trust]) => {
+      if (trust.isDelegator === true) {
+        trustedDelegators.push(did as DidUri)
+      }
+    })
+    if (trustedDelegators.length > 0) {
+      const delegation = await DelegationNode.fetch(credential.delegationId)
+      const { node } = await delegation.findAncestorOwnedBy(trustedDelegators)
+      if (node)
+        return {
+          attester,
+          revoked,
+          matchedTrustPolicy: { isDelegator: node.account },
+        }
+    }
   }
-  throw new SDKErrors.CredentialUnverifiableError('Attester unknown')
+  throw new SDKErrors.CredentialUnverifiableError(
+    'This attestation does not match any given trust policy and thus is not trusted'
+  )
 }
 
 /**
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -333,12 +333,16 @@ export async function verifyAttested(
     owner: attester,
     revoked,
   } = Attestation.fromChain(maybeAttestation, rootHash)
-  if (
-    credential.claim.cTypeHash !== cTypeHash ||
-    credential.delegationId !== delegationId
-  ) {
+  const ctypeMismatch = credential.claim.cTypeHash !== cTypeHash
+  const delegationMismatch = credential.delegationId !== delegationId
+  if (ctypeMismatch || delegationMismatch) {
     throw new SDKErrors.CredentialUnverifiableError(
-      'Attestation does not match credential'
+      `Some attributes of the on-chain attestation diverge from the credential: ${[
+        'cTypeHash',
+        'delegationId',
+      ]
+        .filter((_, i) => [ctypeMismatch, delegationMismatch][i])
+        .join(', ')})}`
     )
   }
   if (revoked && allowRevoked !== true) {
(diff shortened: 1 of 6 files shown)


packages/core/src/credential/Credential.ts
```diff
@@ -292,118 +277,114 @@ export function fromClaim(
...
- * @param allowedAuthorities A map from DIDs to trust policies to be applied for that DID. If the credential's attestation cannot be linked to one of the identies in this set, verifiation will fail. If omitted, the issuer of the attestation will not be checked.
- * @param allowRevoked If true, a revoked attestation will not fail verification.
+ * @param throwOnRevoked If not set to false, a revoked attestation will result in a rejected Promise.
  * @returns Information on the attester and revocation status of the on-chain attestation, as well as info on which trust policy led to acceptance of the credential.
  */
 export async function verifyAttested(
   credential: ICredential,
-  allowedAuthorities?: TrustPolicies,
-  allowRevoked = false
+  throwOnRevoked = true
```
ntn-x2, line 318: Why doesn't this function also returns a `VerifiedCredential`? Plus, I agree that removing the `throwOnRevoked` flag is more future-proof and less prone to mistakes, as people might want to check other stuff as well.

ntn-x2 (commented):
So now verifying a delegation and the list of trusted attesters would be outside of any function calls, is that correct?

rflechtner:
> So now verifying a delegation and the list of trusted attesters would be outside of any function calls, is that correct?

Correct! And given you agree that the revocation check can go to, so would be checking if revoked === false.

ntn-x2:
+1 for this solution then. I like it. Not sure if the additional type would be more confusing for the users. Maybe @weichweich can provide his opinion on this a well. Maybe @arty-name also?

rflechtner:
Yes, would love to hear their opinions. Compare this one as well against having a separate function for checking the attestation (i.e. not changing verifyCredential at all - see for example https://github.com/KILTprotocol/sdk-js/pull/709, but could also be a regular function) and let me know which of the three you think is best.

arty-name:
I like this much more than the proposal I have seen previously! The alternative of having a separate function does not appeal to me either.

rflechtner:
So to summarise, this seems to be the preferred solution rn, and we'd probably drop the `throwOnRevoked` parameter too. In that case I'd probably rename `recheckRevocationStatus` to `refreshRevocationStatus` (or `updateRevocationStatus`).
You'd get an object back which gives you revocation status and attester, and you'd compare their values against your expectations (e.g. you reject revoked credentials or not, or you only allow attesters from a self-curated list).

If the `VerifiedCredential` type is confusing, we can also introduce a `CredentialStatus` type instead, which would change the signature of `refreshRevocationStatus` too:
```ts
type CredentialStatus = {
  attester: DidUri
  revoked: boolean
}

async function refreshRevocationStatus(
   credential: ICredential,
   lastStatus: CredentialStatus
): Promise<CredentialStatus>
```

packages/core/src/credential/Credential.ts
```diff
@@ -292,118 +277,114 @@ export function fromClaim(
...
+  ctype?: ICType
+  challenge?: string
+  didResolveKey?: DidResolveKey
+  throwOnRevoked?: boolean
+}
+
+/**
+ * Verifies data structure & data integrity of a credential object.
+ * This combines all offline sanity checks that can be performed on an ICredential object.
+ * A credential is valid only iff it is well formed and there is an onchain attestation record referencing its root hash.
```
Dudleyneedham, line 290:
```suggestion
 * A credential is valid only if it is well formed and there is an on-chain attestation record referencing its root hash.
```

packages/core/src/credential/Credential.ts
```diff
@@ -292,118 +277,114 @@ export function fromClaim(
...
+ * @param throwOnRevoked If not set to false, a revoked attestation will result in a rejected Promise.
+ * @returns A promise of resolving to the same object but with the `revoked` property updated. The promise rejects if the attestation has been deleted or its data changed since verification.
+ */
+export async function recheckRevocationStatus(
+  verifiedCredential: VerifiedCredential,
+  throwOnRevoked = true
+): Promise<VerifiedCredential> {
+  const { revoked, attester } = await verifyAttested(
+    verifiedCredential,
+    throwOnRevoked
```
Dudleyneedham, line 359: Do we not want this always to be the case? We decided that if there was an error, we'd always throw it.

Is there a reason why we might not?

packages/core/src/credential/Credential.ts
```diff
@@ -413,21 +394,25 @@ export async function verifyCredential(
  *
  * @param presentation - The object to check.
  * @param options - Additional parameter for more verification steps.
+ * @param options.ctype - CType which the included claim should be checked against.
```
Dudleyneedham, line 397: Shouldn't this always be passed?

packages/core/src/credential/Credential.ts
```diff
@@ -292,118 +277,114 @@ export function fromClaim(
...
-  /**
-   * If set to true, attestations issued directly by this identity are accepted.
-   */
-  isAttester?: boolean
-  /**
-   * If set to true, attestations linked to a delegation node which this identity has control over are accepted.
-   */
-  isDelegator?: boolean
+type VerifyOptions = {
+  ctype?: ICType
```
Dudleyneedham, line 281: I find it strange that CType isn't required. Is there a specific reason why?

Dudleyneedham (commented):
Mostly comments. The code looks good.

packages/core/src/credential/Credential.ts
arty-name: I’m not the only one thrown off by the usage of "iff" in the sense "if and only if" :)

weichweich: Is this even true? If I have a well structured credential with an on-chain attestation record it can still be invalid? E.g. if the delegation record is wrong or doesn't exist.

rflechtner: You're right, this isn't complete; the attestation also needs to match the credential, i.e. delegation and ctype ids need to be identical to what was committed to in the credential. BTW that's something we should have designed differently; the runtime should combine at least delegation & ctype info into the credential hash so there is no chance these data do not match if the hash is identical. Ideally also the attester and block number, just to make sure the same attestation is never re-created.

packages/core/src/credential/Credential.ts
rflechtner: see previous comment

packages/core/src/credential/Credential.ts
rflechtner: Anyway, we would do these checks as part of verifyCredential now.

packages/core/src/credential/Credential.ts
rflechtner: Because it does not actually verify a credential. Just checks the attestation. Name suggestions for that function are welcome.

packages/core/src/credential/Credential.ts
ntn-x2: This should be part of the discussion for credentials v2. We can build on the current limitations and do better.

ntn-x2:
Just a point, that relying too much on Typescript could leave Javascript developers in the dark. But I like the name `VerifiedCredential`. It also does not feel right that a `verify` function returns a `CredentialStatus`. Different is, if the function is renamed, then it might make sense.

arty-name:
I think both options are fine, and I have a slight preference for the current VerifiedCredential

rflechtner:
> Just a point, that relying too much on Typescript could leave Javascript developers in the dark.

I hope that can be resolved with good docstrings.

weichweich:
> Just a point, that relying too much on Typescript could leave Javascript developers in the dark. But I like the name `VerifiedCredential`. It also does not feel right that a `verify` function returns a `CredentialStatus`. Different is, if the function is renamed, then it might make sense.

I agree with Antonio. In JS code i would just use

```
if (verifyCredential(...)) {
  ...
}
```

And that's always true.

rflechtner:
> In JS code i would just use
> 
> ```
> if (verifyCredential(...)) {
>   ...
> }
> ```
> 
> And that's always true.

And thaaaaaat's why we banished you to the Rust dungeon! Please don't ever do that ever. You're calling an async function, so it returns a Promise which is always truthy, even if the function throws an exception, and even if the value it wraps is itself falsy.

So I don't think the average JS developer works on the basis of 'just call some function and assume it returns a boolean'. You read the docstring at the very least, how else would you know what it does, when it throws, and what it returns? The same approach would fail miserably with vc.js's verify function signature (https://github.com/digitalbazaar/vc#verifying-a-verifiable-presentation), where you get the final result from the `verified` property of the (awaited) return value. I think Antonio's point was rather to not assume people have access to interfaces and types, which I don't think is a big issue here though. I think we can describe the required structure by saying 'this structure, but with two fields added' or simply by saying 'the return value of this function', and we should have JS developers covered?

packages/core/src/credential/Credential.ts
```diff
@@ -293,27 +285,109 @@ type VerifyOptions = {
...
 
   if (ctype) {
-    verifyAgainstCType(credential, ctype)
+    verifyClaimAgainstSchema(credential.claim.contents, ctype)
   }
 }
 
 /**
- * Verifies data structure, data integrity and the claimer's signature of a credential presentation.
+ * Queries the attestation record for a credential and matches their data. Fails if no attestation exists, if it is revoked, or if the attestation data does not match the credential.
```
ntn-x2, line 309: > as well as info on which trust policy led to acceptance of the credential.

Where?

ntn-x2 (commented):
Looks very good to me! I could not find any mention in the JSDocs about the consumer actually having to check the trusted attester list and the delegation information.

packages/core/src/credential/Credential.ts
rflechtner: I think you reviewed an old version? I recognise the quote but I changed that afaik. At least it's not in the line you commented on?

rflechtner:
> Looks very good to me! I could not find any mention in the JSDocs about the consumer actually having to check the trusted attester list and the delegation information.

Again, you don't _have_ to match the attester to some list of trusted attesters in your application; your application could also limit itself to simply displaying the attester DID or web3name to the user, like most of our apps do. That's why this solution is better than the previous candidate, which I think was too opinionated. I could add something like "this data can be used in logic deciding whether to accept or reject a credential submission", but a naive user would probably need more context to understand what that means, which they can only get from the docs - the docstring wouldn't be the right place to explain conceptual fundamentals, would it?

packages/core/src/credential/Credential.ts
ntn-x2: Shit ok, no idea what happened to my VS Code.

ntn-x2 (approved):
I reviewed a VERY old version of this PR, somehow. The new changes look way better, sorry about that. Great work!

PR #709, opened 2023-02-01
feat: credential verification derives

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/2397

Potential alternative to #708 & #702.

Instead of changing how `verifyCredential` & `verifyPresentation` work, we add a `verifyAttested` function that is to be called after these. 

Uses api derives to add credential verification functionality to the api object, with the following signature:

```js
const { attester, revoked, error } = await api.derive.credential.verifyAttested(credential)
```

This fetches the attestation and checks it against the credential as well, returning `revoked` status and `attester` if successful, and `error` if there is an issue.

__Pros__

- Subscriptions are possible.
- verifyCredential & verifyPresentation are left untouched.
- Bundles credential verification with the api.

__Cons__

-  Ultimately doesn't differ much from a function-based approach that calls api.query internally.
- Derives need to be added during api initialisation, so if people construct the api themselves they could forget.

## How to test:
_tbd_

## Checklist:

### DIFF
--- a/packages/core/src/credential/derives.ts
+++ b/packages/core/src/credential/derives.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2018-2023, BOTLabs GmbH.
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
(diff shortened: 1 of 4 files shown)


rflechtner:
closing in favour of https://github.com/KILTprotocol/sdk-js/pull/708

PR #711, opened 2023-02-06
add: constant for well-Known-did-config context

kilted-andres:
When creating a well-known-did-configuration VerifiableDomainLinkagePresentation, I can import all the required constants from here but one. That seems inconsistent to me. I suggest adding it.


rflechtner (commented):
I see your point. So far we have only included constants here that are required to form any KILT-type verifiable credential, which is a purpose general envelope to attest to some data; there will always be applications and use cases that come with additional conventions and standards (such as the domain linkage credential specifications) and we shouldn't try to cater to all of them. So I'm not sure this is the right place.

kilted-andres:
ok, I understand. Is sad though.

PR #712, opened 2023-02-06
feat: add did contexts

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/1999

Adds static copies of json-ld context files relevant to KILT DID documents to the sdk, as well as the respective URI to DID documents exported to the conformant json-ld representation.

## How to test:
n/a

## Checklist:

### DIFF
--- a/packages/did/src/DidDocumentExporter/DidContexts.ts
+++ b/packages/did/src/DidDocumentExporter/DidContexts.ts
@@ -1,0 +1,196 @@
+/**
+ * Copyright (c) 2018-2023, BOTLabs GmbH.
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * IPFS URL identifying a JSON-LD context file describing terms used in DID documents of the KILT method that are not defined in the W3C DID core context.
+ * Should be the second entry in the ordered set of contexts after [[W3C_DID_CONTEXT_URL]] in the JSON-LD representation of a KILT DID document.
+ */
+export const KILT_DID_CONTEXT_URL =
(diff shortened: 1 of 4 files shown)


packages/did/src/DidDocumentExporter/DidContexts.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Must be the first entry in the ordered set of contexts in a JSON-LD representation of a DID document.
+ * See https://www.w3.org/TR/did-core/#json-ld.
+ */
+export const W3C_DID_CONTEXT_URL = 'https://www.w3.org/ns/did/v1'
+/**
+ * URL identifying a JSON-LD context file proposed by the W3C Credentials Community Group defining a number of terms which are used in verification methods on KILT DID documents.
+ * See https://w3c-ccg.github.io/security-vocab/.
+ * This document is extended by the context file available under the [[KILT_DID_CONTEXT_URL]].
+ */
+export const W3C_SECURITY_CONTEXT_URI = 'https://w3id.org/security/v2'
```
ntn-x2, line 25: This is also a URL tho, right? What about renaming them all to URI? Since the IPFS link is definitely not a URL.

ntn-x2 (approved):
LGTM!

ntn-x2:
I'm merging the PR since I will build on top of this to tackle https://github.com/KILTprotocol/ticket/issues/2093.

packages/did/src/DidDocumentExporter/DidContexts.ts
rflechtner: actually I think I wanted to name it "URL" bc last thing I heard was that the distinction IRI / URI / URL is being abandoned in favour of calling everything "URL". Why is an ipfs link not a URL? It does resolve and thus let you locate a resource, no?

ntn-x2: As far as I know, URL is only used for web resources.

PR #713, opened 2023-02-06
fix: align dependency versions

rflechtner:
Fixes minor incoherences in dependencies that are shared between multiple sub-packages.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -17,11 +17,11 @@
   "homepage": "https://github.com/KILTprotocol/sdk-js#readme",
   "dependencies": {
     "@kiltprotocol/sdk-js": "workspace:*",
-    "@polkadot/keyring": "^10.0.0",
+    "@polkadot/keyring": "^10.2.1",
     "@polkadot/types": "^9.10.2",
-    "@polkadot/util": "^10.0.0",
+    "@polkadot/util": "^10.2.1",
     "typescript-logging": "^0.6.4",
-    "@polkadot/util-crypto": "^10.0.0",
+    "@polkadot/util-crypto": "^10.2.1",
     "testcontainers": "^8.6.1"
   }
 }
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -38,7 +38,7 @@
   "devDependencies": {
     "@polkadot/api": "^9.10.2",
     "@polkadot/typegen": "^9.10.2",
-    "@types/websocket": "^1.0.0",
+    "@types/websocket": "^1.0.5",
     "glob": "^7.1.1",
     "rimraf": "^3.0.2",
     "ts-node": "^10.4.0",
--- a/packages/augment-api/package.json
+++ b/packages/augment-api/package.json
@@ -43,7 +43,7 @@
     "rimraf": "^3.0.2",
     "ts-node": "^10.4.0",
     "typescript": "^4.8.3",
-    "websocket": "^1.0.31",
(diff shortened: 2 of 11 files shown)


docs/package.json
```diff
@@ -17,11 +17,11 @@
   "homepage": "https://github.com/KILTprotocol/sdk-js#readme",
   "dependencies": {
     "@kiltprotocol/sdk-js": "workspace:*",
-    "@polkadot/keyring": "^10.0.0",
+    "@polkadot/keyring": "^10.2.1",
```
rflechtner, line 20: all 9.10.2 versions of the api packages require at least 10.2.1 of the utils packages

packages/augment-api/package.json
```diff
@@ -38,12 +38,12 @@
...
     "@polkadot/api": "^9.10.2",
     "@polkadot/typegen": "^9.10.2",
-    "@types/websocket": "^1.0.0",
+    "@types/websocket": "^1.0.5",
     "glob": "^7.1.1",
     "rimraf": "^3.0.2",
     "ts-node": "^10.4.0",
     "typescript": "^4.8.3",
-    "websocket": "^1.0.31",
+    "websocket": "^1.0.34",
```
rflechtner, line 46: aligning with polkadot/typegen

packages/chain-helpers/package.json
```diff
@@ -31,7 +31,7 @@
   "homepage": "https://github.com/KILTprotocol/sdk-js#readme",
   "devDependencies": {
     "@kiltprotocol/testing": "workspace:*",
-    "@polkadot/keyring": "^10.0.0",
+    "@polkadot/keyring": "^10.2.1",
```
rflechtner, line 34: see above

packages/did/package.json
```diff
@@ -40,11 +40,11 @@
...
-    "@polkadot/keyring": "^10.0.0",
+    "@polkadot/keyring": "^10.2.1",
     "@polkadot/types": "^9.10.2",
     "@polkadot/types-codec": "^9.10.2",
-    "@polkadot/util": "^10.0.0",
-    "@polkadot/util-crypto": "^10.0.0",
-    "cbor": "^8.0.2"
+    "@polkadot/util": "^10.2.1",
+    "@polkadot/util-crypto": "^10.2.1",
+    "cbor": "^8.1.0"
```
rflechtner, line 48: the core package requires at least 8.1.0

ntn-x2 (approved):
LGTM! I was wondering, is there any tool that works like [Cargo root workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) does, which defines all dependencies and their versions only once?

arty-name (approved):

arty-name:
I think this is one of those problems that would be inconceivable with the good old single package approach :)

rflechtner:
> LGTM! I was wondering, is there any tool that works like [Cargo root workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) does, which defines all dependencies and their versions only once?

Well that's at odds with the idea of having multiple packages with multiple package.json's defining dependency ranges, isn't it? We have one lock file in the root workspace, but I think that's as close to this as it gets.

rflechtner:
> I think this is one of those problems that would be inconceivable with the good old single package approach :)

I knooooowwww, one of these days... it's quite a big breaking change though, so probably not happening anytime soon.

PR #725, opened 2023-02-14
fix: light DID decoding logic

ntn-x2:


packages/did/src/DidDetails/LightDidDetails.ts
```diff
@@ -132,10 +136,13 @@ function serializeAdditionalLightDidDetails({
...
-      id: resourceIdToChain(id),
-      ...rest,
-    }))
+    objectToSerialize[SERVICES_MAP_KEY] = service.map(
+      ({ id, serviceEndpoint, type }) => ({
+        id: resourceIdToChain(id),
+        serviceEndpoint,
+        type,
+      })
+    )
```
rflechtner, line 145: hmm how does this help? I think it was more compatible the way it was before. This way we'll just end up with serialised service endpoints where serviceEndpoint is undefined and url is silently dropped.

rflechtner (changes requested):
I can't see a reason to change the section I flagged but the rest looks good!

packages/did/src/DidDetails/LightDidDetails.ts
ntn-x2: This is serialization of the new stuff. So new endpoints always serialize to id, serviceEndpoint, type. The trick is in the DEserialization, which attemps to use serviceEndpoint and type, and falls back to types and urls if those are not found.

ntn-x2: In this snippet here I replaced `...rest`, which was anyway just `serviceEndpoint` and `type` with its building components.

rflechtner (approved):

rflechtner:
At least the example uses the proper keys (`[{ id: 'my-service-id', type: ['my-service-type'], serviceEndpoint: ['my-service-url'] }]`)

ntn-x2:
Because we changed the spec. Updating the spec does not mean we should not support old DIDs, right? If we do, that would also be fine, but then the resolver (driver) should not fail with 500 internal error. I think since we have the fix here, it would not hurt to have it in. I also don't expect there to be that many old DIDs left. But some might, and might also even have credentials.

ntn-x2:
Or maybe I am still missing something that happens between deserialization and re-serialization. I only know the SDK throws and the DID resolver returns 500. What fix do we want to implement then?

rflechtner:
Ok let's merge it then.

packages/did/src/DidDetails/LightDidDetails.ts
rflechtner: Well not if you're dealing with a legacy DID, then it's the legacy keys

rflechtner: although the services validation runs beforehand and should not allow the use of any other keys 🤔 anyway, it's good that we reverted it bc that was not the issue and the remainder syntax is less brittle.

PR #726, opened 2023-02-15
fix: current CType meta schema ignores additional properties in claim contents

rflechtner:
## fixes KILTProtocol/ticket#2378

Introduces a new meta schema to fix the issue that additional properties were not validated.
Newly created CTypes would use this meta schema by default; legacy CTypes are still validated against the old meta schema.

- [x] Makes sure that no additional properties are present on an IClaim
- [x] Allows for selective disclosure (does not require all allowable properties to be present)
- [x] Meta schema id uses content integrity protected URI, ideally uses decentralised solution
- [x] Comes with a strategy for expanding claims to URIs and for checking the same claims on the credentialSubject of a Verifiable Credential
   - Unfortunately it seems that the only viable strategy so far is deriving a schema from the CType.
   - Better compatibility could be achieved by defining an optional `@id` property on all CTypes (to avoid failing credentialSubjecct verification on VCs, where this would be present after expanding), but this comes at a cost; any IClaim would pass verification even with an extra `@id` property.
- [x] Upgrade path / interoperability strategy for credentials issued for the old CType version
   -  Because the currently used json schema version does not allow defining a CType which derives the full set of properties from another CType (you can only add references for each property individually) the recommended upgrade path would be to create a new CType with the same properties and publish proof that it is equivalent to the old e.g. by publishing code that reproduces the derivation. Creating a CType where each property is derived from the old CType is another option, but checking that no properties were omitted and that the property names are identical is still a manual process. Compound/nested CTypes are also a bit more complicated to handle, so this seems like the less preferable option.
   - If all properties on the new CType are identical, verifiers would only need to add the new CType to the set of accepted CTypes. The credential could then be processed as before, no matter which version of the CType was used. 

## How to test:

Updated tests to cover both the old and new meta schema and CTypes created based on them. 

## Checklist:

### DIFF
--- a/packages/core/src/ctype/CType.schemas.ts
+++ b/packages/core/src/ctype/CType.schemas.ts
@@ -9,25 +9,33 @@ import { JsonSchema } from '@kiltprotocol/utils'
 
 export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
   // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
-  $id: 'ipfs://bafybeibjbq3tmmy7wuihhhwvbladjsd3gx3kfjepxzkq6wylik6wc3whzy',
+  $id: 'ipfs://bafybeifzfxz6tfd2xo7ijxbfceaxo3l655yg7sovlsnpxgq2rwfl4kbfgm',
   $schema: 'http://json-schema.org/draft-07/schema#',
-  additionalProperties: false,
+  title: 'CType Metaschema (V1)',
   description:
-    'Describes a CType, which is a json schema for validating KILT claim types.',
+    'Describes a CType, which is a JSON schema for validating KILT claim types.',
+  type: 'object',
   properties: {
     $id: { pattern: '^kilt:ctype:0x[0-9a-f]+$', type: 'string' },
     $schema: {
       type: 'string',
       // can't use a const referencing schema id for a content-addressed schema
     },
-    additionalProperties: { const: false, type: 'boolean' },
+    title: { type: 'string' },
+    type: { const: 'object', type: 'string' },
     properties: {
       patternProperties: {
-        '^.*$': {
+        '^.+$': {
           oneOf: [
             {
               additionalProperties: false,
-              properties: { $ref: { format: 'uri', type: 'string' } },
+              properties: {
+                $ref: {
+                  pattern: '^kilt:ctype:0x[0-9a-f]+(#/properties/.+)?$',
+                  format: 'uri',
+                  type: 'string',
+                },
+              },
               required: ['$ref'],
             },
             {
@@ -47,9 +55,9 @@ export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
       },
       type: 'object',
     },
-    title: { type: 'string' },
-    type: { const: 'object', type: 'string' },
+    additionalProperties: { const: false, type: 'boolean' },
   },
+  additionalProperties: false,
   required: [
     '$id',
     '$schema',
@@ -58,8 +66,6 @@ export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
     'title',
     'type',
   ],
-  title: 'CType Metaschema (V1)',
-  type: 'object',
 }
 
 export const CTypeModelDraft01: JsonSchema.Schema & { $id: string } = {
--- a/packages/core/src/ctype/CType.schemas.ts
+++ b/packages/core/src/ctype/CType.schemas.ts
@@ -7,7 +7,7 @@
 
 import { JsonSchema } from '@kiltprotocol/utils'
 
-const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
+export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
   $id: 'http://kilt-protocol.org/draft-01/ctype#',
   $schema: 'http://json-schema.org/draft-07/schema#',
   type: 'object',
@@ -65,7 +65,7 @@ const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
   required: ['$id', 'title', '$schema', 'properties', 'type'],
 }
 
-const CTypeModelV2: JsonSchema.Schema & { $id: string } = {
+export const CTypeModelV2: JsonSchema.Schema & { $id: string } = {
   // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
   $id: 'ipfs://bafybeigphkblu3nvk5tpz4xog3flpv4h6o4x6fjml6haead3pul2mny3ee',
   $schema: 'http://json-schema.org/draft-07/schema#',
--- a/packages/core/src/ctype/CType.schemas.ts
+++ b/packages/core/src/ctype/CType.schemas.ts
@@ -7,7 +7,7 @@
 
 import { JsonSchema } from '@kiltprotocol/utils'
 
-export const CTypeModel: JsonSchema.Schema = {
+const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
   $id: 'http://kilt-protocol.org/draft-01/ctype#',
   $schema: 'http://json-schema.org/draft-07/schema#',
   type: 'object',
@@ -63,6 +63,101 @@ export const CTypeModel: JsonSchema.Schema = {
   },
   additionalProperties: false,
   required: ['$id', 'title', '$schema', 'properties', 'type'],
+}
+
+const CTypeModelV2: JsonSchema.Schema & { $id: string } = {
+  // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
+  $id: 'ipfs://bafybeigphkblu3nvk5tpz4xog3flpv4h6o4x6fjml6haead3pul2mny3ee',
+  $schema: 'http://json-schema.org/draft-07/schema#',
+  type: 'object',
+  properties: {
+    $id: {
+      type: 'string',
+      pattern: '^kilt:ctype:0x[0-9a-f]+$',
+    },
+    $schema: {
+      type: 'string',
+      // can't use a const refercing schema id for a content-addressed schema
+    },
+    title: {
+      type: 'string',
+    },
+    type: {
+      type: 'string',
+      const: 'object',
+    },
+    additionalProperties: {
+      type: 'boolean',
+      const: false,
+    },
+    properties: {
+      type: 'object',
+      patternProperties: {
+        '^.*$': {
+          type: 'object',
+          oneOf: [
+            {
+              properties: {
+                type: {
+                  type: 'string',
+                  enum: ['string', 'integer', 'number', 'boolean'],
+                },
+                format: {
+                  type: 'string',
+                  enum: ['date', 'time', 'uri'],
+                },
+              },
+              required: ['type'],
+              additionalProperties: false,
+            },
+            {
+              properties: {
+                $ref: {
+                  type: 'string',
+                  format: 'uri',
+                },
+              },
+              required: ['$ref'],
+              additionalProperties: false,
+            },
+          ],
+        },
+      },
+    },
+  },
+  additionalProperties: false,
+  required: [
+    '$id',
+    'title',
+    '$schema',
+    'properties',
+    'type',
+    'additionalProperties',
+  ],
+}
+
+export const CTypeModels: Record<string, JsonSchema.Schema> = Object.freeze({
+  [CTypeModelV1.$id]: CTypeModelV1,
+  [CTypeModelV2.$id]: CTypeModelV2,
+})
+
+export const CTypeModel: JsonSchema.Schema = {
+  oneOf: [
+    CTypeModelV1,
+    {
+      allOf: [
+        {
+          properties: {
+            $schema: {
+              type: 'string',
+              const: CTypeModelV2.$id,
+            },
+          },
+        },
+        CTypeModelV2,
+      ],
+    },
+  ],
 }
 
 export const MetadataModel = {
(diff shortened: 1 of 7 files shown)


packages/core/src/ctype/CType.schemas.ts
```diff
@@ -65,6 +122,25 @@ export const CTypeModel: JsonSchema.Schema = {
   required: ['$id', 'title', '$schema', 'properties', 'type'],
 }
 
+export const CTypeModel: JsonSchema.Schema = {
+  oneOf: [
+    CTypeModelDraft01,
```
rflechtner, line 127: we're embedding the schemas here; we could also use a `$ref` here, which would however require passing the two schemas when calling `verifyObjectAgainstSchema` (`verifyObjectAgainstSchema(ctype, CTypeModel, [CTypeModelV1, CTypeModelDraft01])`)

packages/core/src/ctype/CType.schemas.ts
```diff
@@ -7,9 +7,66 @@
...
+  ],
+  title: 'CType Metaschema (V1)',
+  type: 'object',
+}
+
+export const CTypeModelDraft01: JsonSchema.Schema & { $id: string } = {
   $id: 'http://kilt-protocol.org/draft-01/ctype#',
   $schema: 'http://json-schema.org/draft-07/schema#',
+  title: 'CType Metaschema (draft-01)',
+  description: `Describes a CType, which is a json schema for validating KILT claim types. This version has known issues, the use of schema ${CTypeModelV1.$id} is recommended instead.`,
```
rflechtner, line 69: we should update the hosted version if merged

packages/core/src/ctype/CType.ts
```diff
@@ -93,15 +98,17 @@ export function getIdForSchema(
...
  * @param schema Schema to verify against.
  * @param messages Optional empty array. If passed, this receives all verification errors.
+ * @param referencedSchemas If schema contains references ($ref) to other schemas, their definitions must be added here in form of an array.
  */
 export function verifyObjectAgainstSchema(
   object: Record<string, any>,
   schema: Record<string, any>,
-  messages?: string[]
+  messages?: string[],
+  referencedSchemas?: Array<Record<string, any>>
```
ntn-x2, line 107: `validator.addSchema` takes a `Schema` parameter, so is this `Record<string, any>` just that?

packages/core/src/ctype/CType.ts
```diff
@@ -175,21 +182,8 @@ export function verifyClaimAgainstNestedSchemas(
...
-  if (valid === true) return
-  if (messages) {
-    errors.forEach((error) => {
-      messages.push(error.error)
-    })
-  }
-  throw new SDKErrors.NestedClaimUnverifiableError(undefined, {
-    cause: errors,
-  })
+  verifyObjectAgainstSchema(cType, CTypeModel, messages)
```
ntn-x2, line 185: Question, probably out of scope for this PR: how is a verifier supposed to know what are the nested CTypes? Are they assumed to have this knowledge? Or might they only know the top-level CType they want to verify a credential against?

packages/core/src/ctype/CType.schemas.ts
ntn-x2: I like this approach more, perhaps a comment about what this actually does (i.e., not breaking existing credentials) would help!

ntn-x2 (commented):
Looks very good! I would like to understand a bit more how it all works, if you answer my questions. Then I would take a second look after that :)

packages/core/src/ctype/CType.schemas.ts
rflechtner: maybe the way to go is using the strategy described as "bundling" schemas (https://json-schema.org/understanding-json-schema/structuring.html#bundling). It's sort of a hybrid between the two options we had so far: you use `$ref`s, but you put the referenced schemas in the `definitions` section of the document so that they are still available to the validator simply by loading the bundled schema.

packages/core/src/ctype/CType.schemas.ts
```diff
@@ -7,9 +7,66 @@
...
 import { JsonSchema } from '@kiltprotocol/utils'
 
-export const CTypeModel: JsonSchema.Schema = {
+export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
+  // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
+  $id: 'ipfs://bafybeibjbq3tmmy7wuihhhwvbladjsd3gx3kfjepxzkq6wylik6wc3whzy',
+  $schema: 'http://json-schema.org/draft-07/schema#',
+  additionalProperties: false,
+  description:
+    'Describes a CType, which is a json schema for validating KILT claim types.',
```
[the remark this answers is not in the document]
rflechtner: would require updating the schema on ipfs and its CID here; would do that in case there are any other changes required as it's merely cosmetic.

packages/core/src/ctype/CType.schemas.ts
```diff
@@ -65,6 +122,25 @@ export const CTypeModel: JsonSchema.Schema = {
   required: ['$id', 'title', '$schema', 'properties', 'type'],
 }
 
+export const CTypeModel: JsonSchema.Schema = {
+  oneOf: [
+    CTypeModelDraft01,
+    {
+      allOf: [
+        {
+          properties: {
```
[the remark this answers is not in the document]
rflechtner: no, allOf does not represent a merge; it's just a logical AND for schemas, i.e. both schemas must validate. The first item in the array is a complete schema, it's just very simple. It only checks that the `$schema` property is equal to some value, which is the CType meta schema V1's id. Implies that we are dealing with an object. I could make this required too but that condition is already checked in the CType meta schema itself.

packages/core/src/ctype/CType.schemas.ts
```diff
@@ -7,9 +7,66 @@
...
 
-export const CTypeModel: JsonSchema.Schema = {
+export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
+  // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
+  $id: 'ipfs://bafybeibjbq3tmmy7wuihhhwvbladjsd3gx3kfjepxzkq6wylik6wc3whzy',
+  $schema: 'http://json-schema.org/draft-07/schema#',
+  additionalProperties: false,
+  description:
+    'Describes a CType, which is a json schema for validating KILT claim types.',
+  properties: {
```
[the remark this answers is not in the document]
rflechtner: You might remember that they can't be because then CTypes would not validate in case of selective disclosure.

packages/core/src/ctype/CType.spec.ts
```diff
@@ -28,76 +27,39 @@ const encodedAliceDid = ApiMocks.mockChainQueryReturn(
...
-  it('verifies the claim structure', () => {
-    expect(() =>
-      CType.verifyClaimAgainstSchema(claim.contents, claimCtype)
-    ).not.toThrow()
-    claim.contents.name = 123
-    expect(() =>
-      CType.verifyClaimAgainstSchema(claim.contents, claimCtype)
-    ).toThrow()
+it('consistent CType id generation', () => {
+  const ctype = CType.fromProperties('CtypeModel 1', {
```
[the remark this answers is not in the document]
rflechtner: As in test the old version here? Do you think that's necessary? We have a bunch of test cases below that test both CType versions, and id generation does not really depend on the CType version as it simply treats the ctype as a JSON document where the `$id` property must be stripped.

packages/core/src/ctype/CType.ts
rflechtner: I was thinking the same thing but yes, I think it's out of scope. Maybe we need a CType registry in the sdk? Would also let us validate a CType only once (when adding to the registry) and not every time it is used.

rflechtner:
As far as I can see all the upgrade paths described above but the one where people simply create a new CType based on the old would break quite a few things. So I'd suggest sticking to that. Besides this, I see only two open issue that we should clarify before putting the new CType out there: 
1. Do we consider it V1 or is it simply draft-02? (maybe a tech board question?)
2. Do we allow nested properties (objects and arrays as values) even though selective disclosure is only possible for top-level properties? I.e. you can't omit an item from an array or a property from an object.

packages/core/src/ctype/CType.schemas.ts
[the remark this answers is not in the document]
ntn-x2: Alright, clear 👍

packages/core/src/ctype/CType.ts
```diff
@@ -201,24 +195,34 @@ export function verifyCTypeMetadata(metadata: ICTypeMetadata): void {
...
+
 /**
  * Creates a new [[ICType]] object from a set of atomic claims and a title.
  * The CType id will be automatically generated.
  *
  * @param title The new CType's title as a string.
  * @param properties Key-value pairs describing the admissible atomic claims for a credential with this CType. The value of each property is a json-schema (for example `{ "type": "number" }`) used to validate that property.
- * @returns A ctype object, including cTypeId, $schema, and type.
+ * @param version Use 'V1' to create a CType according to the latest metaschema version (default) and 'draft-01' to produce a legacy CType. Included for backwards-compatibility.
+ * @returns A complete json schema (CType) with an $id derived from the hashed schema. The referenced meta $schema varies by CType version.
```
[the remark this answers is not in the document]
ntn-x2: Sorry the previous suggestion was wrong.

packages/core/src/ctype/CType.ts
ntn-x2: So not planning to change it?

packages/core/src/ctype/CType.ts
ntn-x2: That would make our SDK less stateless, which is something we wanted to strive for as much as possible. it might be a good idea to at least reference this in a comment or something.

ntn-x2:
> Do we consider it V1 or is it simply draft-02? (maybe a tech board question?)

I think this version should be quite stable, right? At least as far as our credentials currently allow. If we want to support nested selected disclosure, we would have a v2 out there. So i would be in favour of having this a v1.

> Do we allow nested properties (objects and arrays as values) even though selective disclosure is only possible for top-level properties? I.e. you can't omit an item from an array or a property from an object.

I think, for the sake of completeness, we could still allow any property, making it clear (either in the JSDocs or in the documentation itself), that only the top-level property can be revealed or hidden as a whole.

packages/core/src/ctype/CType.ts
rflechtner: just did! seems fine so far, nothing breaking internally

rflechtner:
> > Do we allow nested properties (objects and arrays as values) even though selective disclosure is only possible for top-level properties? I.e. you can't omit an item from an array or a property from an object.
> 
> I think, for the sake of completeness, we could still allow any property, making it clear (either in the JSDocs or in the documentation itself), that only the top-level property can be revealed or hidden as a whole.

As it turns out, there are quite a few details to figure out if we were to allow this, so I think it should be its own ticket with design phase and all. It would also be something that does not affect existing CTypes at all, so it could be quite simply the next meta schema version.

packages/core/src/ctype/CType.schemas.ts
```diff
@@ -7,9 +7,66 @@
...
+            {
+              additionalProperties: false,
+              properties: { $ref: { format: 'uri', type: 'string' } },
+              required: ['$ref'],
+            },
+            {
+              additionalProperties: false,
+              properties: {
+                format: { enum: ['date', 'time', 'uri'], type: 'string' },
+                type: {
```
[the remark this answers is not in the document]
rflechtner: see https://github.com/KILTprotocol/ticket/issues/2456

packages/core/src/ctype/CType.ts
rflechtner: Guess that's true. The way I see it choosing a CType is a very conscious decision for a verifier, so they'd be familiar with it and know whether or not it is nested. At least the need to known what sort of data they will be working with in their application, so they'd look into that for sure.

packages/core/src/ctype/CType.spec.ts
```diff
@@ -28,76 +27,52 @@ const encodedAliceDid = ApiMocks.mockChainQueryReturn(
...
-    const wrongHashCtype: ICType = {
-      ...claimCtype,
-      $id: 'kilt:ctype:0x1234',
-    }
-    const faultySchemaCtype: ICType = {
-      ...claimCtype,
-      properties: null as unknown as ICType['properties'],
-    }
+  expect(ctypeV0.$id).toMatchInlineSnapshot(
+    `"kilt:ctype:0xba15bf4960766b0a6ad7613aa3338edce95df6b22ed29dd72f6e72d740829b84"`
```
ntn-x2, line 50: Isn't this supposed to be the same as the old value, which is `kilt:ctype:0xdfe0fd007b86ef280f366d25547873da0305538d8190a2eae2f2d875cb5cf30d`?

ntn-x2 (approved):
Just a question, otherwise LGTM! Very very very very nice job! 👍

packages/core/src/ctype/CType.spec.ts
rflechtner: That's impossible unfortunately as the id is the Schema hash and the Schema hash is computed over the entire Schema (excluding $id obviously) so including the meta Schema reference $schema and the additionalProperties assertion

rflechtner: In short: new version, new ctype.

ntn-x2: Got it. There's something new even for old CTypes.

PR #730, opened 2023-02-27
feat: split `associateAccountToChainArgs`

weichweich:
## fixes [#2484](https://github.com/KILTprotocol/ticket/issues/2484)

This makes it possible to use the high level `associateAccountToChainArgs` in most cases, but also use the building blocks in edge cases where parts are already done by other parts of the app (e.g. the wrapping of bytes).

## Checklist:

### DIFF
--- a/packages/did/src/DidLinks/AccountLinks.chain.ts
+++ b/packages/did/src/DidLinks/AccountLinks.chain.ts
@@ -16,6 +16,7 @@ import {
   u8aConcatStrict,
   u8aToHex,
   u8aWrapBytes,
+  BN,
 } from '@polkadot/util'
 import { ApiPromise } from '@polkadot/api'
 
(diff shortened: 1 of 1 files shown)


Dudleyneedham:
The linter failed...

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -194,5 +192,66 @@ export async function associateAccountToChainArgs(
...
+  nBlocksValid = 10
+): Promise<AssociateAccountToChainResult> {
+  const api = ConfigService.get('api')
+
+  const blockNo = await api.query.system.number()
+  const validTill = blockNo.addn(nBlocksValid)
+
+  const challenge = await getLinkingChallenge(did, validTill)
+
+  const predictedType = accountAddress.length === 20 ? 'ethereum' : 'sr25519'
```
rflechtner, line 245: wait we can't just assume it's an sr25519 address can we?

rflechtner (commented):
Would we merge this before changing the challenge? Does that make sense?

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: There are only two different wrapping strategies. `ethereum` and the substrate way (`sr25519`, `ed25519`, `ecdsa`). So for anything not ethereum, we can just pretend it's one of the others. Doesn't make any difference which one we choose.

But yeah it's an ugly place in code. Don't really like it as well.

weichweich:
> Would we merge this before changing the challenge? Does that make sense?
> 
> Also, this does not change the signature of `associateAccountToChainArgs` right?

I would change the challenge, if we decide to, in another PR.
Also the signature of `associateAccountToChainArgs` should remain untouched. I only want to expose the basic building blocks.

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -125,27 +126,20 @@ function getUnprefixedSignature(
 }
 
 /**
- * Builds the parameters for an extrinsic to link the `account` to the `did` where the fees and deposit are covered by some third account.
- * This extrinsic must be authorized using the same full DID.
- * Note that in addition to the signing account and DID used here, the submitting account will also be able to dissolve the link via reclaiming its deposit!
+ * Generates the challenge that links a DID to an account.
+ * The account has to sign the raw challenge, while the DID will sign the extrinsic that contains the challenge and will
```
ntn-x2, line 130:
```suggestion
 * The account has to sign the challenge, while the DID will sign the extrinsic that contains the challenge and will
```

What is a raw challenge?

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -194,5 +192,66 @@ export async function associateAccountToChainArgs(
...
+ * `\x19Ethereum Signed Message:\n` and the length of the message.
+ *
+ * For all other key types the message will be wrapped in `<Bytes>..</Bytes>`.
+ *
+ * @param type The key type that will sign the challenge.
+ * @param challenge The challenge to proof ownership of both account and DID.
+ * @returns The wrapped challenge.
+ */
+export function getWrappedChallenge(
+  type: KeypairType,
```
ntn-x2, line 211: Why can't the type be a union of `ethereum` and `dotsama`?

packages/did/src/DidLinks/AccountLinks.chain.ts
```diff
@@ -158,33 +152,37 @@ export async function associateAccountToChainArgs(
...
+ * @param validUntil Last blocknumber that this challenge is valid for.
+ * @param signature The signature for the linking challenge.
+ * @param type The key type used to sign the challenge.
+ * @returns The arguments for the call that links account and DID.
+ */
+export async function getLinkingArguments(
+  accountAddress: Address,
+  validUntil: BN,
+  signature: Uint8Array,
+  type: KeypairType
```
ntn-x2, line 173: Same here, since the `type` parameter was introduced in this PR, as far as I can see, and is not already part of some user-exposed API.

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: So we can add new types in the future, e.g., Bitcoin, without necessarily breaking the interface.

weichweich: yeah didn't want to create a new type for that. Also i hate `dotsama`. Shouldn't that be `substrate` since there are chains outside of kusama/polkadot that are also compatible?

ntn-x2: Technically Substrate is the framework, so it would be wrong to indicate with Substrate also chains that implement the Polkadot specification using a different framework. If you don't like dotsama, which I personally do not dislike, what about polkadot?

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: I think i wanted to stress the point that the account signs the challenge, while the did signs the challenge as well, but it's inside the extrinsic

packages/did/src/DidLinks/AccountLinks.chain.ts
ntn-x2: It depends on the meaning you give to the word polkadot. I have seen Joe giving it [this](https://wiki.polkadot.network/docs/learn-polkadot-host) meaning, which would capture all networks you and me are thinking of.

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: [065949a](https://github.com/KILTprotocol/sdk-js/pull/730/commits/065949a9cdb11b8825a6467e57fd31d4f3367a6c)

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: [065949a](https://github.com/KILTprotocol/sdk-js/pull/730/commits/065949a9cdb11b8825a6467e57fd31d4f3367a6c)

weichweich: Actually I'm also not really happy with that solution

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: Here we actually need to use something that can construct a `EncodedSignature`. 

```
  const proof = { [type]: signature } as EncodedSignature
```

the type here needs to be the signature type. So `KeypairType` is quite accurate.

packages/did/src/DidLinks/AccountLinks.chain.ts
weichweich: BTW since typescript also uses duck typing, we don't break the interface if we change the type to support more options.

ntn-x2 (approved):
Beyond perhaps a suggestion to change "raw challenge" to something else, I have nothing else against it, and I am glad we caught this stuff before the next release 🚀

rflechtner (approved):
Code looks good, pls figure out why some account linking integration tests started failing with "signature unverifiable"

rflechtner:
oh and pls wait for https://github.com/KILTprotocol/sdk-js/actions/runs/4223205084 to be merged bc we're not actually testing ethereum linking rn

rflechtner (approved):
nice, let's merge this

PR #733, opened 2023-03-04
feat!: new vc export

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/2154

Refactors the vc-export package to produce credentials based on our [draft specifications](https://github.com/KILTprotocol/spec-KiltCredentialV1) for secure and fully VC compliant credentials.
Support for the old export format is dropped; this format was experimental and had a bunch of issues, using it in production is not recommended.

__This is increment 1/3__
- Implements helpers to create a KiltCredentialV1 verifiable credential object and to validate its structure
- Implements helpers to create a KiltAttestationProofV1 for such a credential and to update it when selective disclosure is applied.
- Implements functions to verify such a proof and to check its revocation status.
- Implements a preliminary form of CType verification over such a credential.
- Implements helper functions to turn an existing ICredential into a KiltCredentialV1 and KiltAttestationProofV1 so that the existing attestation remains valid for this derived credential.

_Vc-js related functionality is removed in this PR as it is reimplemented in https://github.com/KILTprotocol/sdk-js/pull/735_

~outstanding TODOs:~

- [x] Rebase after 0.31.0 release
- [x] Rethink function signatures of verify & checkStatus
- [x] Reduce code duplication
  - [x] in function internals
  - [x] CAIP-related functionality
- [x] proof creation helpers (attestation)
- [x] Improve test coverage
- [x] Maybe split up unit tests into multiple files

## How to test:

Unit tests are included.

## Checklist:


packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,633 @@
...
+  statementDigests: Uint8Array[],
+  salt: Uint8Array[]
+): Uint8Array[] {
+  return statementDigests.map((digest, index) => {
+    // initialize array with 36 + 2 + 64 bytes
+    const bytes = new Uint8Array(102)
+    // add salt to array
+    bytes.set(salt[index])
+    // add bytes 0x30 & 0x78
+    bytes.set([48, 120], 36)
```
rflechtner, line 197: Just realising that we could potentially pull these bytes into the salt, making it 38 bytes. We'd set them for existing credentials, for new ones I'd just be additional entropy

packages/vc-export/src/fromICredential.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright (c) 2018-2023, BOTLabs GmbH.
+ *
+ * This source code is licensed under the BSD 4-Clause "Original" license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
+
+import { DidUri, ICredential, ICType } from '@kiltprotocol/types'
```
ntn-x2, line 8: Should these be type imports?

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+ * @param issuer The issuer of the attestation to this credential (attester).
+ * @param chainGenesisHash Genesis hash of the chain against which the credential is verifiable.
+ * @param timestamp Timestamp of the block referenced by blockHash in milliseconds since January 1, 1970, UTC (UNIX epoch).
+ * @param ctype Optional: The CType object referenced by the [[ICredential]].
+ * @returns A KiltCredentialV1 with embedded KiltAttestationProofV1 proof.
+ */
+export function fromICredential(
+  input: ICredential,
+  issuer: DidUri,
+  chainGenesisHash: Uint8Array,
```
ntn-x2, line 336: Should this also default to Spiritnet?

packages/vc-export/src/index.ts
```diff
@@ -9,9 +9,11 @@
  * @module @kiltprotocol/vc-export
  */
 
+export * from './fromICredential.js'
+export * as KiltCredentialV1 from './KiltCredentialV1.js'
```
ntn-x2, line 13: Especially because the function with the same name is exported, would it make sense to namespace them?

packages/vc-export/src/KiltRevocationStatusV1.ts
```diff
@@ -0,0 +1,100 @@
...
+  rootHash: U8aLike
+): KiltRevocationStatusV1 {
+  const chainId =
+    typeof chainIdOrGenesisHash === 'string' &&
+    chainIdOrGenesisHash.startsWith('polkadot')
+      ? chainIdOrGenesisHash
+      : Caip2.chainIdFromGenesis(u8aToU8a(chainIdOrGenesisHash))
+
+  return {
+    id: `${chainId}/kilt:attestation/${base58Encode(rootHash)}`,
```
ntn-x2, line 97: As mentioned in another comment, the namespace should be rather something generic, like `att` (since max 8 letters can be used), and the reference could be `KiltAttestationV1` or something like that. Great to have the identifier be the base58 encoding of the root hash.

packages/vc-export/src/KiltRevocationStatusV1.ts
```diff
@@ -0,0 +1,100 @@
...
+ * If not given this function will try to retrieve a cached connection from the [[ConfigService]].
+ */
+export async function checkStatus(
+  credential: VerifiableCredential,
+  opts: { api?: ApiPromise } = {}
+): Promise<void> {
+  const { credentialStatus } = credential
+  if (credentialStatus?.type !== KILT_REVOCATION_STATUS_V1_TYPE)
+    throw new Error(
+      `credential must have a credentialStatus of type ${KILT_REVOCATION_STATUS_V1_TYPE}`
```
ntn-x2, line 40:
```suggestion
      `Credential must have a credentialStatus of type ${KILT_REVOCATION_STATUS_V1_TYPE}`
```

packages/vc-export/src/KiltRevocationStatusV1.ts
```diff
@@ -0,0 +1,100 @@
...
+    )
+  const { api = ConfigService.get('api') } = opts
+  const { assetNamespace, assetReference, assetInstance } =
+    assertMatchingConnection(api, credential)
+  if (assetNamespace !== 'kilt' || assetReference !== 'attestation') {
+    throw new Error(
+      `cannot handle revocation status checks for asset type ${assetNamespace}:${assetReference}`
```
ntn-x2, line 47:
```suggestion
      `Cannot handle revocation status checks for asset type ${assetNamespace}:${assetReference}`
```

packages/vc-export/src/KiltRevocationStatusV1.ts
```diff
@@ -0,0 +1,100 @@
...
+    )
+  }
+  if (!assetInstance) {
+    throw new Error(
+      'The CAIP-19 identifier must contain a token id decoding to the credential root hash'
```
ntn-x2, line 52:
```suggestion
      'The CAIP-19 identifier for the credential must contain an asset identifier decoding to the credential root hash'
```

packages/vc-export/src/common.ts
```diff
@@ -0,0 +1,110 @@
...
+export function delegationIdFromAttesterDelegation(
+  delegation: KiltAttesterDelegationV1
+): Uint8Array {
+  if (delegation.type !== KILT_ATTESTER_DELEGATION_V1_TYPE) {
+    throw new Error(`type must be ${KILT_ATTESTER_DELEGATION_V1_TYPE}`)
+  }
+  const match = delegationIdPattern.exec(delegation.id)
+  if (!match || !match.groups?.delegationId)
+    throw new Error(
+      `not a valid id for type ${KILT_ATTESTER_DELEGATION_V1_TYPE}: ${delegation.id}`
```
ntn-x2, line 66:
```suggestion
      `Not a valid id for type ${KILT_ATTESTER_DELEGATION_V1_TYPE}: ${delegation.id}`
```

packages/vc-export/src/types.ts
```diff
@@ -5,82 +5,133 @@
...
+
+export interface VerifiableCredential {
+  /**
+   * References to json-ld contexts defining the terms used.
+   */
+  '@context': typeof DEFAULT_CREDENTIAL_CONTEXTS
+  /**
+   * Credential identifier.
+   */
+  id: `${typeof KILT_CREDENTIAL_IRI_PREFIX}${string}`
```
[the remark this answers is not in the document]
ntn-x2: The idea is to fill this repo (https://github.com/ChainAgnostic/namespaces/tree/main) with a polkadot-specific section about chain IDs (this will most likely not change from what we have here), and assets. There was some discussion about having a namespace for generic on-chain attestations, that is why I am suggesting to not simply use `kilt` as the namespace, but rather use some KILT-derived references for this on-chain attestation namespace. For CTypes the situation is a bit different, but I guess we could define them as on-chain credential schema definitions, and define a namespace for it as well. The references would then define instructions on how to interpret the results, namely what to do with an attestation hash or a CType ID. This is all wishy-washy still, so I am not sure what's the best way to proceed here for stuff like credentials that can be persisted and can be around potentially forever unless we deprecate them.

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,633 @@
...
+        `No commitment for statement with digest ${u8aToHex(
+          digests[index]
+        )} and salt ${proof.salt[index]}`
+      )
+  })
+  // 11. Compute root hash
+  const rootHash = calculateRootHash(credential, proof)
+  // 12. Compare against credential id
+  if (credentialIdFromRootHash(rootHash) !== credential.id)
+    throw new Error('root hash not verifiable')
```
ntn-x2, line 378: I guess we want to throw credential or proof errors, right?

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,633 @@
...
+  )
+    throw new Error(
+      `block time ${new Date(
+        timestamp
+      ).toISOString()} does not match issuedAt (${credential.issuanceDate})`
+    )
+  // 17. + 18. validate federatedTrustModel items
+  await Promise.all(
+    credential.federatedTrustModel?.map(async (i) => {
+      if (i.type === KILT_ATTESTER_DELEGATION_V1_TYPE) {
```
ntn-x2, line 410: Perhaps this logic can be split into smaller chunks for easier readibility.

ntn-x2 (commented):
This is great progress, and really a lot of code 😃 I have left few questions, the most important of which rotate around how we want to identify credentials, CTypes, and delegations.

Since it's a lot of code, I can't focus on it for that long, and I left out the proof generation/verification logic to tomorrow, hope that's fine for you. So I'll give it another look and review tomorrow.

packages/vc-export/src/types.ts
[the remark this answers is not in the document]
rflechtner: > Didn't we decide in favor of using CAIP-19 identifiers for credentials, CTypes,and everything else? This would make it possible to issue public credentials to these credentials, as with the Kiltagram example. Something like `<kilt_caip_2>:att:kilt-attestation:<attestation_hash>`.

This is not in the specs right now, and I think for a good reason: credentials are not on-chain artifacts, attestations are. I use a CAIP identifier for attestations in the specs, but not for credentials; and I think doing so would be misleading. Because the data structure that is being identified simply is not on chain.
While I wouldn't want to use a CAIP id here, I'm happy to switch to a different uri pattern, as I'm not to convinced about this one. Suggestions welcome.

> Now that I think of it, it would make sense to also change what the CType ID is, since we have a breaking change we will roll out soon anyway. That would be future-proof and avoid breaking again. Then we can start issuing credentials to CTypes as well.

In theory we can change the CType id format any time, unfortunately the current ids are hashed into existing credentials though. Using a different id scheme for CTypes adhering to the meta schema would of course be possible, but it's not on the roadmap right now - and it would make the transition from the legacy meta schema to the new one way less smooth.
We can still refer to CTypes by their CAIP identifier though if we want to issue credentials to them, I see no issue with that.

packages/vc-export/src/types.ts
```diff
@@ -5,82 +5,133 @@
...
 
 export interface VerifiablePresentation {
-  '@context': [typeof DEFAULT_VERIFIABLECREDENTIAL_CONTEXT, ...string[]]
-  type: [typeof DEFAULT_VERIFIABLEPRESENTATION_TYPE, ...string[]]
+  '@context': [typeof W3C_CREDENTIAL_CONTEXT_URL, ...string[]]
+  type: [typeof W3C_PRESENTATION_TYPE, ...string[]]
   verifiableCredential: VerifiableCredential | VerifiableCredential[]
   holder?: string
-  proof: Proof | Proof[]
+  proof: Record<string, unknown> | Array<Record<string, unknown>>
```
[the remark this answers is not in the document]
rflechtner: The definition is that of a set, yes, but for the data type it's better to stick to an array I think. Sets are still rarely used in js

packages/vc-export/src/types.ts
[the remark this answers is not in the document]
rflechtner: Would be cool though if we could properly define CAIP namespaces for attestations and delegations before this goes live, so if there is a consensus forming let me know! I was also thinking that pallet and storage names could be obvious choices for asset namespace and class, but the restriction to 8 characters for the namespace makes that a little tricky. Maybe truncating helps? Anyway, these discussions should probably be had on the [specs repo](https://github.com/KILTprotocol/spec-KiltCredentialV1).

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+      type: KILT_ATTESTER_LEGITIMATION_V1_TYPE,
+    }
+    if ('credentialSubject' in legitimation) {
+      entry.verifiableCredential = legitimation
+    }
+    federatedTrustModel.push(entry)
+  })
+  if (delegationId) {
+    const delegation: KiltAttesterDelegationV1 = {
+      id: `kilt:delegation/${base58Encode(hexToU8a(delegationId))}`,
```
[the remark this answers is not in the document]
rflechtner: That is a partial CAIP identifier too; happy to redefine namespace and class as mentioned before

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+    'id',
+    'credentialSubject',
+    'issuer',
+    'issuanceDate',
+    'credentialStatus',
+    'credentialSchema',
+  ],
+}
+
+const schemaValidator = new JsonSchema.Validator(credentialSchema, '7')
```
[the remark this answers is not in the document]
rflechtner: > Maybe pull `'7'` out to a const somewhere as well?

 Actually I think we can drop the 7. The JSON schema version should be inferred from the $schema property of the schema.

> Plus, do we instantiate it outside of the function for caching? I'm not super familiar with how and when this would be called. Once the module is imported? Every time it is imported?

Yes, this is run on import, once, instead of being run every time the function is called.

packages/vc-export/src/KiltCredentialV1.ts
rflechtner: good idea. I'd have to restructure the function signature then; but I was thinking that a destructured input object makes more sense anyway, so that you have named arguments in arbitrary order

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+ * @param input.subject Did of the credential subject (claimer).
+ * @param input.claims A record of claims about the subject.
+ * @param input.cType The CType (or alternatively its id) to which the claims conform.
+ * @param input.issuer The issuer of the credential.
+ * @param input.timestamp Timestamp of a block at which the credential can be verified, in milliseconds since January 1, 1970, UTC (UNIX epoch).
+ * @param input.chainGenesisHash Genesis hash of the chain against which this credential is verifiable.
+ * @param input.claimHash Optional: digest of the credential contents needed to produce a credential id.
+ * @param input.legitimations Optional: array of credentials which function as legitimations to this credential.
+ * @param input.delegationId Optional: the id of a delegation node which was used in attesting this credential.
+ * @returns A VerfiableCredential (without proof) conforming to the KiltCredentialV1 data model. The `id` is omitted if no `claimHash` was specified.
```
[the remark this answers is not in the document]
rflechtner: exactly, if there is no proof

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,633 @@
...
+      type: 'array',
+      items: { type: 'string' },
+    },
+  },
+
+  additionalProperties: false,
+  required: ['type', 'block', 'commitments', 'salt'],
+}
+
+const schemaValidator = new JsonSchema.Validator(proofSchema)
```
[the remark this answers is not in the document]
rflechtner: ok seems like our json-schema library is a little off spec regarding choosing the right processing mode based on the credential $schema, so I better add the 7 here actually

packages/vc-export/src/index.ts
rflechtner: I don't get it, `KiltCredentialV1` _is_ the namespace?

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: let's define a namespace on the specs repo. How you like using the pallet name as a namespace (truncated to 8 letters -> `attestat`) and the storage item (`attestations`) as the reference? That could be a general approach that works for practically anything on a substrate chain.

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: 673a7a95fe712b267c66fbd3d9610f246f66c6b5

packages/vc-export/src/common.ts
rflechtner: 673a7a95fe712b267c66fbd3d9610f246f66c6b5

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: 673a7a95fe712b267c66fbd3d9610f246f66c6b5

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+    } else {
+      credentialSubject[key] = claim
+    }
+  })
+
+  const credentialSchema: JsonSchemaValidator2018 = {
+    id: cType.$id,
+    type: JSON_SCHEMA_TYPE,
+  }
+  if ('properties' in cType) {
```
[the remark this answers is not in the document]
rflechtner: refactored in 083baf66438944ddad93be306ebcd2721a5994a4

packages/vc-export/src/KiltCredentialV1.ts
rflechtner: 9a0ded49479e5153709c2e287bdb23df3ed15731

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,633 @@
...
+  credential: ICredential,
+  blockHash: Uint8Array
+): KiltAttestationProofV1 {
+  // `block` field is base58 encoding of block hash
+  const block = base58Encode(blockHash)
+  // `commitments` (claimHashes) are base58 encoded in new format
+  const commitments = credential.claimHashes.map((i) =>
+    base58Encode(hexToU8a(i))
+  )
+  // salt/nonces must be sorted by statment digest (keys) and base58 encoded
```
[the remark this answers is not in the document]
rflechtner: 673a7a95fe712b267c66fbd3d9610f246f66c6b5

packages/vc-export/src/fromICredential.ts
rflechtner: 9a0ded49479e5153709c2e287bdb23df3ed15731

packages/vc-export/src/types.ts
```diff
@@ -5,82 +5,133 @@
...
+  id: Caip2ChainId
+  type: typeof KILT_REVOCATION_STATUS_V1_TYPE
+}
+
+export interface CredentialSubject {
+  '@context': {
+    '@vocab': string
+  }
+  id: DidUri
+  [k: string]: Record<string, unknown> | string | number | boolean
```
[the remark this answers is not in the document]
rflechtner: 673a7a95fe712b267c66fbd3d9610f246f66c6b5

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+    credentialSchema.schema = cType
+  }
+
+  const federatedTrustModel: VerifiableCredential['federatedTrustModel'] = []
+  legitimations?.forEach((legitimation) => {
+    const entry: KiltAttesterLegitimationV1 = {
+      id: legitimation.id,
+      type: KILT_ATTESTER_LEGITIMATION_V1_TYPE,
+    }
+    if ('credentialSubject' in legitimation) {
```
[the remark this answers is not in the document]
rflechtner: refactored in 083baf66438944ddad93be306ebcd2721a5994a4. This was checking whether we only got a legitimation id, or the entire credential

packages/vc-export/src/KiltCredentialV1.ts
```diff
@@ -0,0 +1,365 @@
...
+    },
+    type: {
+      type: 'array',
+      uniqueItems: true,
+      minItems: 2,
+      maxItems: 2,
+      oneOf: [
+        { items: { enum: DEFAULT_CREDENTIAL_TYPES } },
+        {
+          items: {
```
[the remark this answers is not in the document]
rflechtner: 6f7c63f523ebff1ce1b09ea561b856f9c1c5c2f7

packages/vc-export/src/types.ts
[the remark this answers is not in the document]
ntn-x2: > This is not in the specs right now, and I think for a good reason: credentials are not on-chain artifacts, attestations are. I use a CAIP identifier for attestations in the specs, but not for credentials; and I think doing so would be misleading.

Right, probably it would be wrong to use the `att` namespace, but we could perhaps use the `cred` namespace to identify generic verifiable credentials, in which we would fit current KILT credentials, new VC-compatible credentials, and public credentials we well (?)

> We can still refer to CTypes by their CAIP identifier though if we want to issue credentials to them, I see no issue with that.

That sounds like a good compromise to me!

packages/vc-export/src/KiltCredentialV1.ts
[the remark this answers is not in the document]
ntn-x2: I see it used in a few places. Can we pull it out in a const then?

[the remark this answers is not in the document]
ntn-x2: Or perhaps some sort of builder that would return the right object given the schema definition.

packages/vc-export/src/index.ts
ntn-x2: Nevermind, I have probably been mistaken.

packages/vc-export/src/KiltRevocationStatusV1.ts
ntn-x2: I don't think that is the right direction. How would the pallet name be defined? By its crate name? But its runtime declaration? And then each chain would basically have its own namespace, while that should be left to references and not namespaces. I am in talks with Juan that manages the CAIP stuff, so let's hold on on this for a while longer. It might either make sense to have a generic namespace for each of ctype, attestation, delegation, and credential, or then a `kilt` namespace which would give us the freedom to define whatever within it.

packages/vc-export/src/KiltRevocationStatusV1.ts
ntn-x2: Token ID applies to assets that are tokens. CAIP-19 is a generic identifier structure for assets, with a namespace, a references, and an identifier. That's why I made the suggestion, since token_id could be misleading in the context of a credential.

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+  cTypeId: ICType['$id']
+  delegationId: IDelegationNode['id'] | null
+}> {
+  const { api = ConfigService.get('api') } = opts
+  const apiAt = await api.at(blockHash)
+  // TODO: should we look up attestation storage as well and take attestation info from there? That gives us the definitive attestation state in this block (e.g. makes sure it hasn't been removed after)
+  const [events, time] = await apiAt.queryMulti<
+    [Vec<FrameSystemEventRecord>, u64]
+  >([
+    [api.query.system.events as QueryableStorageEntry<'promise'>],
```
ntn-x2, line 263: Wouldn't it be easier to simply query the attestation storage instead? What was the reason for parsing the block events instead?

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+ * @returns Object indicating whether proof could be verified.
+ */
+export async function verifyProof(
+  credentialInput: VerifiableCredential,
+  proof: KiltAttestationProofV1,
+  opts: { api?: ApiPromise } = {}
+): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { proof: _, ...credential } = credentialInput
+  // 0. check proof structure
```
ntn-x2, line 344: Why don't we check the credential structure first? Is it more efficient to do so?

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+
+/**
+ * Initialize a new, prelimiary [[KiltAttestationProofV1]], which is the first step in issuing a new credential.
+ *
+ * @example
+ * const [proof, args] = initializeProof(credential)
+ * const tx = api.tx.attestation.add(...args)
+ * await tx.signAndSend(signerKeypair)
+ *
+ * @param credential A KiltCredentialV1 for which a proof shall be created.
```
ntn-x2, line 552:
```suggestion
 * @param credential A [[KiltCredentialV1]] for which a proof shall be created.
```

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+ * ```
+ * const { proof, credential } = deriveProof(
+ *  originalCredential,
+ *  originalProof,
+ *  ['name', 'address']
+ * )
+ * const derivedCredential = { ...credential, proof }
+ * ```
+ */
+export function deriveProof(
```
ntn-x2, line 497: I would rename this function so that it is not super confusing with the other ones, particularly `createProof`. Maybe something like `createPresentation`?

packages/vc-export/src/CredentialSchema.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Validates the claims in the VC's `credentialSubject` against a CType definition on the `credentialSchema` property.
+ *
+ * @param credential A verifiable credential where `credentialSchema.schema` is an [[ICType]].
+ * @param credential.credentialSubject The credentialSubject to be validated.
+ * @param credential.credentialSchema The credentialSchema to be applied.
+ * @param cType Optionally pass the CType definition to be used if it is not embedded in the credentialSchema.
+ */
+export function validateSchema(
```
ntn-x2, line 21: Because it's exported under its own namespace, would it make sense to simply rename it to `validate`?

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+
+function makeCommitments(
+  statementDigests: Uint8Array[],
+  salt: Uint8Array[]
+): Uint8Array[] {
+  return statementDigests.map((digest, index) => {
+    // initialize array with 36 + 2 + 64 bytes
+    const bytes = new Uint8Array(102)
+    // add salt to array
+    bytes.set(salt[index])
```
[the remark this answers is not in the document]
ntn-x2: Should we at the very list truncate to the length we expect? In that case it would still be deterministic, but we have to specify it cannot be less than 36 bytes long.

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+ * This includes querying the KILT blockchain with the credential id, which returns an attestation record if attested.
+ * This record is then compared against attester address and delegation id (the latter of which is taken directly from the credential).
+ *
+ * @param credentialInput Verifiable Credential to verify proof against.
+ * @param proof KiltAttestationProofV1 proof object to be verified. Any proofs embedded in the credentialInput are stripped and ignored.
+ * @param opts Additional parameters.
+ * @param opts.api A polkadot-js/api instance connected to the blockchain network on which the credential is anchored.
+ * @returns Object indicating whether proof could be verified.
+ */
+export async function verifyProof(
```
ntn-x2, line 337: Should this function also be renamed to `verify` only?

packages/vc-export/src/KiltRevocationStatusV1.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Check attestation and revocation status of a credential at the latest block available.
+ *
+ * @param credential The KiltCredentialV1 to which the status method is linked to.
+ * @param opts Additional parameters.
+ * @param opts.api An optional polkadot-js/api instance connected to the blockchain network on which the credential is anchored.
+ * If not given this function will try to retrieve a cached connection from the [[ConfigService]].
+ */
+export async function checkStatus(
```
ntn-x2, line 34: Perhaps simply `check`?

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+  return {
+    verified: true,
+    timestamp,
+    attester,
+    cTypeId,
+    delegationId,
+  }
+}
+
+async function findDelegators(
```
ntn-x2, line 306: Because it returns a `Promise<void>`, perhaps `find` is not the best term. What about `checkDelegation` or `checkDelegationHierarchy`?

ntn-x2 (commented):
Ok review round 2, focused on the proof stuff. I think I will never be able to grasp the whole flow given how much new code there is, but in general, there are few points I would like to make:

1. Perhaps you could go over the usage API again, and rename functions where it makes sense. For instance the verification function for a proof could simply be called `verify` since it's already namespaced. I left comments in few places, but might have also missed few other places.
2. It would be nice, for the attestation stuff, to simply have a `create` and a `verify` function, so that people would not misuse its API. I think what is exposed right now might be a bit too coarse.
3. I see that in a lot of places we always do a lot of base58 and hex encoding and decoding. Would it make sense and would it be possible to leave all those crypto details as byte array, and perhaps simply override the `toString` method for when the credential and its proof are printed? Or is this format required by other library, e.g., `vc-js`?

In general the test cases look good to me, so it's good enough for me. I will approve it once my points have been addressed, but I am really looking forward to seeing this integrated into our product line and beyond 🚀🚀🚀

rflechtner:
> 1. Perhaps you could go over the usage API again, and rename functions where it makes sense. For instance the verification function for a proof could simply be called `verify` since it's already namespaced. I left comments in few places, but might have also missed few other places.

Fair point, especially w/r/ to verifyProof. Will take another look at that. 

> 2. It would be nice, for the attestation stuff, to simply have a `create` and a `verify` function, so that people would not misuse its API. I think what is exposed right now might be a bit too coarse.

I realised I forgot did signing on createProof. But besides that I'm not sure simplifying the proof issuance api is practical. As long as creating a proof requires a write operation to the blockchain, it's going to be complicated. Especially as we need to know the inclusion block as well. I'd prefer to continue exposing the functions performing the steps before and after this write operation.

> 3. I see that in a lot of places we always do a lot of base58 and hex encoding and decoding. Would it make sense and would it be possible to leave all those crypto details as byte array, and perhaps simply override the `toString` method for when the credential and its proof are printed? Or is this format required by other library, e.g., `vc-js`?

The proof for a vc is always going to be a JSON object, so I see no way around (de)serializations in a lot of places. I can double check the internals though.

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: It's '0x' in UTF-8 ;) comes from the hash prefix. A result of the requirement that we want backwards compatibility

packages/vc-export/src/types.ts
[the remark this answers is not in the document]
rflechtner: > Right, probably it would be wrong to use the `att` namespace, but we could perhaps use the `cred` namespace to identify generic verifiable credentials, in which we would fit current KILT credentials, new VC-compatible credentials, and public credentials we well (?)

Mhh as I said it feels inappropriate to me to use a CAIP at all if we're trying to identify off-chain credentials. For public credentials it kind of makes sense to me (even though strictly speaking, as is also the case with CTypes, you seem to be identifying primarily the "pointer" to the actual data of which the credential is composed, because the latter is not in state). But I'd resist trying to CAIPify stuff that does not live on a blockchain.

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: Yeah the terminology in CAIP-19 may not be ideal. Have another look at https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md#syntax-1 - if we use the term 'asset identifier', it's easy to understand that referring to the asset_id in CAIP-19 terminology, but it's not that; it's what they call the token_id.

packages/vc-export/src/KiltAttestationProofV1.ts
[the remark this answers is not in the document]
rflechtner: Truncating is in fact what would happen if it were too long; feels wrong though? The specs define this as being exactly 36 bytes long, but we could also take whatever length to be honest and simply concatenate

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: The attestation storage does not tell us when the attestation happened. Therefore we check the event, and check the storage only when checking revocation status.

packages/vc-export/src/KiltAttestationProofV1.ts
```diff
@@ -0,0 +1,668 @@
...
+  if (!valid)
+    throw new ProofMalformedError(
+      `Object not matching ${ATTESTATION_PROOF_V1_TYPE} data model`,
+      {
+        cause: errors,
+      }
+    )
+}
+
+export interface VerificationResult {
```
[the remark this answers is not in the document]
rflechtner: 06f6e12c42cc00da85af1ba03bf55251480af646

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: Ah now I see what you mean! Alright, I think you'll like this: 1b0f5389163fe4beec78c7c40de06a7b72f764f1
(also realised that I wasn't properly checking the delegation node id in case there was no delegation-related item in the federatedTrust set)

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: Conceptually it makes sense to me; the proof needs to make assertions about its own structure; if it's not internally consistent, there is no point in checking whether it proves the integrity of the document for which it was created. And yes, it's probably also more efficient that way, as the proof is just a way smaller structure and thus cheaper to check.

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: `createPresentation` would be _very_ confusing, given that creating a presentation for a VC usually means creating (and signing) a Verifiable Presentation. How about `updateProof` ?

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: Check... what? A proof? A credential? Where is the difference to `verify`? I'd rather go more specific: `checkRevocationStatus`?

rflechtner: Oh now I see it, the namespace should explain it... yeah I guess that works then?

packages/vc-export/src/CredentialSchema.ts
rflechtner: Maybe `validateSubject` because the schema is assumed not to validate the whole credential; it just checks the credentialSubject.

packages/vc-export/src/KiltAttestationProofV1.ts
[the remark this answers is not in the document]
rflechtner: b065621b0ee23255eb9f4280847ba84ffff27736

packages/vc-export/src/KiltAttestationProofV1.ts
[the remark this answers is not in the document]
rflechtner: 06f6e12c42cc00da85af1ba03bf55251480af646

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: 3f07bb85f8c70f4d3d4f056426ec1849131440ad

packages/vc-export/src/CredentialSchema.ts
rflechtner: fba870433efcd82c894e2a95cea764a4620e8bbc

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/commit/f2f6147b0c068b48e3b936f1f5b99764d5ab53f9

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: 62ee6471494952e15a63b068f6f1ee48dfd802bc

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: 1b0f5389163fe4beec78c7c40de06a7b72f764f1

packages/vc-export/src/KiltAttestationProofV1.ts
ntn-x2: I would say that `update` is also misleading, as it returns a new object and does not change the input one.
When would this function be called outside of presentations?

packages/vc-export/src/types.ts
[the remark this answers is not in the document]
ntn-x2: > Mhh as I said it feels inappropriate to me to use a CAIP at all if we're trying to identify off-chain credentials.

I see. Perhaps you are misguided by your assumption of what a CAIP should do. But in theory it's Chain Agnostic, meaning that it could also be used for stuff that does not necessarily live on a blockchain. But this is still an open point I am also discussing with others, so we can let it be for now.

> even though strictly speaking, as is also the case with CTypes, you seem to be identifying primarily the "pointer" to the actual data of which the credential is composed, because the latter is not in state

An asset could be defined as a combination of on-chain and off-chain information, hence I would think it's totally viable to define a CType asset as a mix of on-chain and off-chain information, although off-chain is not REALLY offchain as can be an IPFS link, but rather still conceptually quite tied to the blockchain itself.

packages/vc-export/src/KiltCredentialV1.ts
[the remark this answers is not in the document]
ntn-x2: You're right, I did not see that is a partial one. Let's stick with this definition for now then.

packages/vc-export/src/KiltCredentialV1.ts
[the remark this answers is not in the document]
ntn-x2: > This value always depends on the schema version used by the schema defined in the same file; a global const seems misplaced here.

Is there any occurrence in which, within the same file, you might want to create definitions with different versions? I did not mean a global constant, but just a way to set it once, and simply use it, even within the same file.

packages/vc-export/src/KiltRevocationStatusV1.ts
ntn-x2: I see, still talking about tokens in this context also seems less than ideal for me 😁 Can we use another phrasing?

ntn-x2:
> I realised I forgot did signing on createProof. But besides that I'm not sure simplifying the proof issuance api is practical. As long as creating a proof requires a write operation to the blockchain, it's going to be complicated. Especially as we need to know the inclusion block as well. I'd prefer to continue exposing the functions performing the steps before and after this write operation.

Alright, but we need to make sure to properly advertise what the entrypoint is, if someone simply wants to create or verify a signature. 

> The proof for a vc is always going to be a JSON object, so I see no way around (de)serializations in a lot of places. I can double check the internals though.

I see, so it does not make to have an `export` function that stringifies all the stuff we would handle as byte arrays internally, ok. But yeah, perhaps there could be few places where we could avoid doing unnecessary (de/en)coding.

packages/vc-export/src/KiltAttestationProofV1.ts
rflechtner: Practically never, but the result is not a presentation. We could also call it `selectiveDisclosure`? Guess that's closest to what it does.

ntn-x2: I think the term proof is then the most accurate, since that's what it is, indeed, which is then used in presentations. So I guess `deriveSelectiveDisclosureProof` would work, not sure if it's too long. Also `deriveSDProof` might work.

packages/vc-export/src/KiltRevocationStatusV1.ts
rflechtner: this discussion should move to https://github.com/KILTprotocol/spec-KiltCredentialV1/issues/7

ntn-x2 (approved):
🚀🚀🚀

PR #740, opened 2023-03-16
feat: enums and arrays on new ctype metaschema

rflechtner:
## fixes KILTProtocol/ticket#2456

Amends the new CType metaschema with new features:
- Allowable values for properties of string, number & integer types can be restricted with an enum validator.
- Arrays are allowed, with all items validating against a schema normally used to validate simple properties (i.e., boolean, string, or number, schema, or a CType reference).
- Length assertions are allowed for arrays and strings.
- Maximum and minimum value assertions are allowed for numbers and integers.

## How to test:

It's easiest to play with the metaschema (which you can also download from https://gateway.pinata.cloud/ipfs/bafybeiah66wbkhqbqn7idkostj2iqyan2tstc4tpqt65udlhimd7hcxjyq) by validating your desired CTypes against it on https://www.jsonschemavalidator.net.

## Checklist:

### DIFF
--- a/packages/core/src/ctype/CType.schemas.ts
+++ b/packages/core/src/ctype/CType.schemas.ts
@@ -9,7 +9,7 @@ import { JsonSchema } from '@kiltprotocol/utils'
 
 export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
   // $id is not contained in schema when fetched from ipfs bc that is impossible with a content-addressed system
-  $id: 'ipfs://bafybeifzfxz6tfd2xo7ijxbfceaxo3l655yg7sovlsnpxgq2rwfl4kbfgm',
+  $id: 'ipfs://bafybeiah66wbkhqbqn7idkostj2iqyan2tstc4tpqt65udlhimd7hcxjyq',
   $schema: 'http://json-schema.org/draft-07/schema#',
   title: 'CType Metaschema (V1)',
   description:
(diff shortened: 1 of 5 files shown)


packages/core/src/ctype/CType.schemas.ts
```diff
@@ -66,6 +49,89 @@ export const CTypeModelV1: JsonSchema.Schema & { $id: string } = {
...
+    array: {
+      additionalProperties: false,
+      properties: {
+        type: { const: 'array' },
+        items: {
+          oneOf: [
+            { $ref: '#/definitions/string' },
+            { $ref: '#/definitions/number' },
+            { $ref: '#/definitions/boolean' },
+            // TODO: do we have reasons not to allow CType references in arrays?
```
ntn-x2, line 121: Do we have reasons not to allow arrays and objects inside arrays?

ntn-x2 (approved):
Looks good to me! Just a question: I guess this definition does not allow for nested object definitions, right?

packages/core/src/ctype/CType.schemas.ts
ntn-x2: Ok, so we just allow for 1 layer of nesting at the CType level, and anything else would be split up into nested CType structures, correct?

rflechtner:
Exactly

packages/core/src/ctype/CType.schemas.ts
rflechtner: Exactly

rflechtner, line 121: Remove TODO
```suggestion
```

PR #743, opened 2023-03-30
feat: DIP provider type definitions

ntn-x2:
~Contains all PRs related to DIP, and will be merged at the end of the development iteration for DIP v1.~

~Companion of https://github.com/KILTprotocol/kilt-node/pull/494.~

Most of the features are under migration to https://github.com/KILTprotocol/dip-sdk.
This PR only contains the type definitions for the DIP provider stuff as exposed in https://github.com/KILTprotocol/kilt-node/pull/494.


packages/did/src/DidDetails/FullDidDetails.ts
```diff
@@ -40,6 +40,7 @@ const methodMapping: Record<string, VerificationKeyRelationship | undefined> = {
   'did.submitDidCall': undefined,
   didLookup: 'authentication',
   publicCredentials: 'assertionMethod',
+  dipProvider: 'authentication',
```
rflechtner, line 43: unrelated, but do you think this configuration could be moved into runtime constants at some point, either on the pallet level or in the did pallet?

packages/type-definitions/src/index.ts
```diff
@@ -119,9 +122,13 @@ const defaultTypesBundle: OverrideVersionedType[] = [
...
   },
   {
-    minmax: [10900, undefined],
+    minmax: [10900, 11199],
     types: types10900,
   },
+  {
+    minmax: [11200, undefined],
+    types: types11200,
+  },
```
rflechtner, line 131: do we know for sure this is going to be the runtime version with which this is released?

rflechtner (commented):
can you also run type generation and commit the results so we know this still works?

packages/did/src/DidDetails/FullDidDetails.ts
ntn-x2: You mean exposed as part of the metadata?

packages/type-definitions/src/index.ts
ntn-x2: No, we don't. The current release of KILT won't have the provider pallet deployed. The provider template runtime does. Maybe I should move these definitions to the new repo, and add it here only when it's going to be included? Otherwise it's really difficult to test this.

packages/did/src/DidDetails/FullDidDetails.ts
rflechtner: Yeah that's probably how these are exposed then. If we found a solution for this chain-side, we wouldn't have to duplicate it client-side, making the logic so much more versatile

packages/type-definitions/src/index.ts
rflechtner: Mhh yeah maybe this is the way then.

packages/did/src/DidDetails/FullDidDetails.ts
ntn-x2: I mean, in this case I am not sure it is worth the roundtrip and load on a fullnode to derive this information. Better would be to have a WASM blob, which we tried and failed earlier this year.

ntn-x2: Unless we can expose it in a constant way via metadata, but I am not sure that is possible. Will make a ticket to investigate that at some point.

ntn-x2:
Putting this on hold until we have a version of our chain that has the provider pallet deployed. Until then, all type augmentation and runtime decoration will happen in the integration tests of the [`@kiltprotocol/dip-sdk`](https://github.com/KILTprotocol/dip-sdk) repo.

packages/did/src/DidDetails/FullDidDetails.ts
rflechtner: Yes, check it out, but I would guess that the constant values are exposed via the metadata and thus available as soon as the metadata has been loaded - I don't think this puts any additional load on the node

packages/type-definitions/src/index.ts
ntn-x2: Metadata V15 is not out yet. V14 only exposes stuff that is directly used by the runtime. The only way to make it work would be to expose a config pallet which exposes to metadata whatever stuff it is passed to its `Config` trait. Not sure what the downsides of that would be. For one definitely WASM size.

PR #745, opened 2023-04-03
ci: update GitHub workflows

rflechtner:
## re/ KILTProtocol/ticket#2264

Updates GitHub workflows to no longer use deprecated actions and features which will be retired shortly.

## How to test:

Run as many workflows as possible I guess.

## Checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -139,7 +139,7 @@ jobs:
         run: node --version
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v1
+        uses: aws-actions/configure-aws-credentials@v2
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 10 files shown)


.github/workflows/tests.yml
```diff
@@ -123,22 +124,20 @@ jobs:
     continue-on-error: ${{ matrix.required == 'optional' }}
 
     steps:
-      - name: Log out node version
-        run: node --version
       - name: Use Node.js 14
         uses: actions/setup-node@v3
         with:
           node-version: 14
```
ggera, line 130:
```suggestion
          node-version: 16
```

ggera, line 130:
```suggestion
          node-version: 16
```

.github/workflows/tests.yml
```diff
@@ -175,15 +174,15 @@ jobs:
...
```
ggera, line 176:
```suggestion
          node-version: 16
```

ggera (approved):
is the `node-version` upgraded accross the board? if so there are places where it's still 14

rflechtner:
> is the `node-version` upgraded accross the board? if so there are places where it's still 14

I haven't actually updated the node version of our build/test jobs, just the actions that were still using node v12. I'd do that in a separate PR so it's easier to revert if something breaks.

PR #754, opened 2023-05-17
feat: cbor-web replaces cbor

arty-name:
https://github.com/KILTprotocol/ticket/issues/2687


packages/core/package.json
```diff
@@ -50,6 +50,7 @@
     "@polkadot/types": "^10.0.0",
     "@polkadot/util": "^11.0.0",
     "@polkadot/util-crypto": "^11.0.0",
-    "cbor": "^8.1.0"
+    "cbor": "^8.1.0",
```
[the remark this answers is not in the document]
arty-name: Yeah, there’s more cleanup in order

packages/did/src/cbor-web.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'cbor-web'
```
rflechtner, line 1: I'm debating - is an empty declaration file better than a `@ts-expect-error` in front of the import? The result is the same - it's treated as any

rflechtner: I don't think that's the solution, this way we'll still keep the dependency on cbor - extracting type definitions from another package and adding it to your build is not really what tsc is made for I think

arty-name: [Here’s a thing by the way](https://github.com/hildjj/node-cbor/commit/bb3787a11f7ad6d8508574b91e293be825765430)

arty-name: [And its discussion](https://github.com/hildjj/node-cbor/pull/177)

PR #755, opened 2023-05-23
fix: make config package cjs-only   

rflechtner:
## fixes KILTProtocol/ticket#2717

The config package is build mainly as cjs, with only an additional index.mjs simplifying imports from esm. This is to make sure that only one file holds all state, so that any settings that are made are respected by all imports of the package, as long as there is no package duplication.

## How to test:

This issue prevented using the sdk as a transitive dependency of a package that was built as cjs in a project that was esm based (sporran). I verified that with this change, this is now possible.

## Checklist:

    * Link the PR/Ticket here


rflechtner:
> The approach looks a little backwards to me, but not having the capacity to fix Sporran right now, I approve it

It's not just about sporran - any esm consumer of the kilt extension library would have issues with this, and I suspect the issue could come back to bite us in the ass at any time. Better to fix the root cause once and for all I'd say.

arty-name:
While other consumers could have the same misconfiguration as Sporran, they could also be fixed, I assume

rflechtner:
Is using ESM as a module system 'a misconfiguration'? I see the entire ecosystem shifting towards it, seems like an odd characterisation to me

arty-name:
Configuring the bundler in a way that it loads the same package twice in different versions is a misconfiguration. This is just webpack behaviour, and another bundler is likely to behave differently. It’s on a totally different level than Node.js stuff. Webpack should have only taken the ESM version.

rflechtner:
> Configuring the bundler in a way that it loads the same package twice in different versions is a misconfiguration. This is just webpack behaviour, and another bundler is likely to behave differently. It’s on a totally different level than Node.js stuff. Webpack should have only taken the ESM version.

That is true, if you are not using a bundler though you would get the same behaviour. Node.js would follow the esm path from an esm dependency and the cjs path from a cjs dependency, and thus also import two versions of the same package.

arty-name:
I don’t think we have a CJS dependency for node yet, and if one is in the works, it could be done ESM-compatible, I guess. This would be future-proof.

arty-name:
By the way, check out the reason [Sporran build fails for the new SDK](https://github.com/BTE-Trusted-Entity/sporran-extension/pull/1981)

arty-name:
I have not investigated the exact root cause, but I think this showcases that the benefits of CJS-only approach are not as clear-cut as expected

rflechtner:
I mean it makes sense, if the problem was before that your bundler couldn't deduplicate the ESM/CJS versions of the config package, it's now probably failing to deduplicate the imports of the config package. At least it doesn't break now, but there's a ton of dead code added. I could remove dependencies of the package that we don't really need, if you think that helps?

arty-name:
I need to investigate it first

PR #758, opened 2023-05-24
chore: release 0.33.0

rflechtner:
### Features

* feat!: enums and arrays on new ctype metaschema by @rflechtner in https://github.com/KILTprotocol/sdk-js/pull/740
* feat!: fetch attestation as part of verifyCredential by @rflechtner in https://github.com/KILTprotocol/sdk-js/pull/708
* feat: cbor-web replaces cbor by @arty-name in https://github.com/KILTprotocol/sdk-js/pull/754

### Bugfixes

* fix!: current CType meta schema ignores additional properties in claim contents by @rflechtner in https://github.com/KILTprotocol/sdk-js/pull/726
* fix: integration test error matching by @rflechtner in fea185fa41f1c8d7b1aa4cc4bbaa30e4c7432cd0
* fix: augmented type generation by @rflechtner in #747

### Other

* chore(deps)!: update polkadot/common dependencies to v12 by @renovate & @rflechtner in https://github.com/KILTprotocol/sdk-js/pull/749
* refactor!: remove kilt dependencies from config package by @rflechtner in https://github.com/KILTprotocol/sdk-js/pull/763
* chore(deps): update polkadot dependencies by @renovate in https://github.com/KILTprotocol/sdk-js/pull/732
* chore: update metadata to 11000 by @rflechtner in 5920dce977e4c7e16c88fcd11278d95ebddcbed2
* test: breaking changes by @tjwelde in https://github.com/KILTprotocol/sdk-js/pull/642


rflechtner:
Unfortunately the `polkadot/common` libraries were bumped again within a minor release of the `polkadot/api` libs, which means that some apps (I had the issue in sporran) will have to use `resolutions` to avoid dependency duplication. That sucks; the alternative would be to pull in https://github.com/KILTprotocol/sdk-js/pull/749. This however ends support for node 14, and leads to some interface changes on the keypair object.

arty-name (approved):
The topics I’m familiar with look good

ntn-x2 (commented):
LGTM as well, as far as my feedback is valuable.

rflechtner:
> @rflechtner we [bumped the number of URLs per service to 2](https://github.com/KILTprotocol/kilt-node/pull/517), so you need to update the integration tests to use 3 instead of 2.

Depending on which runtime is used? well that's going to be fun

ntn-x2:
> Depending on which runtime is used? well that's going to be fun

No I mean for the tests. Setting it to 3 will always fail.

rflechtner:
> > Depending on which runtime is used? well that's going to be fun
> 
> No I mean for the tests. Setting it to 3 will always fail, which is what we want. 2 does not **always** fail anymore.

I see. The deposit test are also breaking though, so there's more.

PR #762, opened 2023-05-31
feat: export imported types

rflechtner:
## fixes KILTProtocol/ticket#2697

Re-exports all the types we import in the types package.

## Checklist:

    * Link the PR/Ticket here


packages/types/src/Address.ts
```diff
@@ -5,10 +5,9 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import type { Prefix } from '@polkadot/util-crypto/address/types'
-import type { HexString } from '@polkadot/util/types'
 import '@polkadot/keyring' // TS needs this for the augmentation below
-import type { KeyringPair } from './index.js'
+
+import type { HexString, KeyringPair, Prefix } from './index'
```
arty-name, line 10: Should this be consistent with the rest?
```suggestion
import type { HexString, KeyringPair, Prefix } from './Imported'
```

arty-name (approved):
Should other SDK packages import these types from the types package?

packages/types/src/Address.ts
rflechtner: good spot, I wonder if that happened due to the merge. fixed in 0f58cd5df20f67db2b300a3a0e28f2a16762fe25

rflechtner:
> Should other SDK packages import these types from the types package?

Good point!

PR #765, opened 2023-06-08
chore: update metadata

rflechtner:
## Checklist:

    * Link the PR/Ticket here


packages/core/src/__integrationtests__/Did.spec.ts
```diff
@@ -54,11 +54,6 @@ beforeAll(async () => {
   paymentAccount = await createEndowedTestAccount()
 }, 30_000)
 
-it('fetches the correct deposit amount', async () => {
```
ntn-x2, line 57: Do you want to replace this with the new base, key, and service deposits?

rflechtner: and switch between the new and the old based on the current runtime? mhhh no thanks :D

rflechtner: I started with that idea in mind but quickly abandoned it after realising that the tests agains the current release would then fail

rflechtner: I mean how would we know that we still support the current release then?

ntn-x2: When we open a PR with `master`. I don't remember we always cared about keeping `develop` in sync with the latest release. If that's the case, feel free to disregard this, since I don't have full memory of how we dealt with this in the past.

rflechtner: the CI will fail afaik if stuff does not with the current release, and generally I feel we shouldn't go there

PR #767, opened 2023-06-09
feat: vite SSR cbor-web workaround

arty-name:
The bundler vite is unhappy when an SSR script attempts a named import from the CJS library cbor-web.

## How to test:

Try to build [this branch](https://github.com/galaniprojects/ctypehub/pull/97).

```
SyntaxError: Named export 'decode' not found. The requested module 'cbor-web' is a CommonJS module, which may not support all module.exports as named exports.
```

I have done a cursory test of the proposed fix in our projects and it seems to work for all the tools we have.

## Checklist:

    * Link the PR/Ticket here


rflechtner (approved):
That makes sense - ESM's named imports are not possible when the dependency is a standard CJS module.

PR #772, opened 2023-06-14
fix: cbor-web imports do not work in esm

rflechtner:
## fixes KILTProtocol/ticket#2749

Fixes a broken import in esm code. The fix is ugly, but it's actually what a recent post on dual-esm-cjs libraries recommends (https://evertpot.com/universal-commonjs-esm-typescript-packages/)  

## How to test:

I tried importing the relevant functions in both an mjs and a cjs script, that worked.

## Checklist:

    * Link the PR/Ticket here


rflechtner:
@arty-name what I could do to make this a little nicer is move that import magic to a file in the utils package, and then import from there?

arty-name (approved):
Yeah, I guess a separate file in utils will be a nice abstraction

rflechtner:
> Yeah, I guess a separate file in utils will be a nice abstraction

Hope that's better?

PR #779, opened 2023-07-21
test: esm variant of integration tests

rflechtner:
## fixes KILTProtocol/ticket#2753

Cleans up testing infrastructure and sets up integration tests to be run against both the cjs and esm variants of the code.
Includes:
- updating jest to latest version
- moving integration tests, breaking changes tests and test utils to `tests` folder
- stop using jest-runner-groups in favour of separate test configurations

## How to test:

Tests are run in CI and should complete successfully.

## Checklist:

    * Link the PR/Ticket here


jest.config.js
```diff
@@ -15,37 +21,43 @@ module.exports = {
...
-  rootDir: 'packages',
-  coverageDirectory: 'coverage',
-  moduleDirectories: [
-    "node_modules",
-    "packages/*/src"
-  ]
 }
+
+module.exports = {
+  ...common,
```
arty-name, line 49: Is common needed here?

jest.config.js
```diff
@@ -15,37 +21,43 @@ module.exports = {
...
+      displayName: 'unit',
+      roots: ['<rootDir>/packages'],
+    },
+    {
+      ...common,
+      displayName: 'breaking',
+      roots: ['<rootDir>/tests/breakingChanges'],
+    },
+  ]
+}
```
arty-name, line 63: Will the project benefit from an `.editorconfig` file?

arty-name (approved):
So much noise from reordered imports and a different snapshot format :/

jest.config.js
rflechtner: looks like it is not well supported in VSCode, which most of the team uses (https://asp.net-hacker.rocks/2020/01/13/editorconfig.html), but it can't hurt either I assume. I did format it though?

jest.config.js
rflechtner: it is currently hard to see which setting is expected in which context, I had a lot of issues with jest ignoring some settings. This avoids them by defining all options both within the projects context and the outer context.

rflechtner:
> So much noise from reordered imports and a different snapshot format :/

Yeah sorry for these. I could have avoided some of the noise from the reordering, but these are files where I had to touch the imports anyway, so I figured I may as well leave them cleaned up...

jest.config.js
arty-name: Weird how Jest does that :/

PR #781, opened 2023-07-28
refactor!: remove messages

weichweich:
## Checklist:

    * Link the PR/Ticket here


rflechtner (commented):
LGTM, just one little request: Given that we do not know for sure whether we want to get rid of Quotes, and given they are fairly independent of messages, can you maybe split this PR in two, the first of which removes Messaging, and the second of which gets rid of Quotes based on that? This makes it easier for us to revert if we figure out we want to keep quotes after all.

Ad96el:
If I remember correctly, yesterday we decided to keep the quotes but relocate them into the kilt extension API as it is currently implemented. Is the request still valid?

rflechtner (approved):
Ok, re-adding quotes wouldn't be too hard in any case, let's do this!

PR #786, opened 2023-08-09
feat: legacy credentials package

rflechtner:
## addresses https://github.com/KILTprotocol/ticket/issues/2156

Adds a package for legacy kilt credential support, which also becomes the new home for transformations between VCs and legacy credentials.

## How to test:

Tests included. Doesn't hurt to try exporting a credential with it though.

## Checklist:

    * Link the PR/Ticket here


packages/legacy-credentials/src/vcInterop.ts
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * Produces an instance of [[KiltAttestationProofV1]] from an [[ICredential]].
+ *
+ * @param credential Input credential.
+ * @param opts Additional parameters required for creating a proof from an [[ICredential]].
+ * @param opts.blockHash Hash of a block at which the proof must be verifiable.
+ * @returns An embedded proof for a verifiable credential derived from the input.
+ */
+export function proofFromICredential(
```
ntn-x2, line 32: Maybe?
```suggestion
export function vcProofFromICredential(
```

packages/legacy-credentials/src/vcInterop.ts
```diff
@@ -0,0 +1,180 @@
...
+ * Transforms an [[ICredential]] object to conform to the KiltCredentialV1 data model.
+ *
+ * @param input An [[ICredential]] object.
+ * @param options Additional required and optional parameters for producing a VC from an [[ICredential]].
+ * @param options.issuer The issuer of the attestation to this credential (attester).
+ * @param options.timestamp Timestamp of the block referenced by blockHash in milliseconds since January 1, 1970, UTC (UNIX epoch).
+ * @param options.chainGenesisHash Optional: Genesis hash of the chain against which this credential is verifiable. Defaults to the spiritnet genesis hash.
+ * @returns A KiltCredentialV1 with embedded KiltAttestationProofV1 proof.
+ */
+export function vcFromICredential(
```
ntn-x2, line 65: I guess we can simply avoid exporting `*FromICredential`, right? I haven't seen those imported anywhere else, and anyway they are not exported from the index.

packages/legacy-credentials/src/vcInterop.ts
```diff
@@ -0,0 +1,180 @@
...
+ * @param opts Additional required and optional parameters for producing a VC from an [[ICredential]].
+ * @param opts.issuer The issuer of the attestation to this credential (attester).
+ * @param opts.blockHash Hash of any block at which the credential is verifiable (i.e. Attested and not revoked).
+ * @param opts.timestamp Timestamp of the block referenced by blockHash in milliseconds since January 1, 1970, UTC (UNIX epoch).
+ * @param opts.chainGenesisHash Optional: Genesis hash of the chain against which this credential is verifiable. Defaults to the spiritnet genesis hash.
+ * @returns A KiltCredentialV1 with embedded KiltAttestationProofV1 proof.
+ */
+export function toVc(
+  input: ICredential,
+  { blockHash, issuer, chainGenesisHash, timestamp }: Params
```
ntn-x2, line 120: How is a user supposed to retrieve the additional pieces required to convert the credential to a VC? If there is a method, then perhaps it could be specified in the JSDoc of the function. If there's not, we should think about a way to make it easy for them.

packages/legacy-credentials/src/vcInterop.spec.ts
```diff
@@ -210,3 +207,30 @@ it('VC has correct format (full example)', () => {
...
+  const VC = toVc(credential, {
+    issuer: attestation.owner,
+    chainGenesisHash: mockedApi.genesisHash,
+    blockHash,
+    timestamp,
+  })
+  expect(fromVC(VC)).toMatchObject(credential)
+})
+
+it.skip('it verifies credential with selected properties revealed', async () => {
```
ntn-x2, line 221: Is this `skip` on purpose?

packages/legacy-credentials/src/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function jsonLDcontents(
+  claim: PartialClaim,
+  expanded = true
+): Record<string, unknown> {
+  const { cTypeHash, contents, owner } = claim
+  if (!cTypeHash) throw new SDKErrors.CTypeHashMissingError()
+  const vocabulary = `${CType.hashToId(cTypeHash)}#`
+  const result: Record<string, unknown> = {}
+  if (owner) result['@id'] = owner
```
ntn-x2, line 28: Can you still add a block with `{}`? It's easier to read and less prone to mistakes 😁

packages/legacy-credentials/src/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!expanded) {
+    return {
+      ...result,
+      '@context': { '@vocab': vocabulary },
```
ntn-x2, line 32: I would put the `@context` last, as it is possible that the `contents` contains a `@context` property. At the very least we should make sure that's not a key of the `contents` object, otherwise the context would get overridden.

ntn-x2: The same applies for `result`, really. Maybe `contents` can be put first, then `result` and then our `@context`.

packages/legacy-credentials/src/utils.ts
ntn-x2: Also, for the same reason as below, I would add the `@id` property after the rest of the document has been created.

ntn-x2 (commented):
I assume the `Claim.ts` and `Credential.ts` have simply been copy-pasted from the core package? I also did not check the tests in detail yet, also not sure I would be able to provide much feedback on those.

In general, I like a lot the API of the legacy package, with one question about we can make creating a VC from a legacy credential eas*ier* for people.

packages/vc-export/src/__mocks__/testData.ts
```diff
@@ -0,0 +1,191 @@
...
+  rootHash: '0x',
+}
+_legacyCredential.rootHash = Credential.calculateRootHash(_legacyCredential)
+
+// eslint-disable-next-line import/no-mutable-exports
+export let legacyCredential: ICredential = JSON.parse(
+  JSON.stringify(_legacyCredential)
+)
+beforeEach(() => {
+  legacyCredential = JSON.parse(JSON.stringify(_legacyCredential))
```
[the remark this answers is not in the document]
rflechtner: to avoid that changes to it persist between tests

packages/legacy-credentials/src/vcInterop.ts
rflechtner: maybe? It's not public though, just an internal

packages/legacy-credentials/src/vcInterop.ts
rflechtner: that's true, I thought they may be imported in tests, but currently they are not

packages/vc-export/src/__mocks__/testData.ts
```diff
@@ -0,0 +1,191 @@
...
+import { ApiMocks } from '../../../../tests/testUtils'
+
+import { fromInput } from '../KiltCredentialV1'
+import { ATTESTATION_PROOF_V1_TYPE } from '../constants'
+import { KiltCredentialV1 } from '../types'
+
+export const mockedApi = ApiMocks.createAugmentedApi()
+
+const attestationCreatedIndex = u8aToU8a([
+  62,
```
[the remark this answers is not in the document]
rflechtner:
```
  mockedApi.runtimeMetadata.asLatest.pallets
    .find((x) => x.name.match(/attestation/i))!
    .index.toNumber(),
```
seems to work

packages/legacy-credentials/src/vcInterop.ts
rflechtner: this requires an indexer, which we probably should deploy somewhere and default to... 
made a ticket (https://github.com/KILTprotocol/ticket/issues/2896)
I guess we need to defer this feature until the indexer is ready?

packages/legacy-credentials/src/utils.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/786/commits/7cb673598639334c520a65606b7973e4dac4caf4

packages/legacy-credentials/src/utils.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/786/commits/7cb673598639334c520a65606b7973e4dac4caf4

packages/legacy-credentials/src/vcInterop.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/786/commits/7cb673598639334c520a65606b7973e4dac4caf4

rflechtner:
> I assume the `Claim.ts` and `Credential.ts` have simply been copy-pasted from the core package?

Yes indeed, with only minuscule adjustments.

> In general, I like a lot the API of the legacy package, with one question about we can make creating a VC from a legacy credential eas_ier_ for people.

This is what is not super easy per se, because it involves an indexer service that we have yet to implement - once we have that we may try to hide this aspect a little better?

ntn-x2 (commented):
LGTM so far!

ntn-x2 (approved):
Not much has changed, and nothing seems to be off.

PR #787, opened 2023-08-10
chore: create new asset-credentials package

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2862 by renaming the old `asset-did` package into `asset-credentials`, and including the public credentials stuff which used to live in `core`.


ntn-x2:
@rflechtner I don't know why the `yarn install --immutable` fails. I don't get any updates on my Mac...

packages/augment-api/package.json
```diff
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@polkadot/api": "^10.4.0",
     "@polkadot/typegen": "^10.4.0",
+    "@types/node": "^20.4.9",
```
rflechtner, line 45: node 16 is still our default so that's probably safer?

packages/utils/src/Chain.ts
```diff
@@ -5,6 +5,9 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
+// Import needed to turn off TS build issue where `call.args[0].flatMap` cannot be called on a generic `Codec` type.
+import '@kiltprotocol/augment-api'
```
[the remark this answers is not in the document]
ntn-x2: Solved.

packages/augment-api/package.json
ntn-x2: Solved.

rflechtner (approved):
LGTM!

PR #789, opened 2023-08-11
refactor!: remove error types

Ad96el:
## fixes [KILTProtocol/ticket#2868](https://github.com/KILTprotocol/ticket/issues/2868)

BREAKING CHANGE: This PR removes error types that require users to update their code.

### DIFF
--- a/packages/utils/src/SDKErrors.ts
+++ b/packages/utils/src/SDKErrors.ts
@@ -35,8 +35,6 @@ export class CTypeIdMismatchError extends SDKError {
   }
 }
 
-export class CTypeUnknownPropertiesError extends SDKError {}
-
 export class UnsupportedKeyError extends SDKError {
   constructor(keyType: string) {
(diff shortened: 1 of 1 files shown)


arty-name (commented):
Should this be marked as a breaking change?

arty-name:
> I would rather mark the first PR as a breaking change. How can I do that? 

The formatting rules for the commits are called [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), and they also describe how to mark a commit as a breaking change. Since the other PR is already merged, it would have been too late to mark it as breaking, if it weren’t already. 

Still, both of these PRs are breaking, so I would mark this appropriately.

rflechtner:
This removes errors that are now not thrown anywhere any more? Makes sense to remove them then

rflechtner (approved):
LGTM

arty-name (approved):

PR #790, opened 2023-08-11
chore: move utils.Chain into chain-helpers package

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2869. 

We discussed about moving the `chain-helpers` module into the `utils` package. but after checking both packages, I found it would be more logic to simply move the chainy stuff of the `utils` package under the `Blockchain` namespace of the `chain-helpers`.

The new utility functions to parse blocks and retrieve extrinsics, can be accessed as in the following snippet:

```ts
import { Blockchain } from "@kiltprotocol/chain-helpers"

const extrinsic = await Blockchain.retrieveExtrinsicFromBlock(
    blockNumber,
    ({ events }) =>
      events.some(
        (event) =>
          api.events.publicCredentials.CredentialStored.is(event) &&
          event.data[1].toString() === credentialId
      ),
    api
  )
```

I made the input `api: ApiPromise` optional, so that it would allow people to use the default returned by `ConfigService.get('api')` if they need, but internally it needs to be retrieved once and passed along.


packages/utils/src/Chain.ts
```diff
@@ -1,75 +0,0 @@
...
- * Copyright (c) 2018-2023, BOTLabs GmbH.
- *
- * This source code is licensed under the BSD 4-Clause "Original" license
- * found in the LICENSE file in the root directory of this source tree.
- */
-
-import type { ApiPromise } from '@polkadot/api'
-import type { TxWithEvent } from '@polkadot/api-derive/types'
-import type { GenericCall, GenericExtrinsic, Vec } from '@polkadot/types'
-import type { Call, Extrinsic } from '@polkadot/types/interfaces'
```
rflechtner, line 11: I think this can be dropped from the dependencies now

rflechtner (approved):
I like it!

PR #793, opened 2023-08-16
refactor!: remove message encryption types and tools

arty-name:
[We’ve removed messages](https://github.com/KILTprotocol/sdk-js/pull/781) but some message-related types and tools remained, this PR cleans it up. I’d argue that they should rather be added in the [kilt-extension-api](https://github.com/KILTprotocol/kilt-extension-api). 

## Checklist:

    * Link the PR/Ticket here


package.json
```diff
@@ -20,7 +20,7 @@
     "set:version": "npm version --no-git-tag-version --no-workspaces-update --workspaces --include-workspace-root",
     "style": "prettier -l packages",
     "style:fix": "yarn style --write",
-    "test": "jest --coverage --selectProjects=unit",
+    "test": "env LANG=C jest --coverage --selectProjects=unit",
```
arty-name, line 23: Consistent number formatting regardless of system locale

packages/did/src/DidDetails/LightDidDetails.spec.ts
```diff
@@ -20,12 +25,20 @@ import * as Did from '../index.js'
...
+  }
+}
+
 describe('When creating an instance from the details', () => {
   it('correctly assign the right sr25519 authentication key, x25519 encryption key, and service endpoints', () => {
     const authKey = Crypto.makeKeypairFromSeed(undefined, 'sr25519')
-    const encKey = Crypto.makeEncryptionKeypairFromSeed(
-      new Uint8Array(32).fill(1)
-    )
+    const encKey = makeEncryptionKeypair()
```
arty-name, line 41: Not the same public key as before, since it’s not a real public key. This causes light DIDs below to change.

packages/utils/package.json
```diff
@@ -41,7 +41,6 @@
     "@polkadot/util": "^12.0.0",
     "@polkadot/util-crypto": "^12.0.0",
     "cbor-web": "^9.0.0",
-    "tweetnacl": "^1.0.3",
```
arty-name, line 44: SDK can drop it from dependencies now

packages/did/src/DidDetails/LightDidDetails.spec.ts
```diff
@@ -5,7 +5,12 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import { DidDocument, DidServiceEndpoint, DidUri } from '@kiltprotocol/types'
+import {
```
rflechtner, line 8:
```suggestion
import type {
```

packages/utils/src/Crypto.ts
```diff
@@ -365,18 +267,3 @@ export function makeKeypairFromUri<
...
- * @returns The keypair.
- */
-export function makeEncryptionKeypairFromSeed(
-  seed = randomAsU8a(32)
-): KiltEncryptionKeypair {
-  return {
-    ...naclBoxPairFromSecret(seed),
-    type: 'x25519',
-  }
-}
```
rflechtner, line 382: I get removing the encrypt/decrypt stuff, but don't you think we should keep key creation around as we allow a certain set of public key types on DIDs and removing this would mean x25519 keys are the only ones that cannot be created using the sdk?

rflechtner (commented):
Is it a good idea to remove x25519 key creation? How would you imagine people would create and add x25519 public keys to their DID instead?

packages/utils/src/Crypto.ts
arty-name: This is more of a discussion PR. If these keys are only usable for messaging, and the messaging is deprecated, probably we don’t need to support them by the SDK directly. If you think they’d still be useful, let’s close this :)

rflechtner: yeah I would rather keep it around for now. x25519-based key agreement is a use case for our DIDs, and even if there are other encryption schemes based on that besides the one implemented here, I think it's fair to have at least one supported by the sdk.

PR #798, opened 2023-08-23
refactor!: VCs in core

rflechtner:
## fixes KILTProtocol/ticket#2827

Moves the VC type credentials into the core package and removes legacy credential support from it.
THIS DOES NOT YET IMPLEMENT A USER FRIENDLY ISSUANCE, PRESENTATION AND VERIFICATION API - THIS IS TBD!

To be done in a follow-up PR:

- Add support for presentation signing and verification
- Implement simple apis for credential verification, presentation, and issuance

## How to test:

Play around with it?

## Checklist:

    * Link the PR/Ticket here


packages/core/src/credentialsV1/KiltCredentialV1.ts
```diff
@@ -300,36 +323,6 @@ export function fromInput({
...
-}
-
-/**
- * A factory for a CType loader that caches a CType definition once it has been loaded.
- * Used in validating the credentialSubject of a [[KiltCredentialV1]] against the Claim Type referenced in its `type` field.
- *
- * @param initialCTypes An array of CTypes with which the cache is to be initialized.
- * @returns A function that takes a CType id and looks up a CType definition in an internal cache, and if not found, tries to fetch it from the KILT blochchain.
- */
-export function newCachingCTypeLoader(
```
rflechtner, line 316: moved to CType features

packages/core/src/credentialsV1/KiltCredentialV1.ts
```diff
@@ -8,43 +8,66 @@
...
 export {
   credentialIdFromRootHash as idFromRootHash,
   credentialIdToRootHash as idToRootHash,
   getDelegationNodeIdForCredential as getDelegationId,
 } from './common.js'
 
+/**
+ * Credential context URL required for Kilt credentials.
+ */
+export const CONTEXT_URL = 'https://www.kilt.io/contexts/credentials'
```
rflechtner, line 45: moved constants to the files to which they relate most to make the relationships of the them clearer

packages/core/src/credentialsV1/common.ts
```diff
@@ -10,15 +10,17 @@ import { base58Decode, base58Encode } from '@polkadot/util-crypto'
...
-  KILT_ATTESTER_DELEGATION_V1_TYPE,
-  KILT_CREDENTIAL_IRI_PREFIX,
-} from './constants.js'
-import { CredentialMalformedError } from './errors.js'
 import type { KiltAttesterDelegationV1, KiltCredentialV1 } from './types.js'
 
+export const spiritnetGenesisHash = hexToU8a(
+  '0x411f057b9107718c9624d6aa4a3f23c1653898297f3d4d529d9bb6511a39dd21'
+)
+export const KILT_ATTESTER_LEGITIMATION_V1_TYPE = 'KiltAttesterLegitimationV1'
```
rflechtner, line 20: wanted to expose them under the KiltCredentialV1 namespace but was afraid circular dependencies may break things, so they are defined here and exposed there

packages/core/src/presentation/Presentation.spec.ts
rflechtner, line 1: signing is still missing here

packages/core/src/presentation/Presentation.ts
rflechtner, line 1: signing is still TBD

packages/vc-export/src/index.ts
```diff
@@ -5,16 +5,7 @@
...
-export * as Presentation from './Presentation.js'
-export * as DidJWT from './DidJwt.js'
-export * as vcjs from './vc-js/index.js'
-export * as Types from './types.js'
-export * as constants from './constants.js'
-export * from './errors.js'
+export * from './suites/index.js'
+export * from './purposes/index.js'
+export * from './documentLoader.js'
+export { validationContexts as contexts } from './context/index.js'
```
rflechtner, line 11: this package only contains the jsonld-signatures support now. moved this a level up

tests/bundle/bundle-test.ts
```diff
@@ -243,48 +242,53 @@ async function runAll() {
...
+  console.info('Credential structure validated')
+
+  await KiltAttestationProofV1.verify(issued, issued.proof, {
+    cTypes: [DriversLicense],
+  })
+  console.info('Credential proof verified')
+
+  await KiltRevocationStatusV1.check(issued)
+  console.info('Credential status verified')
 }
```
rflechtner, line 293: presentation also still missing

ntn-x2 (commented):
I. Love. It. Very very nice. We just need to expose an easier way to verify a credential, and a way to sign presentations, and then we're golden.

packages/core/src/ctype/CTypeLoader.ts
```diff
@@ -0,0 +1,40 @@
...
+export function newCachingCTypeLoader(
+  initialCTypes: ICType[] = []
+): CTypeLoader {
+  const ctypes: Map<string, ICType> = new Map()
+
+  initialCTypes.forEach((ctype) => {
+    ctypes.set(ctype.$id, ctype)
+  })
+
+  async function getCType(id: ICType['$id']): Promise<ICType> {
```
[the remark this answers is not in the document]
rflechtner: does that help in any way? is this to distinguish between say, network errors and not founds?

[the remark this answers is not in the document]
ntn-x2: I mean, you call. I guess a network error could throw, but a credential not found could return `null`. But that's my opinion. If you are fine with the function throwing in either case, I'm fine with that 😊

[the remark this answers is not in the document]
rflechtner: Phew, that's a tricky one. Ultimately we're talking about the proper design of the CTypeLoader interface, which right now is `(id: ICType['$id']) => Promise<ICType>`. Making it a `(id: ICType['$id']) => Promise<ICType | null>` would mean you always have to handle two different error cases (reject or resolve to null), so I'd only introduce this if we can see how depending implementations (e.g. credential verification) would implement different logic/flows for each of these error cases. Otherwise we'd only introduce an annoyance I would say. Do you have an idea of how this would improve depending functions?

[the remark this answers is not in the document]
ntn-x2: I can only comment as a Rust programmer, that it would be much easier to have as much information as possible conveyed in the function signature. I agree probably handling two different error cases is an annoyance for a TS developer. It's just that exceptions in TS and JS require really good documentation, otherwise it's easy for a consumer to forget about a try-catch and the application to crash.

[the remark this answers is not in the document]
rflechtner: so it seems that `fetchFromChain` also does not make a distinction between connection error and not found, so I'd leave the new CTypeLoader as it is right now as well, because the info we wanted to convey is simply missing

rflechtner:
Note that test pass, just the coverage is too low at the moment - I would defer ramping up coverage to when we have the complete feature set implemented though.

PR #799, opened 2023-09-29
feat!: native DIDDocuments and proper DID resolution interface

ntn-x2:
Second attempt to fix https://github.com/KILTprotocol/ticket/issues/2804. It also fixes https://github.com/KILTprotocol/ticket/issues/2340.

### `DidDetails.ts`

* Types that are only used within the DID package have been moved here from the `@kiltprotocol/types` package
* Most utility functions have been removed. We could re-add some functions or add new utility functions that apply both to light and full DIDs here if we need/want to.

### `FullDidDetails.ts`

* Renaming of `verificationKey` to `verificationMethod`

### `LightDidDetails.ts`

* Types that are only used to deal with light DIDs have been moved inside here
* Renaming of `verificationKey` to `verificationMethod` where needed

### `DidResolver.ts`

* Implements the new (and proper) DID resolution standard as defined in [DID Core](https://www.w3.org/TR/did-core/) and [DID Resolution](https://w3c-ccg.github.io/did-resolution/). The interface for this is declared in `@kiltprotocol/types`, in the `DidResolverV2.ts` module. I tried to carry over as much as possible from the current implementation of the resolver.
* Unit tests still to write.

### `Did.chain.ts`, `Did.rpc.ts`, `Did.signature.ts`

* They have been updated to refer to the new `DidDocument` type, dealing as much as possible with verification methods and DID Documents instead of our current intermediate representation.

## Where the "raw" keypair type still appears

To aid usability, I left the `DidVerificationKey` and `DidEncryptionKey` types in a few places. For instance, when creating a light DID, so that the key does not have to first be converted into a fully-fledged verification method, for which the controller URI is still not known. Also the `Did2.chain.ts` includes mentions of `DidKey`s in the `getStoreTxFromInput`, whereas the other function, `getStoreTxFromDocument` takes a DID Document and composes a DID creation tx out of it. The old `getStoreTx` is hence split into these two functions.

## What else

This is WIP, and I have mostly added new files in the DID and types packages so that old stuff is not broken yet. Comments, optimizations, etc will be improved once we agree this is the right level of abstraction. Additional utility functions can always be added if needed, as long as we don't break this newly-created semantic where everything is a DID Document, a verification method, or a service.


packages/did/src/DidResolver/DidResolverV2.ts
```diff
@@ -0,0 +1,310 @@
...
+    if (accept === 'application/did+ld+json') {
+      const jsonLdDoc: DidDocumentV2.JsonLd<DidDocumentV2.DidDocument> = {
+        ...didDocument,
+        '@context': [W3C_DID_CONTEXT_URL, KILT_DID_CONTEXT_URL],
+      }
+      return JSON.stringify(jsonLdDoc)
+    }
+    // contentType === 'application/did+cbor
+    return cbor.encode(didDocument)
+  })()
```
rflechtner, line 188: Idea for an optimisation:
```suggestion
  const encodeDocument = (() => {
    switch(accept) {
    case: 'application/did+json':
     return (didDocument) => JSON.stringify(didDocument)
    }
   case  'application/did+ld+json': 
   return (didDocument) =>{
      const jsonLdDoc: DidDocumentV2.JsonLd<DidDocumentV2.DidDocument> = {
        ...didDocument,
        '@context': [W3C_DID_CONTEXT_URL, KILT_DID_CONTEXT_URL],
      }
      return JSON.stringify(jsonLdDoc)
      }
  case 'application/did+cbor':
    return (didDocument) => cbor.encode(didDocument)
  default:
    return;
  })()
```
Move this up to the beginning of the function to replace isValidContentType; return 'representationNotSupported' if this does not return a function. If it does, you can call that function below to encode the did document

rflechtner:
looks really nice so far, but it is still hard to figure out how this affects depending functions until the changes are made in-place instead of to copies

packages/did/src/DidResolver/DidResolverV2.ts
ntn-x2: Nice suggestion. I applied it here: https://github.com/KILTprotocol/sdk-js/pull/799/commits/79a75c1c45f69673cf3f8aca60f549c629bf0d16

packages/did/src/Did.rpc.ts
```diff
@@ -191,30 +150,68 @@ export interface DidInfo {
...
+
+  const services = servicesFromChain(serviceEndpoints)
+  if (services.length > 0) {
+    did.service = services
   }
 
   const web3Name = w3n.isNone ? undefined : w3n.unwrap().toHuman()
+  if (web3Name !== undefined) {
+    did.alsoKnownAs = [`w3n:${web3Name}`]
+  }
```
rflechtner, line 210:
```suggestion
  if (w3n.isSome) {
    did.alsoKnownAs = [`w3n:${w3n.unwrap().toHuman()}`]
  }
```

packages/types/src/Did.ts
```diff
@@ -0,0 +1,102 @@
...
+  | 'capabilityDelegation'
+  | 'assertionMethod'
+export type EncryptionRelationship = 'keyAgreement'
+
+export type VerificationRelationship =
+  | SignatureVerificationRelationship
+  | EncryptionRelationship
+
+export type DidSignature = {
+  signerUrl: DidUrl
```
rflechtner, line 45: what's the reason for this rename?

rflechtner (commented):
submitting a first batch of comment to not leave them hanging any longer. more to follow!

rflechtner (commented):
Looks really damn nice, great job and thanks for the contribution! I'm thinking that at some point we may want to introduce two resolution options: one that controls which key type is returned (to aid backwards-compatibility; _if_ anybody is already working with actual did documents at this point) and one that controls whether you get fully qualified DID URIs vs fragments as your ids. But I'd probably just add that after this merge so you can move on.

Just get rid of the Buffer though, we shouldn't need to use that.

packages/did/src/Did.chain.ts
```diff
@@ -419,19 +441,172 @@ export async function getStoreTx(
...
+ * - The service has at most 1 service type, with a value that is at most 50 bytes long.
+ * - The service has at most 1 URI, with a value that is at most 200 bytes long, and which is a valid URI according to RFC#3986.
+ *
+ * @param input The DID Document to store.
+ * @param submitter The KILT address authorized to submit the creation operation.
+ * @param sign The sign callback. The authentication key has to be used.
+ *
+ * @returns The SubmittableExtrinsic for the DID creation operation.
+ */
+export async function getStoreTxFromDidDocument(
```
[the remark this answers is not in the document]
ntn-x2: I just split up the old function that was taking either a DID Document or a bunch of keys and services. Now instead of one function that takes both, it's two functions that each take one, with a clearer naming. What would you suggest to do here?

packages/types/src/Did.ts
```diff
@@ -0,0 +1,102 @@
...
+   * A list of URIs the endpoint exposes its services at.
+   */
+  serviceEndpoint: string[]
+}
+
+export type DidDocument = {
+  id: DidUri
+  alsoKnownAs?: string[]
+  verificationMethod?: VerificationMethod[]
+  authentication?: UriFragment[]
```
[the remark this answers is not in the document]
ntn-x2: No. The current resolver interface allows us to define additional flags for the resolution options, or even specify additional query parameters. We have support for those now, so having an optional flag to specify the type of key returned is as simple as adding logic for parsing this flag, and transform the result before returning it to the client.

packages/types/src/Did.ts
ntn-x2: You mean for the `DidSignature` field? If that's what you are referring to, switching to proper DID Documents implies, in my opinion, using also the right terminology for DID URI vs DID URL. A DID URI + a fragment for a key is a DID URL. So here I wanted to convey that what's attached to a signature is the URL of the signer (legacy name: key URI). Maybe I could rename the `signer` part to `verificationMethod`, if you so like.

packages/did/src/Did.signature.ts
```diff
@@ -95,22 +104,64 @@ export async function verifyDidSignature({
...
+    throw new SDKErrors.DidError(
+      `Signer DID URL "${signerUrl}" is not a valid DID resource.`
+    )
+  }
 
-  const { publicKey } = await didResolveKey(keyUri, expectedVerificationMethod)
+  const { contentStream, contentMetadata } = await dereferenceDidUrl(
+    signer.did,
+    {}
+  )
```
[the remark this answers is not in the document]
ntn-x2: The main problem with dereferencing the verification method, is that it does not tell you which relationships it has with the controller. I opened a PR in the DIF extensions repo: https://github.com/decentralized-identity/did-spec-extensions/pull/21, which will do just that. So, when this PR is merged, we can switch to dereferencing just the method, and requiring the method to be under a specific relationship.
Until then, as you mentioned, it is just simpler to resolve the full DID Document and parse its verification methods ourselves to find the one we are looking for. But again, since the current version of the resolver should be open for extensions, we can add this new logic once the PR is merged.

packages/did/src/Did.utils.ts
```diff
@@ -43,30 +46,54 @@ type IDidParsingResult = {
...
   authKeyTypeEncoding?: string
   encodedDetails?: string
 }
 
+// Exports the params section of a DID URL as a map.
+// If multiple keys are present, only the first one is returned.
+function exportQueryParamsFromUri(didUri: DidUrl): Record<string, string> {
+  const urlified = new URL(didUri)
+  const params: Record<string, string> = {}
+  urlified.searchParams.forEach((value, key) => {
```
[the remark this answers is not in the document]
ntn-x2: The `URLSearchParams` only exposes a `forEach` function. Using `reduce` would imply we first have to convert that to an array, and then call `reduce` on it. I though it would be overkill for such a small thing. What do you think? Also, using `reduce` does not spare us from having to add the key to the set only if not present, otherwise we would take the last value instead of the first one.

packages/did/src/Did.utils.ts
```diff
@@ -98,6 +134,131 @@ export function parse(didUri: DidUri | DidResourceUri): IDidParsingResult {
...
+  sr25519: MULTICODEC_SR25519_PREFIX,
+}
+
+/**
+ * Decode a Multikey representation of a verification method into its fundamental components: the public key and the key type.
+ *
+ * @param publicKeyMultibase The verification method's public key in Multikey format (i.e., multicodec-prefixed, then multibase encoded).
+ * @returns The decoded public key and [[DidKeyType]].
+ */
+export function multibaseKeyToDidKey(
```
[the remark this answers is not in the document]
ntn-x2: We only support multikeys, but we can add support for additional resolution flags if we want to. The resolver interface is flexible enough to do that, but I'd do it in a different PR, if you agree on that.

packages/did/src/Did.chain.ts
```diff
@@ -419,19 +441,172 @@ export async function getStoreTx(
...
+  const {
+    authentication,
+    assertionMethod,
+    keyAgreement,
+    capabilityDelegation,
+    service,
+    verificationMethod,
+  } = input
+
+  const authKey = (() => {
```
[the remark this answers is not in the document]
ntn-x2: Good idea. Done in https://github.com/KILTprotocol/sdk-js/pull/799/commits/eb9c0721890e9add7efdaacdfa64c53234527871. I still verify that at least an authentication key is present or else throw. The rest is more or less what you suggested.

packages/did/src/Did.rpc.ts
ntn-x2: Good: https://github.com/KILTprotocol/sdk-js/pull/799/commits/2274c27128ef1dd9b93583c212db66784859570e

packages/did/src/DidDetails/index.ts
```diff
@@ -5,6 +5,16 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-export * from './DidDetails.js'
+// We don't export the `add*VerificationMethod` functions, they are meant to be used internally
+export {
```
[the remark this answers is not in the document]
ntn-x2: Done: https://github.com/KILTprotocol/sdk-js/pull/799/commits/a98f360394db4fabe3935051d89682a070aebe76

packages/types/src/DidResolver.ts
```diff
@@ -5,74 +5,250 @@
...
-export type DidResolve = (did: DidUri) => Promise<DidResolutionResult | null>
+export type RepresentationResolutionResult<
+  ContentType extends string = string
+> = Pick<ResolutionResult, 'didDocumentMetadata'> & {
+  /**
+   * If the resolution is successful, and if the resolveRepresentation function was called, this MUST be a byte stream of the resolved DID document in one of the conformant representations.
+   * The byte stream might then be parsed by the caller of the resolveRepresentation function into a data model, which can in turn be validated and processed.
+   * If the resolution is unsuccessful, this value MUST be an empty stream.
+   */
+  didDocumentStream?: Buffer
```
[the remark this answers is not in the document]
ntn-x2: You are totally right: https://github.com/KILTprotocol/sdk-js/pull/799/commits/447e64ba692de3e215c90556ecc9a4f12daf9d33.

packages/utils/src/Crypto.spec.ts
```diff
@@ -8,6 +8,7 @@
 import * as string from '@polkadot/util/string'
 import nacl from 'tweetnacl'
 import * as Crypto from './Crypto'
+import { Buffer } from './index.js'
```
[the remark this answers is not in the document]
ntn-x2: Same as comment below: https://github.com/KILTprotocol/sdk-js/pull/799/commits/447e64ba692de3e215c90556ecc9a4f12daf9d33.

ntn-x2:
@rflechtner the reason for the failing unit tests is a bit more complicated to solve. The reason is not only the missing context, but also the fact that the verification suite expects the verification method in the legacy format, while now it's returned in a `Multikey` format.
I fixed the lack of context [here](https://github.com/KILTprotocol/sdk-js/pull/799/commits/caf9e926498a334cd3bddb3eac5c39baf98ef146), and also re-uploaded the new context to IPFS and updated the CID (you can verify it via the Crust resolver, which is the only one I have seen working fine). As for fixing the tests, we need to decide whether we want to fix them in this PR by adding the query param to the resolver so that the old key format is returned, or if we want to merge this PR as is and fix those in a different PR. At most, I could add some tests for the new `DataIntegrity` suite which expects verification methods in `Multikey` formats. What do you think?

rflechtner:
> @rflechtner the reason for the failing unit tests is a bit more complicated to solve. The reason is not only the missing context, but also the fact that the verification suite expects the verification method in the legacy format, while now it's returned in a `Multikey` format.
> 
> I fixed the lack of context [here](https://github.com/KILTprotocol/sdk-js/pull/799/commits/caf9e926498a334cd3bddb3eac5c39baf98ef146), and also re-uploaded the new context to IPFS and updated the CID (you can verify it via the Crust resolver, which is the only one I have seen working fine). As for fixing the tests, we need to decide whether we want to fix them in this PR by adding the query param to the resolver so that the old key format is returned, or if we want to merge this PR as is and fix those in a different PR. At most, I could add some tests for the new `DataIntegrity` suite which expects verification methods in `Multikey` formats, but I feel that's rather part of https://github.com/KILTprotocol/sdk-js/pull/801. What do you think?

Isn't that fixed best by amending the suite so it accepts both multikeys and the legacy key representations? I think this would be fair in general, given that we have been using the legacy format for so long. It is also something that i would want to see more generally in the sdk, but thought we may re-add support for this later, after you merged. For now I'd implement minimal support for legacy keys just for the broken suites, just enough to get the currently failing tests to pass

packages/did/src/Did.signature.ts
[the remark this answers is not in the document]
rflechtner: That's an idea too! So far, as I mentioned, this is typically solved in a two-step process: resolve the verificationMethod first (which is the key), then resolve the controller document (i.e., DID document) based on the verificationMethod's controller. But this can certainly be optimised; you know you need the complete DID document eventually, so you might as well just load that and look up the key from it.

packages/types/src/Did.ts
rflechtner: I see, so it's for consistency's sake. I was irritated a bit by it because it seemed a needless breaking change. Yes the name was not ideal, but is it worth introducing a change that will break interoperability with older sdk versions (which will not be able to verify this) for this rename? 
If we are going to rename it though, I much prefer simply calling it `verificationMethod`, aligning with the convention of LD proofs like the DataIntegrityProof

packages/types/src/Did.ts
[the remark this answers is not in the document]
rflechtner: I was able to work around the URI fragments in https://github.com/KILTprotocol/sdk-js/pull/801, happy to implement this flag in a later addition.

rflechtner (commented):
Amazing! Just two little things and then we're done here :)

packages/types/src/Did.ts
ntn-x2: `verificationMethod` is used to represent the whole object. This is technically a `verificationMethodId`. Would you use this instead?

ntn-x2: Or `verificationMethodUrl`, to distinguish it from cases where only the fragment is present. What do you think?

packages/did/src/Did.chain.ts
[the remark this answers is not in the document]
ntn-x2: Fair point. I noticed it was only used in tests, so I moved it in the `TestUtils` module: https://github.com/KILTprotocol/sdk-js/pull/799/commits/6e42e192e71f2acbdd0b4aec2f6a3c7d8cdef535.

packages/did/src/Did.utils.ts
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/799/commits/1cfe243e301e47b7601be5abd4a8ae6314aea1b1

packages/did/src/Did.utils.ts
```diff
@@ -43,30 +47,54 @@ type IDidParsingResult = {
...
       ? parseInt(versionString, 10)
       : FULL_DID_LATEST_VERSION
+    const queryParameters = (() => {
+      try {
+        const queryParams = exportQueryParamsFromUri(didUri as DidUrl)
+        return Object.keys(queryParams).length > 0 ? queryParams : undefined
+      } catch {
+        throw new SDKErrors.InvalidDidFormatError(didUri)
+      }
+    })()
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/799/commits/1cfe243e301e47b7601be5abd4a8ae6314aea1b1

packages/vc-export/src/documentLoader.ts
```diff
@@ -78,17 +81,65 @@ export const kiltContextsLoader: DocumentLoader = async (url) => {
...
+
+// Returns legacy representations of a KILT DID verification method.
 export const kiltDidLoader: DocumentLoader = async (url) => {
   const { did } = parse(url as DidUri)
-  const { didDocument, didResolutionMetadata } = await resolveCompliant(did)
-  if (didResolutionMetadata.error) {
-    throw new Error(
-      `${didResolutionMetadata.error}:${didResolutionMetadata.errorMessage}`
-    )
+  const { dereferencingMetadata, contentStream } = await dereferenceDid(did)
```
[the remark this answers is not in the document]
ntn-x2: Yup. That's a typo from an old version where I used to resolve it directly in the JSON-LD format, which `resolve` does not support. But since now we always resolve the full DID document in pure JSON, I changed it to use the `resolve` instead: https://github.com/KILTprotocol/sdk-js/pull/799/commits/573373ddd52d93d4d85878d0fd12f770408ffc49.

In general `dereference` should be used when only a verification method/service is being resolved (would make even more sense to do so once and if https://github.com/decentralized-identity/did-spec-extensions/pull/21 gets merged), or when a DID Document in a format other than `application/did+json` is to be returned.

packages/types/src/Did.ts
rflechtner: just `verificationMethod`. That's what has been in use in linked data proofs and their successor, the data integrity proofs (see https://www.w3.org/TR/vc-data-integrity/#proofs). Keep in mind that JSON-LD does not care if you specify a node by value of by reference (-> id), so using a string containing the verificationMethod id instead of the actual object is perfectly normal for those people.

ntn-x2: Yes, I see your point, but it is still our SDK we are talking about. See [my comment on your PR](https://github.com/KILTprotocol/sdk-js/pull/801#discussion_r1370098234) for my answer to this point.

arty-name: As a web developer I’m utterly confused by this interpretation of the [difference between URL and URI](https://www.rfc-editor.org/rfc/rfc3986#section-1.1.3). I expect that others will be confused as well.

ntn-x2: I see. But we have to adhere to [this](https://www.w3.org/TR/did-core/#did-syntax) interpretation of URI and [this](https://www.w3.org/TR/did-core/#did-url-syntax) interpretation of URL, according to the DID Core spec.

arty-name: Right, the DID spec authors have done it again :see_no_evil:

rflechtner: but to get back to the issue that I originally wanted to flag here: I'm against renaming the fields on the DID signature, as this is more of a legacy structure and appears on the old presentations as well as on sporran interfaces as far as I know.
Most things will be signed using data integrity proofs anyway in the future, so why needlessly break this?

ntn-x2: Ok, I'll take the breaking change argument as valid. I will also change `DidUri` to `Did` and keep `DidUrl`. Will try to change occurrences everywhere in JSDoc as well.

ntn-x2: Thanks for the good feedback. I have [reverted the commit for the changes in the `DidSignature` type](https://github.com/KILTprotocol/sdk-js/pull/799/commits/b2f12c5a010b77c5e70bbfe6f5069683b691d7e5) and [tried to change `DidUri` to `Did` everywhere](https://github.com/KILTprotocol/sdk-js/pull/799/commits/2fda0d66c1fa5dc0b0655a5ca98035e6921417fc). Let me know what you think.

rflechtner: I personally like it!

rflechtner (approved):
🚀

PR #809, opened 2023-11-09
feat: full DIP SDK support

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2975. A new `@kiltprotocol/dip` package will be the result of this work, which other projects can depend on to sign cross-chain txs using KILT DIDs.

### DIFF
--- a/packages/dip/src/parent.ts
+++ b/packages/dip/src/parent.ts
@@ -54,7 +54,14 @@ export type DipParentProofInput = {
   linkedAccounts?: readonly PalletDidLookupLinkableAccountLinkableAccountId[]
 }
 
-export const DipParentProofBuilder = ObjectBuilder.new<DipParentProofInput>()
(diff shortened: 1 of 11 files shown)


packages/dip/src/parent.ts
```diff
@@ -0,0 +1,205 @@
...
+  includeWeb3Name?: boolean
+  linkedAccounts?: readonly PalletDidLookupLinkableAccountLinkableAccountId[]
+}
+
+/**
+ * Return a new instance of the builder to generate DIP proofs for the parent relaychain.
+ *
+ * @returns A fresh instance of an [[ObjectBuilder]].
+ */
+export function dipParentProofBuilder(): ObjectBuilder {
```
ntn-x2, line 62: Exporting this as a `const` gives a `TS4023` error.

PR #810, opened 2023-11-21
fix: include encoded details in DID validation logic

ntn-x2:
Fixes https://github.com/KILTprotocol/ticket/issues/2568.

A light DID with encoded details that cannot be decoded to anything, is indeed an invalid DID and should be treated as such 😁😁😁


rflechtner (approved):
Thanks for the fix! We're running both light DID decoding and DID parsing twice now on every light DID resolution, but we can open another ticket for making optimisations I think. It's more urgent that it works right than that it works as fast as possible.

PR #813, opened 2023-12-06
refactor!: package surfaces

rflechtner:
## fixes KILTProtocol/ticket#3047

Selective exports in the sdk-js package to prepare v1.

This is a preliminary selection though; I think to make final decisions we need to make an rc and start integrating it, at which point we'll have more insights.

## Checklist:

    * Link the PR/Ticket here


tests/bundle/bundle-test.ts
```diff
@@ -276,9 +273,6 @@ async function runAll() {
   })
   console.info('Credential issued')
 
-  KiltCredentialV1.validateStructure(issued as KiltCredential.Interface)
-  console.info('Credential structure validated')
-
```
Dudleyneedham, line 281: Not sure if this should be in this PR

rflechtner: I chose not to export `KiltCredentialV1` on sdk-js, so this is no longer available in the bundle tests.

ntn-x2 (commented):
I guess my review on the other PR covers most of the things I would have said here: https://github.com/KILTprotocol/sdk-js/pull/814#pullrequestreview-1774616773

Dudleyneedham:
> 
> This was also my first hunch; ideally we'd want to hide all that mess that is the utils package and possibly even the chain-helpers. However, I think we'd be giving people an incomplete set of tools then, forcing them to import from these other packages for many operations. E.g., I did go over the sporran and SKYC repos though and collected what they use, finding that many of the functions from these packages are indeed used there.
> 

I agree with here. Since we are individuals who use this SDK it can be easier for us to know what tools we need at any given time. If we don't package it in a way any body can take this single package and use it, we would have a lot more usability issues. We want to provide a toolset with the SDK out of the box. We can always figure more out when we start updating our applications to know what we need to export and import in general.

> We could of course accept that, and make it our strategy to replace more and more of those imports with stable and neat higher-level functions we add to sdk-js. Maybe it's time to set some goalposts first. For example,

I agree here. We are not going to get this right first time around. However, we should be happy with the first steps in the direction of providing a software that is usable and stable for a given time.

> * Do we want to be able to have people CRUD & resolve DIDs and issue & verify credentials without having to call non-sdk-js functions?
> * Do we want to make sure that people can at least go through the workshops without having to call functions that are not exported via sdk-js?
> * Do we want to have a recipe for each common interaction with the chain (create a DID, make a token transfer, link an address, register a CType, etc) that does not require imports from pjs or a kilt sub-package?

These each are a yes from me. Those individuals who should take the SDK should be able to do the most basic use cases without exporting. I think this is achievable.

rflechtner:
> I tell you what I would export, and we can discuss about this.

To compare my choices here so far to what you suggested:

> I would not export `ChainHelpers` from `@kiltprotocol/chain-helpers` at all.

I did remove that export.

 > I would not export `Blockchain` and `BalanceUtils` from `@kiltprotocol/chain-helpers`. 

I am still exporting `Blockchain` and `BalanceUtils`, alongside `connect`, `disconnect`, and `init`. This is because `Blockchain` gives access to `signAndSubmit`, and `BalanceUtils` helps with formatting kilt balances. Both seem like basic functionality people would expect of an sdk, but as I said, we could also choose to add that step by step. I'm also looking at if these things wouldn't be nicer to use if they're implemented as custom derives.

> I would not export `Utils` from `@kiltprotocol/utils`.

My approach here is to selectively export individual functions that seemed like they get a lot of usage. What do you think about that?

ntn-x2:
The balance is between exporting stuff that gives us legacy right away, and that we need to maintain if we don't want to break the interface, and keeping stuff in a separate package and "promote" them as we see fit. I like the second approach more. If blockchain helpers must be exported because everything depends on them, so be it. But it's much less maintenance effort than exporting everything else.

ntn-x2 (approved):
While waiting to discuss what to export from `sdk-js`, the rest looks damn good to me 🔥🔥🔥

rflechtner:
I think what we did say was that what we definitely want to export is:
```js
{
  init,
  connect,
  disconnect,
  DidResolver,
  Holder,
  Verifier,
  Issuer,
}
```
We also said we _maybe_ want to export
```js
{
  signAndSubmitTx,
  ConfigService,
}
```
as well.

The most immediate problems I see right now are with our bundle tests: they can only import from sdk-js, and without being able to create a DID or even create a keypair, we can't test anything. We could, possible, use the low-low level to create one, but this would get quite ugly. I think we'd need at least the Identity interface here to make this work, which would need to allow creating a DID and submitting calls.

Dudleyneedham (commented):
Didn't we want to keep the signers? Is there a valid/good reason not to include it?

rflechtner:
> Didn't we want to keep the signers? Is there a valid/good reason not to include it?

just discussed it -> let's export `signerFromKeypair`

packages/sdk-js/src/index.ts
```diff
@@ -9,17 +9,29 @@
...
-export * from '@kiltprotocol/types'
-export { SDKErrors } from '@kiltprotocol/utils'
+import { resolver as DidResolver } from '@kiltprotocol/did'
+
+const { signAndSubmitTx } = Blockchain // TODO: maybe we don't even need that if we have the identity class
+const { signerFromKeypair } = Signers
+
+export {
+  init,
+  connect,
```
Dudleyneedham, line 28: This looks good

packages/sdk-js/src/index.ts
```diff
@@ -9,17 +9,29 @@
...
+  disconnect,
+  DidResolver,
+  Holder,
+  Verifier,
+  Issuer,
+  signerFromKeypair,
+  signAndSubmitTx,
+  ConfigService,
```
Dudleyneedham, line 36: This looks good

packages/utils/src/Signers.ts
```diff
@@ -270,13 +270,13 @@ export async function selectSigners<
...
  */
-export async function selectSigner<
+export function selectSigner<
   SelectedSigner extends AllSigners, // eslint-disable-line no-use-before-define
   AllSigners extends SignerInterface = SignerInterface
 >(
   signers: readonly AllSigners[],
   ...selectors: readonly SignerSelector[]
-): Promise<SelectedSigner | undefined> {
+): SelectedSigner | undefined {
```
Dudleyneedham, line 279: The change is based on our conversation?

Dudleyneedham:
Is it known that our tests are failing?

packages/utils/src/Signers.ts
rflechtner: I moved this change to https://github.com/KILTprotocol/sdk-js/pull/819

rflechtner:
I split this up in a PR that contains improvements and fixes, and this now only has the changes to the exports.

rflechtner:
> Is it known that our tests are failing?

They don't actually fail; the coverage threshold is just not reached. I bumped it down for the time being

Dudleyneedham (approved):
LGTM

PR #814, opened 2023-12-08
refactor!: credentials package

rflechtner:
## addresses https://github.com/KILTprotocol/ticket/issues/2156

Isolates credential related functionality in a `@kiltprotocol/credentials` package.

## How to test:

Nothing much changed here; integration and bundle tests show that functionality is still reachable under the new names.

## Checklist:

    * Link the PR/Ticket here


packages/credentials/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+export * as Attestation from './attestation/index.js'
+export * from './V1/index.js'
+export * as Presentation from './presentation/index.js'
+export * as CType from './ctype/index.js'
+export { DelegationNode, DelegationNodeUtils } from './delegation/index.js'
+export * as DataIntegrity from './proofs/DataIntegrity.js'
+
+export * as issuer from './issuer.js'
```
ntn-x2, line 19: Lowercase?

packages/sdk-js/src/index.ts
ntn-x2, line 1: Are all the things we are exporting from here v1 ready now?

packages/credentials/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export * as holder from './holder.js'
+export * as verifier from './verifier.js'
```
Dudleyneedham, line 21: I thought we were going to capitalise on them.

Dudleyneedham (commented):
Comment about the issue, verifier and holder.

packages/sdk-js/src/index.ts
ntn-x2: I tell you what I would export, and we can discuss about this.

I would not export `Blockchain` and `BalanceUtils` from `@kiltprotocol/chain-helpers`.
I would not export `ChainHelpers` from `@kiltprotocol/chain-helpers` at all.
I would not export `Utils` from `@kiltprotocol/utils`.

Exporting from root means commiting to those. While it's much easier to commit for DID and VC stuff because of how they've been (very nicely) designed and implemented, I would not say the same for utility packages or chain helpers. What do you think?

ntn-x2 (commented):
Nice to see the `core` package gone! I just have one reflection point about what's eventually exported from root that will be marked as v1.

packages/credentials/src/index.ts
rflechtner: this happens as part of https://github.com/KILTprotocol/sdk-js/pull/813; I wanted to keep this PR strictly to renaming and reorganising stuff in the core package, so you can skip over the noise created by this in all the other packages

packages/credentials/src/index.ts
rflechtner: this happens as part of https://github.com/KILTprotocol/sdk-js/pull/813

Dudleyneedham: Most excellent

Dudleyneedham (approved):
LGTM

PR #815, opened 2023-12-08
feat: identity class

rflechtner:
closes kiltprotocol/ticket#3043

# TODOs

- I ultimately envision this to have high-level support for managing an on-chain identity as well; but as-is, you can already make any DID operation (including those changing your DID document) through a combination of `authorizeTx`, `submitTx`, and `update`.
- creating a new DID should allow passing account signers instead of keypairs
- the vocabulary/terminology needs to be refined and finalised to make sure all methods and parameters communicate clearly what they are made for.

### DIFF
--- a/packages/sdk-js/src/Identity.ts
+++ b/packages/sdk-js/src/Identity.ts
@@ -109,30 +109,29 @@ async function loadDidDocument(
 }
 
 class IdentityClass implements Identity {
-  public did: Did
   public resolver: typeof resolve
   public didDocument: DidDocument
-  private didSigners: SignerInterface[]
+  protected didSigners: SignerInterface[]
+  get signers(): SignerInterface[] {
+    return [...this.didSigners]
+  }
+
+  public get did(): Did {
+    return this.didDocument.id
+  }
 
   constructor({
-    did,
     didDocument,
     signers,
     resolver = resolve,
   }: {
-    did: Did
     didDocument: DidDocument
     signers?: SignerInterface[]
     resolver?: typeof resolve
   }) {
-    this.did = did
     this.didDocument = didDocument
     this.didSigners = signers ? [...signers] : []
     this.resolver = resolver
-  }
-
-  get signers(): SignerInterface[] {
-    return [...this.didSigners]
   }
 
   public async update({
@@ -239,15 +238,21 @@ export async function makeIdentity({
   signers,
   resolver = resolve,
 }: {
-  did: Did
+  did?: Did
   didDocument?: DidDocument
   signers?: SignerInterface[]
   keypairs?: Array<Keypair | KeyringPair>
   resolver?: typeof resolve
 }): Promise<IdentityClass> {
+  let didDoc = didDocument
+  if (!didDoc) {
+    if (!did) {
+      throw new Error('either `did` or `didDocument` is required')
+    }
+    didDoc = await loadDidDocument(did, resolver)
+  }
   const identity = new IdentityClass({
-    did,
-    didDocument: didDocument ?? (await loadDidDocument(did, resolver)),
+    didDocument: didDoc,
     signers,
     resolver,
   })
--- a/packages/sdk-js/src/Identity.ts
+++ b/packages/sdk-js/src/Identity.ts
@@ -1,15 +1,17 @@
 /**
- * Copyright (c) 2018-2023, BOTLabs GmbH.
+ * Copyright (c) 2018-2024, BOTLabs GmbH.
  *
  * This source code is licensed under the BSD 4-Clause "Original" license
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import type { GenericExtrinsic } from '@polkadot/types'
 import { u8aEq } from '@polkadot/util'
 import type { Keypair } from '@polkadot/util-crypto/types'
 
-import { multibaseKeyToDidKey, resolve } from '@kiltprotocol/did'
+import { Blockchain } from '@kiltprotocol/chain-helpers'
+import { ConfigService } from '@kiltprotocol/config'
+import type { Issuer } from '@kiltprotocol/credentials'
+import { authorizeTx, multibaseKeyToDidKey, resolve } from '@kiltprotocol/did'
 import type {
   Did,
   DidDocument,
@@ -20,7 +22,6 @@ import type {
   UriFragment,
 } from '@kiltprotocol/types'
 import { SDKErrors, Signers } from '@kiltprotocol/utils'
-import type { TransactionResult } from '@kiltprotocol/credentials/src/V1/KiltAttestationProofV1'
 
 /**
  * An Identity represents a DID and signing keys associated with it.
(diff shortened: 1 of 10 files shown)


packages/sdk-js/src/Identity.ts
```diff
@@ -0,0 +1,456 @@
...
+   * @returns The (in-place) modified Identity object for chaining.
+   */
+  addSigner: (...signers: SignerInterface[]) => Identity
+  /**
+   * Convenience function similar to {@link addSigner}, but creates signers for all matching known algorithms from a keypair.
+   *
+   * @param keypairs
+   * @returns The (in-place) modified Identity object for chaining.
+   */
+  addKeypair: (...keypairs: Array<Keypair | KeyringPair>) => Promise<Identity>
```
rflechtner, line 55: this name may be problematic; if I implement a method that helps rotate or add public keys to the DID document, these namings may clash

packages/sdk-js/src/Identity.ts
```diff
@@ -0,0 +1,456 @@
...
+}): Promise<IdentityClass>
+export async function makeIdentity<T extends IdentityClass>(args: {
+  did?: Did
+  didDocument?: DidDocument
+  signers?: SignerInterface[]
+  keypairs?: Array<Keypair | KeyringPair>
+  resolver?: typeof resolve
+  transactionStrategy: TransactionStrategy<T>
+}): Promise<T>
+export async function makeIdentity<T extends IdentityClass>({
```
rflechtner, line 264: there's `makeIdentity` and `newIdentity`; names could be improved to distinguish them better.

packages/sdk-js/src/Identity.ts
```diff
@@ -0,0 +1,456 @@
...
+
+export async function newIdentity(args: {
+  keys: TypedKeyPair | TypedKeyPairs
+  resolver?: typeof resolve
+}): Promise<IdentityClass>
+export async function newIdentity<
+  T extends IdentityClass &
+    Required<Pick<Identity, 'submitTx' | 'submitterAddress'>>
+>(args: {
+  keys: TypedKeyPair | TypedKeyPairs
```
rflechtner, line 334: this parameter could accept signers too; but the name would have to change. I don't like it anyway though - it's too similar but different from the `keypairs` parameter of `makeIdentity`

tests/bundle/bundle-test.ts
```diff
@@ -7,108 +7,46 @@
...
-  api: ApiPromise,
-  call: SubmittableExtrinsic,
-  did: string,
-  signer: SignerInterface,
-  submitter: string,
-  nonce = 1
+async function createFullDidIdentity(
+  payer: SignerInterface<'Ed25519', KiltAddress>,
+  seed: string,
+  signKeyType: KiltKeyringPair['type'] = 'sr25519'
```
arty-name, line 32: SR keys again?

tests/bundle/bundle-test.ts
```diff
@@ -118,114 +56,72 @@ async function runAll() {
...
-      166, 39, 198, 155, 45, 189, 161, 94, 215, 229, 128, 133, 66, 81, 25, 174,
-      3,
-    ]),
-  }
+  const FaucetSeed =
+    'receive clutch item involve chaos clutch furnace arrest claw isolate okay together'
+  const payerKp = generateKeypair({ seed: FaucetSeed, type: 'ed25519' })
   const payerSigner = await signerFromKeypair<'Ed25519', KiltAddress>({
-    keypair: faucet,
+    keypair: payerKp,
```
arty-name, line 63: Not introduced here, but `signerFromKeypair<'Ed25519', KiltAddress>(` looks really ugly, poor DX

tests/bundle/bundle-test.ts
```diff
@@ -118,114 +56,72 @@ async function runAll() {
...
-  const authPublicKey =
-    '0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
-
-  // const encPublicKey =
-  //   '0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'
-
+  const authPublicKey = new Uint8Array([
+    170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170,
+    170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170,
+    170, 170,
```
arty-name, line 87: Would this nice `.fill()` method be helpful here?

tests/bundle/bundle-test.ts
```diff
@@ -267,25 +158,21 @@ async function runAll() {
...
   ) {
     throw new Error('Claim content inside Credential mismatching')
   }
 
-  const issued = await Issuer.issue(credential, {
-    did: alice.id,
-    signers: [...(await aliceSign(alice)), payerSigner],
-    submitterAccount: payerSigner.id,
-  })
+  const issued = await Issuer.issue(credential, alice as any)
```
arty-name, line 175: I’m used to seeing this test as an example of real-life usage of the SDK. As an SDK user I won’t be happy to write `as any` in my code. Is this an indication of an issue? (pun intended)

tests/bundle/bundle-test.ts
rflechtner: true, ed25519 is our default now right?

tests/bundle/bundle-test.ts
rflechtner: ahh that's a good point!

tests/bundle/bundle-test.ts
rflechtner: can actually be avoided by setting `KiltAddress` as the type default for the Id type on `signerFromKeypair`

tests/bundle/bundle-test.ts
rflechtner: ah, probably a leftover, I don't think it's necessary

rflechtner: But more generally, what do you think about the overall approach to interacting with identity primitives here? Is this something you would have liked to implement sporran and SKYC on?

tests/bundle/bundle-test.ts
arty-name: I believe the key type can also be inferred

rflechtner: Yup, that was already happening, but I had to specify the second type bc there was no default

tests/bundle/bundle-test.ts
arty-name: I think my judgement is currently clouded by the pain of seeing the bundle-test in its current state. I believe it should be talked over.

rflechtner: Yeah, would love to chat, because right now I wouldn't really know what to change here

tests/bundle/bundle-test.ts
```diff
@@ -118,114 +56,64 @@ async function runAll() {
...
+    keys: {
+      authentication: [
+        { secretKey: authPublicKey, publicKey: authPublicKey, type: 'ed25519' },
+      ],
+      keyAgreement: [
+        { secretKey: encPublicKey, publicKey: encPublicKey, type: 'x25519' },
+      ],
+    },
+  })
   const address = api.createType('Address', authPublicKey).toString()
```
Dudleyneedham, line 96: How does this work? I tried using this but failed.

Dudleyneedham: The api throws an error back that it has a secret key.

Dudleyneedham: If I pass only the publickey I get my address.

Dudleyneedham:
@rflechtner Can we close this?

rflechtner:
I do believe this is superseded by https://github.com/KILTprotocol/sdk-js/pull/881, so I'll close this PR. Please do not delete the branch though, as we may want to cherry-pick some features - especially the `generateKeypair` function - into our new solution.

PR #818, opened 2023-12-10
fix: ignore out-of-range log level in ConfigService.spec.ts

1gn0r4nd:
Added // @ts-ignore 

Only values 0-6 are valid. https://github.com/vauxite-org/typescript-logging/blob/cf0d3e7d52b1da0650b16308cc3f1a56bcb95b5b/core/src/typescript/main/core/api/LogLevel.ts#L5C1-L12C10

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/config/src/ConfigService.spec.ts
+++ b/packages/config/src/ConfigService.spec.ts
@@ -31,6 +31,7 @@ describe('Log Configuration', () => {
       ConfigService.LoggingFactory.getLogger('test1').getLogLevel()
     ).toEqual(LogLevel.Info)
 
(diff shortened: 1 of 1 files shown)


rflechtner:
Hi @1gn0r4nd, thanks for pointing this out! Can you provide a little bit of context for why this ignore is necessary? Did you get typescript errors on a higher version of `typescript-logging` than we're using here?

1gn0r4nd:
Hi rflechtner,

I wanted the API of the SDK, which I have now made available on https://buildonkilt.uat.sr/. 
While trying to generate the docs with typedoc, this is the error I got, the ignore allows me to ignore TS2345
```
packages/config/src/ConfigService.spec.ts:34:34 - error TS2345: Argument of type '-100' is not assignable to parameter of type 'LogLevel'.

34     ConfigService.modifyLogLevel(-100)
                                    ~~~~
```

PR #819, opened 2023-12-13
chore: various smaller improvements

rflechtner:
Took smaller improvements out of https://github.com/KILTprotocol/sdk-js/pull/813 to reduce clutter.

- Make selectSigner(s) synchronous. The potential need for async selection rules has not materialised, and we can always have `selectSignerAsync` or such if that becomes useful.
- add a missing signer export, make private key extraction more efficient. 
- Uppercase the roles namespaces.
- Fix some mentions or `core` I had missed.


Dudleyneedham (approved):
LGTM

PR #828, opened 2023-12-18
chore: rename vc export package

rflechtner:
## fixes KILTProtocol/ticket#3060

Renames the vc-export package, as the `legacy-credentials` package now takes care of this; instead, the package just supplies tools to integrate with `@digitalbazaar/vc`.

Also updates the README and applies a rename and slight logic change to the proof purpose to be used with the suite. The rationale here: while other purpose plugins check that the verification method used matches the expected purpose, our proof is not bound to a VM, so we should assert there is no purpose present on the proof. 

## Checklist:

    * Link the PR/Ticket here


packages/jsonld-suites/README.md
```diff
@@ -0,0 +1,58 @@
...
+    - issue a `KiltAttestationProofV1` type proof for a `KiltCredentialV1` type VC.
+    - check the revocation status of a `KiltCredentialV1`.
+
+- `NoProofPurpose`
+
+  - `ProofPurpose` class to be used in combination with the attestation proof suite above.
+
+- `Sr25519Signature2020`
+  - A `suite` implementation for creating and verifying sr25519 linked data signatures.
+- `Sr25519VerificationKey2020`
```
ntn-x2, line 42: Empty line?

ntn-x2 (approved):
LGTM!

ntn-x2 (approved):
LGTM!

packages/jsonld-suites/README.md
rflechtner: the auto-formatter set it up that way, not sure why

PR #833, opened 2024-02-05
chore(deps): bump node to v18

rflechtner:
## re https://github.com/KILTprotocol/ticket/issues/3081

Require at least node v18 (maintenance), CI uses node v20 (current). Testing against v16 has moved to optional.
https://nodejs.org/en/about/previous-releases

## How to test:

CI should not break.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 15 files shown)


ntn-x2 (approved):
Very good! I assume this is not a breaking change for downstream users? They would only get a warning if using a version < 18, right? Will this be released after SDK v1 is out?

rflechtner:
> Very good! I assume this is not a breaking change for downstream users? They would only get a warning if using a version < 18, right? Will this be released after SDK v1 is out?

Depending on their config the package manager may error, but they should be able to disable that.
And yes, id say this gets released with v1, otherwise we'd have to backport (which we may have to do eventually anyway though)

PR #836, opened 2024-02-07
chore: release 0.35.0

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3119

Minor release to add support for the new dispatchAs extrinsic in CType and Public Credential fetching logic.
Also updates type augmentation for the upcoming runtime upgrade to 1.12.1, which additionally introduces changes to the balances pallet that can be considered breaking.
The changes also require an update to polkadot/api^10.7.3. The dependency manifests have been updated accordingly.


packages/augment-api/src/interfaces/augment-api-errors.ts
```diff
@@ -43,35 +43,43 @@ declare module '@polkadot/api-base/types/errors' {
...
+       * Transfer/payment would kill account.
+       **/
+      Expendability: AugmentedError<ApiType>;
       /**
        * Balance too low to send value.
        **/
       InsufficientBalance: AugmentedError<ApiType>;
       /**
-       * Transfer/payment would kill account
+       * Account liquidity restrictions prevent withdrawal.
```
ntn-x2, line 66: Wouldnt' these be breaking changes that require a 0.35 version?

ntn-x2 (approved):
Approving in case you need to merge after I'm gone.

packages/augment-api/metadata/spiritnet.json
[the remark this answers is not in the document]
rflechtner: yup, 1.12.1

packages/augment-api/src/interfaces/augment-api-errors.ts
rflechtner: I was thinking the same, yeah. Let's make it a new minor.

rflechtner:
> We don't have a single test testing for the new stuff? Should we add it in the CType integration tests?

Added integration tests. For that I also had to update to the new standalone-node docker image.

PR #842, opened 2024-02-26
refactor: new augmentation

rflechtner:
## fixes KILTProtocol/ticket#2224

Externalise the augment-api package, now in this repo: https://github.com/KILTprotocol/types-augment
I do think that eventually the type definitions should move there too, as the augmentation and types would probably be released together.
And a versioning based on the runtime version makes a lot of sense to me for the type definitions as well. 

## How to test:

Unit tests are sufficient.

## Checklist:

    * Link the PR/Ticket here


packages/sdk-js/package.json
```diff
@@ -39,10 +39,12 @@
     "webpack-cli": "^4.9.2"
   },
   "dependencies": {
+    "@kiltprotocol/augment-api": "1.11210.0-rc",
```
ntn-x2, line 42: So no we're pulling a specific version RC into `develop`? Would we want to have `master` being merged here, and then pull from a specific branch upon a release candidate?

ntn-x2 (commented):
Left one comment about managing versions between the two repos. Also, why not first migrate the types in there too, and then remove them from here?

packages/sdk-js/package.json
rflechtner: 57d08f9f0f6dc0230f3cd93f6f52e776751cc878

ntn-x2: We're still pinning it to a specific version from our `develop` branch? Wouldn't `latest` be enough here?

rflechtner: nope, because this is not latest

rflechtner: > Would we want to have `master` being merged here, and then pull from a specific branch upon a release candidate?

I'm not sure what you're referring to with merging master - this is not going to happen as master is still on pre-V1. But before releasing V1, this would certainly be replaced with a properly released version of the augment-api & type-definitions packages

ntn-x2: No I mean that this `develop` branch would simply track the `master` branch of the new repo. Then each release would track a specific version of the new repo. So we can make sure that any changes on `master` in the new repo are compatible with the `develop` version of this, kinda like we do for integration tests against the chain `develop` branch.

ntn-x2: I agree that releasing will involve pinning the version, but before then I think it's a bit "weird" to have `develop` being pinned to a specific released version of the new augmentation package.

rflechtner: a) latest is a tag. this will select one specific version tagged with latest, which is the latest full release and not the rc that I'm currently experimenting with.
b) what's the point of a peer dependency if you then require one specific version?

ntn-x2: Sorry, I just realized this was changed to be a peer dependency. 

> a) latest is a tag. this will select one specific version tagged with latest, which is the latest full release and not the rc that I'm currently experimenting with.

Is this supposed to change then at some point? Is it just temporarily used to verify that things work?

rflechtner: yes, absolutely. Once we fully release the augment-api package we can switch this to a sensible version range, such as `^1.11210.0`, which is essentially equivalent to `latest` (allows all higher versions accept new majors).

ntn-x2 (approved):

PR #846, opened 2024-03-13
chore: remove type definitions

rflechtner:
## fixes KILTProtocol/ticket#3219

Removes type-definitions from this repo. Companion PR of https://github.com/KILTprotocol/types-augment/pull/2

## Checklist:

    * Link the PR/Ticket here


rflechtner:
@ntn-x2 There's one thing I'm undecided about here: should the type-definitions be defined as a peer dependency so that selecting the version that works with your runtime is a conscious decision (also helps when you want to use types for experimental runtimes in development etc) or should we just require a version? The only thing that does not automatically get matched to the actual runtime version used are the runtime api definitions, right?

Would this justify making people match their dependency version to the runtime version, instead of always using latest, for example?

ntn-x2:
> The only thing that does not automatically get matched to the actual runtime version used are the runtime api definitions, right?

Yes, and the types they use. 

I mean here we're talking about the SDK, so perhaps for me it would make more sense to pin it to a specific version and hence make it a regular dependency. Also, soon we might have also runtime calls info inside, so that would remove the last bit of friction as well. If it wasn't an SDK, it could have made sense to have it as a peer dependency, but I think in this case we could go with a regular one.

rflechtner:
> I mean here we're talking about the SDK, so perhaps for me it would make more sense to pin it to a specific version and hence make it a regular dependency.

I would want to avoid pinning a version at all costs because that means having to release an sdk version just to update the types, which is the annoyance that this whole thing was meant to end. But probably this is not really what you meant? In the end, if we set a dependency like `^1.11210.0` this means people can still pull in the latest type definitions, and could also use selective resolution mechanisms to choose a specific version. So fine by me.

ntn-x2:
> In the end, if we set a dependency like ^1.11210.0 this means people can still pull in the latest type definitions, and could also use selective resolution mechanisms to choose a specific version. So fine by me.

I guess that's the difference between dependencies and peer dependencies? Fine for me as well actually.

rflechtner:
> I guess that's the difference between dependencies and peer dependencies? Fine for me as well actually.

No that's just semver ranges. `^` means "this or a higher minor" (implying this major) and `~` means "this or a higher patch" (implying this minor).

So by allowing any higher minor, we allow using any runtime-specific version of the types to come. Installing without a lock file would pull in the types for the latest runtime.

A peer dependency means that no version will be installed - it's just a hint to the user to add a dependency on type-definitions along with the one on sdk-js, and than _that_ version will be used. So the user controls directly which types are going to be installed, matching it to the runtime they are dealing with. 

But presumably they are always going to be dealing with the latest runtime, so having a regular dependency as I outlined above should be fine too.

ntn-x2 (approved):

PR #847, opened 2024-03-25
refactor!: new data integrity

rflechtner:
Uses the new data integrity suite canonicalisation logic. Breaks verification of existing signatures. 

## How to test:

This is tested as part of Presentation verification.

## Checklist:

    * Link the PR/Ticket here


packages/credentials/src/proofs/DataIntegrity.ts
```diff
@@ -65,6 +66,40 @@ export type DataIntegrityProof = {
...
+  proof: DataIntegrityProof
+  document: Record<string, unknown>
+  suite: CryptoSuite<any>
+  options?: Record<string, unknown>
+}): Promise<Uint8Array> {
+  if (suite.createVerifyData) {
+    return suite.createVerifyData({ proof, document })
+  }
+  // jcs suites will not work with the logic below - this would fix it.
+  if (suite.name.includes('-jcs-')) {
```
ntn-x2, line 84: Do you think this is a good solution in the long term? Do you think having a known set of suites to match against would be a better approach? It would require releasing a new SDK (patch), but it would be more clear as to what it does.

rflechtner: yeah I'm not sure about it either. All of our suites implement `createVerifyData` and thus would be matched on the first branch, so in that case there'd be nothing to match here - or would you still match their names, in case sb uses divergent implementations?

ntn-x2: Yeah I would match by suite name, even just to make deliberate choices as to what suites we found need to be addressed. There won't probably be hundreds new every week, so it's fine to then release new minors to add a new suite when it pops up.

rflechtner: Ok so I would do the eddsa & ecdsa suites, and our ES256k and sr25519 suites. Anything else you can think of?

ntn-x2: Not that I am aware of, besides some old and outdated suites which we can ignore.

rflechtner: cfd64e81c9e7d6bb91962300940f918a9f1a42cb

ntn-x2 (approved):
🚀

PR #848, opened 2024-03-27
feat: allow light did signers on getStoreTx

rflechtner:
re/ https://github.com/KILTprotocol/ticket/issues/3242

## How to test:

- tests tbd

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/did/src/Did.chain.ts
+++ b/packages/did/src/Did.chain.ts
@@ -45,6 +45,10 @@ import {
   getFullDid,
   parse,
 } from './Did.utils.js'
+import {
+  NewLightDidVerificationKey,
+  createLightDidDocument,
+} from './DidDetails/LightDidDetails.js'
 
 export type ChainDidIdentifier = KiltAddress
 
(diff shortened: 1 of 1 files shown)


packages/did/src/Did.chain.ts
```diff
@@ -418,13 +419,18 @@ export async function getStoreTx(
     .createType(api.tx.did.create.meta.args[0].type.toString(), apiInput)
     .toU8a()
 
+  // signers can identify using the authentication public key encoded as address or hex, or a light did authentication key
```
ntn-x2, line 422: We do have other functions to match a DID, also allowing a fragment to be present. Would that make more sense? If there's any match, check that the fragment is `#authentication`.

rflechtner: true, I'll check that

rflechtner: 2f7fe68d90fa644c87eb498001bc5b8ae23e99c8

ntn-x2 (approved):

PR #850, opened 2024-03-28
chore: make resolution options optional

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3268

Don't require passing empty resolution options into did resolver functions.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/did/src/DidResolver/DidResolver.ts
+++ b/packages/did/src/DidResolver/DidResolver.ts
@@ -171,9 +171,9 @@ export async function resolve(
  */
 export async function resolveRepresentation(
   did: Did,
-  { accept }: DereferenceOptions<SupportedContentType> = {
-    accept: DID_JSON_CONTENT_TYPE,
-  }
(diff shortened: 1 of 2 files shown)


ntn-x2:
According to the spec:

> A [metadata structure](https://www.w3.org/TR/did-core/#metadata-structure) containing properties defined in [7.1.1 DID Resolution Options](https://www.w3.org/TR/did-core/#did-resolution-options). This input is REQUIRED, but the structure MAY be empty.

So it seems the object as such is required, that's why I went with a non-optional field. I would tend to leave the interface compliant with the specification, although it is definitely less usable.

rflechtner:
Didn't we agree to be pragmatic about it? We can't implement the interface in a way that there are no typescript errors if you omit the second parameter, so making it optional on the interface is the only way to do so that I see

rflechtner:
I think we reached an agreement here

PR #856, opened 2024-04-16
refactor: use credentialSubject on credential constructor

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3269

Align with VC terminology to avoid confusion. Tom suggested that instead of its goal of easing credential creation the current solution just introduces confusion and is incoherent with our goal of aligning our stack more closely with the specs and their vocabulary.

## How to test:

Tests have been updated.

## Checklist:

    * Link the PR/Ticket here


ntn-x2 (approved):
I personally like the change.

PR #866, opened 2024-05-22
fix: CType ID is checked in presentation

Dudleyneedham:
## fixes [#3391](https://github.com/KILTprotocol/ticket/issues/3391)
Checking the ctype ID in the credential to verify if matching ctype and credential ctype hash

## How to test:

yarn test

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/core/src/credential/Credential.ts
+++ b/packages/core/src/credential/Credential.ts
@@ -301,10 +301,15 @@ export function verifyWellFormed(
   verifyDataIntegrity(credential)
 
   if (ctype) {
+    if (`kilt:ctype:${credential.claim.cTypeHash}` !== ctype.$id) {
+      throw new SDKErrors.CTypeIdMismatchError(
+        ctype.$id,
+        credential.claim.cTypeHash
+      )
+    }
     verifyClaimAgainstSchema(credential.claim.contents, ctype)
   }
 }
-
 /**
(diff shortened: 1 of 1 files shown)


packages/core/src/credential/Credential.ts
```diff
@@ -301,10 +301,15 @@ export function verifyWellFormed(
   verifyDataIntegrity(credential)
 
   if (ctype) {
+    if (`kilt:ctype:${credential.claim.cTypeHash}` !== ctype.$id) {
+      throw new SDKErrors.CTypeIdMismatchError(
+        ctype.$id,
+        credential.claim.cTypeHash
+      )
+    }
```
Dudleyneedham, line 309: I was wondering if we wanted to do it another way.

```
if (ctype && `kilt:ctype:${credential.claim.cTypeHash}` !== ctype.$id) {
      throw new SDKErrors.CTypeIdMismatchError(
        ctype.$id,
        credential.claim.cTypeHash
      )
    }
if (ctype) {
 verifyClaimAgainstSchema(credential.claim.contents, ctype)
  }
```

rflechtner: no, this is good, but we've got the CType.hashToId() helper function for `kilt:ctype:${credential.claim.cTypeHash}`

rflechtner (approved):
LGTM, but please use the existing helper function to compute the ctype id from its hash

PR #868, opened 2024-05-27
chore: update node to >= v20.0

kilted-andres:


rflechtner:
What's the reason for requiring 20.13? Why do we need a specific minor version?

PR #871, opened 2024-06-12
chore: update to node v20 

kilted-andres:
## fixes part of KILTProtocol/ticket#3081

- Makes sure the `package.json` specifies `yarn` version. 
- Upgrades the used **node version**  on the `.nvmrc`  to `lts/Iron`. 
- Makes sure the **GitHub workflows** uses node version 20. 
  For this following must be set: 
  - `actions/checkout@v4`
  - `actions/setup-node@v4` **with:**  _either_
    - **node-version-file:** '`package.json`' or '`.nvmrc`' 
    -  **node-version:**  `'`<Node.js version using [SemVer](https://semver.org/) or [aliases](https://github.com/nodejs/Release?tab=readme-ov-file#release-schedule) > `'`
    - `actions/download-artifact@v4`
    - _optional, but done:_ `actions/cache@v4`
  -  `aws-actions/configure-aws-credentials@v4`
  -  `aws-actions/amazon-ecr-login@v2`
 - Makes sure the **Dockerfiles** also use the node version 20. 
  This means that the following must be set: 
      - `FROM node:20-alpine as builder`
  -  ~~It also updates dependencies: mostly to latest minor, but major for ones with complains.~~ Dropped.

### DIFF
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -61,7 +61,9 @@ jobs:
         node-version: [18.x, 20.x]
         required: ['required']
         include:
-          - node-version: [16.x, 22.x]
+          - node-version: 16.x
+            required: 'optional'
+          - node-version: 22.x
             required: 'optional'
 
     continue-on-error: ${{ matrix.required == 'optional' }}
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -58,10 +58,12 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x]
         required: ['required']
         include:
           - node-version: 16.x
+            required: 'optional'
+          - node-version: 22.x
             required: 'optional'
 
     continue-on-error: ${{ matrix.required == 'optional' }}
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -108,9 +108,9 @@ jobs:
 
       - name: Login to Amazon ECR
         id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
+        uses: aws-actions/amazon-ecr-login@v2
 
-      - name: pull image
+      - name: pull node image
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           ECR_REPOSITORY: kilt/prototype-chain
--- a/.github/workflows/tests-polkadot-deps.yml
+++ b/.github/workflows/tests-polkadot-deps.yml
@@ -58,10 +58,10 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x]
         required: ['required']
         include:
-          - node-version: 16.x
+          - node-version: [16.x, 22.x]
             required: 'optional'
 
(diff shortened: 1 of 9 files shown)


.github/workflows/tests-polkadot-deps.yml
```diff
@@ -58,7 +58,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]
```
rflechtner, line 61: 22 is not yet an active LTS, I'd say we make it optional for now. For this, add it to the 'include'

.github/workflows/tests.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18, 20]
+        node-version: [18, 20, 22]
```
rflechtner, line 83: same here

kilted-andres:
Node v22 will be lts in about 3 months. The docker image running the thing `From:node-alpine` actually has node v22 preinstalled. Andrés RodríguezOn 13. Jun 2024, at 10:35, Raphael Flechtner ***@***.***> wrote:﻿
@rflechtner commented on this pull request.



In .github/workflows/tests-polkadot-deps.yml:
> @@ -58,7 +58,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]

22 is not yet an active LTS, I'd say we make it optional for now. For this, add it to the 'include'



In .github/workflows/tests.yml:
> @@ -80,7 +80,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18, 20]
+        node-version: [18, 20, 22]

same here

.github/workflows/tests.yml
kilted-andres: Did it here [0404a5e](https://github.com/KILTprotocol/sdk-js/pull/871/commits/0404a5e49a9148c1b75bbff5dc392962725c7e86).

.github/workflows/tests-polkadot-deps.yml
```diff
@@ -61,7 +61,9 @@ jobs:
         node-version: [18.x, 20.x]
         required: ['required']
         include:
-          - node-version: [16.x, 22.x]
+          - node-version: 16.x
+            required: 'optional'
```
ntn-x2: I think Node 16 was not optional, so it should not be optional now either. It's ok to be optional for Node 22 tho.

kilted-andres: It was optional before: https://github.com/KILTprotocol/sdk-js/blob/2ea03c967efeb9ef737413d4a0407bb8d185cce9/.github/workflows/tests.yml#L86

kilted-andres:
For some reason. If the node v22 is optional, an integration test fails. If it required, it passes.

ntn-x2:
@kilted-andres as far as I could see, the error you were getting is unrelated to making Node 22 optional or not. It seems to be some sort of concurrency issue in the tests. I would suggest the following: update the PR to make Node 22 optional. If the integration tests fail, re-run only those again from the GitHub UI, then it should be good to go.

kilted-andres:
@ntn-x2 :
> @kilted-andres as far as I could see, the error you were getting is unrelated to making Node 22 optional or not. It seems to be some sort of concurrency issue in the tests. I would suggest the following: update the PR to make Node 22 optional. If the integration tests fail, re-run only those again from the GitHub UI, then it should be good to go.

I tried that before "reverting" it.

rflechtner (approved):
LGTM now

PR #872, opened 2024-06-14
fix: bump @kiltprotocol jcs dependencies

ntn-x2:


packages/did/src/Did.utils.ts
```diff
@@ -161,31 +160,38 @@ const multicodecReversePrefixes: Record<DidVerificationMethodType, number> = {
...
   }
 }
 
+// TODO: This could also be exposed in a new release candidate of the `@kiltprotocol/jcs-data-integrity-proofs-common` package.
+function multibase58BtcKeyBytesEncoding(
+  key: Uint8Array,
+  keyPrefix: number
+): Base58BtcMultibaseString {
+  const encodeVarint =
+    typeof varint.encode === 'function' ? varint.encode : varint.default.encode
```
rflechtner, line 188: I believe this may not be necessary here given that the sdk packages have the `esModulesInterop` typescript setting enabled. Even if it is required, it only needs to be run once and not every time the function is called; that is, the const declaration should be moved outside of the function.

packages/did/src/Did.utils.ts
```diff
@@ -161,31 +160,38 @@ const multicodecReversePrefixes: Record<DidVerificationMethodType, number> = {
...
+  const varintEncodedPrefix = encodeVarint(keyPrefix)
+  const prefixedKey = Uint8Array.from([...varintEncodedPrefix, ...key])
+  const base58BtcEncodedKey = base58Encode(prefixedKey)
+  return `z${Buffer.from(base58BtcEncodedKey).toString()}`
```
rflechtner, line 192: don't use Buffer here; it's not Browser-compatible. Why would we even need to turn a string into a Buffer and back into a string right after?

packages/did/src/Did.utils.ts
```diff
@@ -161,31 +160,38 @@ const multicodecReversePrefixes: Record<DidVerificationMethodType, number> = {
...
 
+// TODO: This could also be exposed in a new release candidate of the `@kiltprotocol/jcs-data-integrity-proofs-common` package.
+function multibase58BtcKeyBytesEncoding(
+  key: Uint8Array,
+  keyPrefix: number
+): Base58BtcMultibaseString {
+  const encodeVarint =
+    typeof varint.encode === 'function' ? varint.encode : varint.default.encode
+  const varintEncodedPrefix = encodeVarint(keyPrefix)
+  const prefixedKey = Uint8Array.from([...varintEncodedPrefix, ...key])
```
rflechtner, line 190: polkadot/utils has the u8aConcat function for that, I think

packages/did/src/Did.utils.ts
rflechtner: you can test by building the code, then trying to require/import and use the function from a .cjs and a .mjs test file

ntn-x2: Yes, I tried that and it worked. Good to know! https://github.com/KILTprotocol/sdk-js/pull/872/commits/bd5df5fec8b12da2f99bd204dee3f23f15954a27

packages/did/src/Did.utils.ts
ntn-x2: Yeah, good catch 😅 too much copy-pasting... https://github.com/KILTprotocol/sdk-js/pull/872/commits/bd5df5fec8b12da2f99bd204dee3f23f15954a27

packages/did/src/Did.utils.ts
ntn-x2: Yup! https://github.com/KILTprotocol/sdk-js/pull/872/commits/bd5df5fec8b12da2f99bd204dee3f23f15954a27

ntn-x2:
@rflechtner I think we agreed you'd take this over. I'll assign you as well for you to remember.

rflechtner (approved):
🚀

PR #881, opened 2024-07-17
feat: high-level did helper api

rflechtner:
## fixes https://github.com/KILTprotocol/ticket/issues/3359

Adds a high-level did api for simplifying the most common DID management operations.
 
## How to test:

Currently covered by integration- and unit tests; I'd suggest we release a beta after merging this to develop.

## Checklist:

    * Link the PR/Ticket here


packages/sdk-js/src/DidHelpers/createDid.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * Creates an on-chain DID based on an authentication key.
+ *
+ * @param options Any {@link SharedArguments} (minus `didDocument`) and additional parameters.
+ * @param options.fromPublicKey The public key that will feature as the DID's initial authentication method and will determine the DID identifier.
+ * @returns A set of {@link TransactionHandlers}.
+ */
+export function createDid(
```
Dudleyneedham, line 36: This is incorrect. This only adds a single key. We discussed adding ALL keys.

packages/sdk-js/src/DidHelpers/verificationMethod.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+  })
+
+  it('creates a remove VM tx', async () => {
+    didDocument.assertionMethod = didDocument.authentication
+    removeVerificationMethod({
+      didDocument,
+      api: mockedApi,
+      submitter: keypair,
+      signers: [keypair],
+      verificationMethodId: didDocument.assertionMethod![0],
```
Dudleyneedham, line 92: I tried this and it seems not to work. I am getting the result back of my previous operation.

`     resultSetAssertion.asConfirmed.didDocument.assertionMethod![0],`

However, it gives me the following as an example: `did:kilt:4oZJFxJJGqJJnoNsajpCQmpWg5RrbSHCYQR3bw1vCeafkHto#0xbfa824220a126fdb757036ac3718b4271976dab2e0555a4a23be4eab6257ee19`

I suspect it only wants the following:`#0xbfa824220a126fdb757036ac3718b4271976dab2e0555a4a23be4eab6257ee19`

packages/sdk-js/src/DidHelpers/createDid.ts
rflechtner: No, in the design docs this was never meant to do more than simply create a DID based on an authentication key. All customisation is to be done after that.

rflechtner: https://hackmd.io/@rflechtner/SkvnDDPVA

rflechtner: adding batching would allow us to offer doing it all in one go though

packages/sdk-js/src/DidHelpers/verificationMethod.spec.ts
rflechtner: It does work exactly like this in the integration tests though, including assertions that the did no longer has a verification method afterwards. Would be interesting to see the code you used to make sure we don't have a hidden bug here

Dudleyneedham: Yes, I was submitting the original did document in the options and not the updated document

rflechtner:
@Dudleyneedham & @ntn-x2 The Did helper implementation is now completed and all issues we found have been fixed. This PR gives you a chance to skim over the aggregated changes; I would would merge this to develop shortly and start preparing a beta release from there.
Things I would tackle via PRs to develop:
- rewrite bundle tests to use the DidHelpers; this is equally a documentation- and a test task, as we'd write the bundle tests such that they run through everything that can be done with DidHelpers to provide code examples for Chris to work with in updating the docs
- finalise the set of functions exported from sdk-js; we talked about adding CType stuff, and we considered removing the `signAndSubmit`.

Dudleyneedham (approved):
LGTM! I personally would like to see some clean up in the future. There are some internal code that are messy. Due to time constraints I would consider this debt we have accumulated. 

Secondly, after using and showing it to people it feels way better and more simple.

Great job guys.

Don't forget we need to expose the CType in the SDK js

packages/credentials/src/V1/KiltAttestationProofV1.ts
```diff
@@ -659,114 +665,99 @@ export function finalizeProof(
   }
 }
 
-export interface TransactionResult {
-  status: 'InBlock' | 'Finalized'
-  includedAt: { blockHash: Uint8Array; blockHeight?: BigInt; blockTime?: Date }
-  events?: EventRecord[]
-}
+async function defaultTxSubmit({
+  didDocument,
```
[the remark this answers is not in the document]
rflechtner: not sure I understand, what do you want to call `options`? the parameter is unnamed here because we destructure it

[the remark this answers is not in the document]
Dudleyneedham: Fair enough

PR #886, opened 2024-07-18
Refactor `checkResult()`

abdulmth:


packages/sdk-js/src/DidHelpers/checkResult.ts
```diff
@@ -0,0 +1,241 @@
...
+function checkStatus(result: SubmittableResultValue): {
+  status: TransactionResult['status'] | undefined
+  error: Error | undefined
+  blockHash: HexString | undefined
+  blockNumber: BigInt | undefined
+} {
+  let status: TransactionResult['status'] | undefined
+  let blockHash
+  let blockNumber
+  let error
```
rflechtner, line 58: let's use returns inside of the switch instead

packages/sdk-js/src/DidHelpers/checkResult.ts
```diff
@@ -0,0 +1,241 @@
...
+import type { Did, HexString } from '@kiltprotocol/types'
+import type { SharedArguments, TransactionResult } from './interfaces.js'
+import { assertStatus, mapError } from './common.js'
+
+function checkEventsForErrors(
+  api: ApiPromise,
+  txEvents: EventRecord[] = []
+): Error | undefined {
+  let error
+  txEvents.forEach(({ event }) => {
```
rflechtner, line 26:
```suggestion
  return txEvents.find(({ event }) => {
```
combine this with returns (I'll make one example below) and our function becomes simpler and faster (exits on first error)

packages/sdk-js/src/DidHelpers/checkResult.ts
```diff
@@ -0,0 +1,241 @@
...
+    if (api.events.system.ExtrinsicFailed.is(event)) {
+      error = mapError(event.data[0], api)
```
rflechtner, line 28:
```suggestion
      return mapError(event.data[0], api)
```

rflechtner: same for all other branches

packages/sdk-js/src/DidHelpers/checkResult.ts
```diff
@@ -0,0 +1,241 @@
...
+    const eventError = checkEventsForErrors(api, txEvents)
+    if (eventError !== undefined) {
+      error = eventError
+    }
+
+    const eventMatch = checkIfEventsMatch(expectedEvents, txEvents)
+    const isSuccess = !error && eventMatch
+
+    status = isSuccess ? 'confirmed' : 'failed'
+    if (!isSuccess && !error) {
```
rflechtner, line 179:
```suggestion
    if (!isSuccess && !eventMatch) {
```
why not be more explicit?

packages/sdk-js/src/DidHelpers/common.ts
```diff
@@ -9,38 +9,25 @@
...
 
 import type {
   AcceptedPublicKeyEncodings,
   SharedArguments,
   TransactionHandlers,
-  TransactionResult,
 } from './interfaces.js'
 
-function mapError(err: SpRuntimeDispatchError, api: ApiPromise): Error {
+export function mapError(err: SpRuntimeDispatchError, api: ApiPromise): Error {
```
rflechtner, line 22: this one and the next are only used in checkResult rn, so they could move there too

packages/sdk-js/src/DidHelpers/checkResult.ts
abdulmth: 5fa8bed

packages/sdk-js/src/DidHelpers/checkResult.ts
rflechtner: ah yes of course, didn't think it through enough. Yes, that'll do the trick

rflechtner (approved):
Nice!

PR #888, opened 2024-07-22
fix: transaction result serialization

abdulmth:
fixes https://github.com/KILTprotocol/ticket/issues/3529

Adds `toJSON` implementation to the `TransactionResult` objects that include a block number.
Another way to solve the serialization issue is to change the block number type to `string`.

### DIFF
--- a/packages/sdk-js/src/DidHelpers/checkResult.ts
+++ b/packages/sdk-js/src/DidHelpers/checkResult.ts
@@ -231,6 +231,7 @@ export async function checkResultImpl(
             ...clone.block,
             number: clone.block.number.toString(),
           }
+          delete clone.toJSON
           return clone
         },
       }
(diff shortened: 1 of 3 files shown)


packages/sdk-js/src/DidHelpers/checkResult.ts
```diff
@@ -251,6 +261,14 @@ export async function checkResultImpl(
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         block: { hash: blockHash!, number: blockNumber! },
         events: txEvents.map(({ event }) => event),
+        toJSON() {
+          const clone = { ...this } as any
+          clone.block = {
+            number: clone.block.number.toString(),
+            hash: this.block.hash,
+          }
```
rflechtner, line 269:
```suggestion
          clone.block = {
            ...clone.block,
            number: clone.block.number.toString(),
          }
```

rflechtner (commented):
Did you check if stringifying the result object now works? I suspect that due to the getters this will still be a problem, and we likely have to implement `toJSON()` on the root as well

abdulmth:
@rflechtner  It doesn't work because `stringify` tries to call the getters, and they will throw except one. But why do we need to implement it for the root? users should match the status anyways before doing anything, and having keys like `asConfirmed` are meant for conversion rather as a JSON key.

rflechtner:
> @rflechtner It doesn't work because `stringify` tries to call the getters, and they will throw except one. But why do we need to implement it for the root? users should match the status anyways before doing anything, and having keys like `asConfirmed` are meant for conversion rather as a JSON key.

I thought so and wondered if we should allow stringifying to either something like
```
{
   status: 'confirmed',
   value: {
     block: {...},
     ...
   }
}
```
or
```
{
   status: 'confirmed',
   block: {...},
   ...
}
```

rflechtner (approved):
👍

PR #893, opened 2024-07-30
refactor!: update sdk surface

rflechtner:
## https://github.com/KILTprotocol/ticket/issues/3502

- Removes signAndSubmitTx
- Updates some outdated docstrings

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/credentials/src/holder.ts
+++ b/packages/credentials/src/holder.ts
@@ -55,10 +55,10 @@ function pointerToAttributeName(
  * @param credential A Verifiable Credential containing a proof.
  * @param proofOptions Options for creating the derived proof.
  * @param proofOptions.disclose Allows selecting which claims/attributes in the credential should be disclosed, hiding others.
- * @param proofOptions.disclose.only An array of JSON-path (IETF RFC 6901) expressions selecting attributes to be disclosed.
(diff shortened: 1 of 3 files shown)


packages/credentials/src/holder.ts
```diff
@@ -55,10 +55,10 @@ function pointerToAttributeName(
  * @param credential A Verifiable Credential containing a proof.
  * @param proofOptions Options for creating the derived proof.
  * @param proofOptions.disclose Allows selecting which claims/attributes in the credential should be disclosed, hiding others.
- * @param proofOptions.disclose.only An array of credentialPath expressions selecting attributes to be disclosed.
+ * @param proofOptions.disclose.only An array of JSON-path (IETF RFC 6901) expressions selecting attributes to be disclosed.
```
ntn-x2, line 58: Could we link to the spec here?

ntn-x2 (approved):

abdulmth (approved):

packages/credentials/src/holder.ts
rflechtner: good idea, also bc it made me realise that it's actually JSON Path that we're using, not JSON pointers 😬

rflechtner: 4f389db22224240d0bf025a7fd03243709d2fcc0

PR #894, opened 2024-08-05
refactor!: integrate feedback on alpha release

rflechtner:
## fixes KILTProtocol/ticket#3540

- breaking: unnest selective disclosure parameters dc8d45865ca7d070f3d38a2451c16aff960a9b46
- breaking: fully switch to using named parameters on credentials functions 3c17ace73089be9b74458d6709881d576b33115b
- non-breaking: allow passing key pairs & signers as submitter parameter on `issue()` bcd882f8e5c10c0dc0d74946851072da1fd842aa

## How to test:

- Actually trying the new interfaces could be done on the beta that is to be released
- Tests for new features are yet to be written

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/credentials/src/V1/KiltAttestationProofV1.ts
+++ b/packages/credentials/src/V1/KiltAttestationProofV1.ts
@@ -49,6 +49,7 @@ import type {
   IDelegationNode,
   KiltAddress,
   SharedArguments,
+  SignerInterface,
 } from '@kiltprotocol/types'
 import { Caip19, JsonSchema, SDKErrors, Signers } from '@kiltprotocol/utils'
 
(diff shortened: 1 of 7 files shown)


abdulmth (approved):

rflechtner:
@abdulmth I'm not sure about the nesting of parameters, seems cumbersome and messy to me. How did you see it? Should we flatten parameters a bit more?

Dudleyneedham (approved):
LGTM. It feels more consistent

Dudleyneedham:
> @abdulmth I'm not sure about the nesting of parameters, seems cumbersome and messy to me. How did you see it? Should we flatten parameters a bit more?

I would argue to call them all options, that would move them up one level.

> I would keep the `issuer` and `holder` objects but maybe flatten the various options objects.

Yes, I would agree.

rflechtner:
@abdulmth had argued that the proof/presentationOptions to carry semantic value. We did decide to group options this way at some point and I'm not sure we want to change that last minute without good reason, so I'd just leave it.

PR #899, opened 2024-08-13
Checking if encodedDetails can be decoded in the parse function

MarkusGarmeister:
## fixes KILTProtocol/ticket#3033
A light DID resolution with encoded details, which can not be decoded, is parsed and decoded twice. I tried to fix this with a check of the encodedDetails in the parse function. However the tests are failling.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/did/src/Did.utils.ts
+++ b/packages/did/src/Did.utils.ts
@@ -13,10 +13,9 @@ import type {
   VerificationMethod,
 } from '@kiltprotocol/types'
 import { DataUtils, Multikey, SDKErrors, ss58Format } from '@kiltprotocol/utils'
-import { blake2AsU8a, encodeAddress } from '@polkadot/util-crypto'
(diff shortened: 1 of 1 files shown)


packages/did/src/Did.utils.ts
```diff
@@ -107,6 +106,14 @@ export function parse(did: Did | DidUrl): IDidParsingResult {
       ? parseInt(versionString, 10)
       : LIGHT_DID_LATEST_VERSION
     const queryParameters = exportQueryParamsFromDidUrl(did as DidUrl)
+    // checking if encodedDetails can be decoded
+    if (encodedDetails) {
+      try {
```
Dudleyneedham, line 111: The library has a validation function. We should be able to use that. @rflechtner What do you think?

```
base58Validate(encodedDetails, true)
```

Dudleyneedham (commented):
I think we can use functionality in the library

PR #902, opened 2024-09-04
fix: rename Docker image

ntn-x2:
The Docker image specified there never existed, and it's only used internally for cache.

### DIFF
--- a/.github/workflows/test-node-candidate.yml
+++ b/.github/workflows/test-node-candidate.yml
@@ -145,7 +145,7 @@ jobs:
       - name: pull node image
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
-          ECR_REPOSITORY: kilt/standalone-node
+          ECR_REPOSITORY: standalone-node
           IMAGE_TAG: ${{ github.event.inputs.docker-image-tag-name }}
         run: |
           docker pull $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
--- a/.github/workflows/test-node-candidate.yml
+++ b/.github/workflows/test-node-candidate.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       docker-image-tag-name:
         type: string
-        description: The tag of the kiltprotocol/prototype-chain Docker image to test against
+        description: The tag of the kiltprotocol/standalone-node Docker image to test against
         required: true
 
 env:
--- a/.github/workflows/test-node-candidate.yml
+++ b/.github/workflows/test-node-candidate.yml
@@ -76,7 +76,7 @@ jobs:
       - name: pull image
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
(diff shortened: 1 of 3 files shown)


rflechtner:
The kilt/prototype-chain image very much did exist on our ECR and we're using it everywhere in our sdk tests. 
@ggera can you confirm which image name we should be using? Can we maybe just configure ECS to act as a pull-through cache for kiltprotocol/standalone-node?

ntn-x2:
Yes it does, but as far as I can see it's only used as an intermediate cache to build `standalone-node` Docker images. If @ggera agrees, we should probably switch SDK tests to use the `standalone-node` Docker image, and in the future perhaps even a Zombienet-based setup with the `kilt-node` Docker image directly, which I am working on.

ntn-x2:
Anyway I'm a bit confused, cause [here](https://github.com/KILTprotocol/sdk-js/blob/55aa2fb34233577ffc5e90950b48e6e78a0d493e/package.json#L29) we use `kiltprotocol/standalone-node:latest-develop` instead.

ntn-x2:
It seems we are not pushing there anymore @ggera? That is what I figured by looking at our deployment pipelines, although I think it was introduced with your refactoring. Since it's used as a caching image, perhaps we want to change the integration tests to use the standalone image rather? This one has pushes from the current release process.

Screenshot 2024-09-03 at 17 09 50

ntn-x2:
It seems we are not pushing there anymore @ggera? That is what I figured by looking at our deployment pipelines, although I think it was introduced with your refactoring. Since it's used as a caching image, perhaps we want to replace it completely in our build pipeline and in the integration tests to use the standalone image rather? This one has pushes from the current release process.

Screenshot 2024-09-05 at 10 59 36

ntn-x2:
Link to the manually-triggered GitHub workflow being changed here: https://github.com/KILTprotocol/sdk-js/actions/runs/10717708953/job/29718068093. This seems to work now.

Dudleyneedham (approved):
LGTM

rflechtner:
We need a cache layer to get around dockerhub rate limits. That's why I suggested a pull through cache

rflechtner (commented):
Is `standalone-node` an image that's being updated on our ECR?

rflechtner:
If this is the image we should be using, can you update all workflows to use it?

ntn-x2:
Yes, `standalone-node` has always been updated, that's why I was confused. I still don't know if stop pushing to the `prototype-chain` image was intentional or not, but I can definitely update the image used everywhere.

ntn-x2:
@rflechtner I talked to @ggera and we don't see any issue in using the final Docker image for all tests, so I updated all references. Once he's back, we'll investigate if we need such an intermediate image, and how to deal with it.

rflechtner (approved):
Lgtm

ntn-x2:
The CI fails because the `latest` tag is broken, but both the latest `rc` as well as `develop` work, so I'll merge this. `latest` will be fixed as soon as we promote the `rc`.

PR #918, opened 2024-11-20
feat: revoke in issuer

aybarsayan:
## fixes KILTProtocol/ticket#3646

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/credentials/src/issuer.ts
+++ b/packages/credentials/src/issuer.ts
@@ -5,26 +5,28 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import type { Did, ICType, IClaimContents } from '@kiltprotocol/types'
-
-import { SDKErrors } from '@kiltprotocol/utils'
-import { KiltAttestationProofV1, KiltCredentialV1 } from './V1/index.js'
-import type { UnsignedVc, VerifiableCredential } from './V1/types.js'
-import type { CTypeLoader } from './ctype/index.js'
+import { SDKErrors, Signers } from '@kiltprotocol/utils'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import type { IssuerOptions, SubmitOverride } from './interfaces.js'
-
-export type { IssuerOptions }
-import * as Kilt from '@kiltprotocol/sdk-js'
 import * as KiltChain from '@kiltprotocol/chain-helpers'
+// import * as Kilt from '@kiltprotocol/sdk-js'
+// import * as KiltChain from '@kiltprotocol/chain-helpers'
 import type {
   KiltAddress,
   SignerInterface,
-  KeyringPair,
-  DidUrl,
+  Did,
+  ICType,
+  IClaimContents,
+  MultibaseKeyPair,
 } from '@kiltprotocol/types'
-import {authorizeTx} from "@kiltprotocol/did"
+import { authorizeTx } from '@kiltprotocol/did'
 import { base58Decode } from '@polkadot/util-crypto'
+import { ConfigService } from '@kiltprotocol/config'
+import type { IssuerOptions, SubmitOverride } from './interfaces.js'
+import type { CTypeLoader } from './ctype/index.js'
+import type { UnsignedVc, VerifiableCredential } from './V1/types.js'
+import { KiltAttestationProofV1, KiltCredentialV1 } from './V1/index.js'
+
+export type { IssuerOptions }
 
 /**
  * Creates a new credential document as a basis for issuing a credential.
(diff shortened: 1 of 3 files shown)


packages/credentials/src/issuer.ts
```diff
@@ -15,6 +15,16 @@ import type { CTypeLoader } from './ctype/index.js'
 import type { IssuerOptions, SubmitOverride } from './interfaces.js'
 
 export type { IssuerOptions }
+import * as Kilt from '@kiltprotocol/sdk-js'
+import * as KiltChain from '@kiltprotocol/chain-helpers'
```
rflechtner: As with the other PR, no imports from sdk-js here and no blanket imports.

packages/credentials/src/issuer.ts
```diff
@@ -15,6 +15,16 @@ import type { CTypeLoader } from './ctype/index.js'
...
+import type {
+  KiltAddress,
+  SignerInterface,
+  KeyringPair,
+  DidUrl,
+} from '@kiltprotocol/types'
+import {authorizeTx} from "@kiltprotocol/did"
```
rflechtner: make sure you run prettier for auto-formatting

packages/credentials/src/issuer.ts
```diff
@@ -134,3 +144,64 @@ export async function issue({
...
+* - info: Object containing blockchain transaction details:
+*   - blockNumber?: Block number where transaction was included
+*   - blockHash?: Hash of the block
+*   - transactionHash?: Hash of the transaction
+*/
+export async function revoke(
+  attester: DidUrl,
+  submitterAccount: KeyringPair,
+  signCallback: any,
+  credential: VerifiableCredential,
```
rflechtner: I'd like to only see two parameters here:
```suggestion
  credential: VerifiableCredential,
  issuer: IssuerOptions,
```
This is enough for issuance, and will also be enough for revocation.

packages/credentials/src/issuer.ts
```diff
@@ -134,3 +144,64 @@ export async function issue({
...
+    }
+    
+  } catch (error: unknown) {
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
+    return {
+      success: false,
+      error: [errorMessage],
+      info: {}
+    }
+ }
```
rflechtner: In order to make sure we can support future credential- and proof types without breaking the function interfaces, this function should be set up to detect what kind of revocation status method we are dealing with, and if it's a of the `KiltRevocationStatusV1` type, call the right logic to revoke the credential.

Can you move the implementation to `packages/credentials/src/V1/KiltRevocationStatusV1.ts`, and in the function here only check which type of credentialStatus method we are dealing with, call the function implemented in KiltRevocationStatusV1.ts if it's the right one, and error if not? the `issue` function above should give you an idea.

Btw all the info you need for revocation is contained in the credentialStatus id, which is composed as follows:
`${chainId}/kilt:attestation/${base58Encode(rootHash)}`
Because we have the chain id here as well, which contains the genesis hash of the chain on which the credential was attested, you can even check if we are connected to the right network (i.e., peregrine vs spiritnet)! See the `check` function in KiltRevocationStatusV1.ts for an example.

rflechtner (changes requested):

packages/credentials/src/issuer.ts
aybarsayan: Done at: https://github.com/KILTprotocol/sdk-js/pull/918/commits/ccac9b90e1ceac14a1c26d443d3fcef994745055

packages/credentials/src/issuer.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/918/commits/ccac9b90e1ceac14a1c26d443d3fcef994745055

packages/credentials/src/issuer.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/918/commits/ccac9b90e1ceac14a1c26d443d3fcef994745055

packages/credentials/src/issuer.ts
aybarsayan: I am working on a second commit for this.

aybarsayan: Fixed at: [110a5cc](https://github.com/KILTprotocol/sdk-js/pull/918/commits/110a5cc482545b61bf5e514493616e2ffd3aaa69)

abdulmth:
Replaced by https://github.com/KILTprotocol/sdk-js/pull/930

PR #919, opened 2024-11-20
feat: nested ctype val

aybarsayan:
## fixes KILTProtocol/ticket#3632

Added nested CType validation support to validateSubject
Implemented automatic reference detection and blockchain fetching
Added support to handle both simple and complex CType structures
Updated tests for new functionality

Note: Improves credential validation by handling nested CTypes automatically.

## How to test:

- Step 1
- Step 2
- etc.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/packages/credentials/src/V1/KiltCredentialV1.ts
+++ b/packages/credentials/src/V1/KiltCredentialV1.ts
@@ -357,7 +357,7 @@ function extractUniqueReferences(
     if (typeof value !== 'object' || value === null) {
       return
     }
-    const objValue = value as Record<string, unknown>    
+    const objValue = value as Record<string, unknown>
     if ('$ref' in objValue) {
       const ref = objValue.$ref as string
       if (ref.startsWith('kilt:ctype:')) {
--- a/packages/credentials/src/V1/KiltCredentialV1.ts
+++ b/packages/credentials/src/V1/KiltCredentialV1.ts
@@ -7,6 +7,7 @@
 
 import { hexToU8a } from '@polkadot/util'
 import { base58Encode } from '@polkadot/util-crypto'
+import * as Kilt from "@kiltprotocol/sdk-js"
 
 import { JsonSchema, SDKErrors } from '@kiltprotocol/utils'
 import type {
@@ -327,15 +328,76 @@ export function fromInput({
 
 const cachingCTypeLoader = newCachingCTypeLoader()
 
+// Helper function to check if CType is nested
+function cTypeTypeFinder(cType: ICType): boolean {
+  function hasRef(obj: any): boolean {
+    if (typeof obj !== 'object' || obj === null) return false
+    
+    if ('$ref' in obj) return true
+    
+    return Object.values(obj).some(value => {
+      if (Array.isArray(value)) {
+        return value.some(item => hasRef(item))
+      }
+      if (typeof value === 'object') {
+        return hasRef(value)
+      }
+      return false
+    })
+  }
+  return hasRef(cType.properties)
+}
+
+// Helper function to extract unique references from CType
+function extractUniqueReferences(cType: ICType): Set<string> {
+  const references = new Set<string>()
+  
+  function processValue(value: any) {
+      if (typeof value !== 'object' || value === null) return
+      
+      if ('$ref' in value) {
+          const ref = value['$ref']
+          // Extract KILT CType reference
+          if (ref.startsWith('kilt:ctype:')) {
+              // Get first part split by #/
+              const baseRef = ref.split('#/')[0]
+              references.add(baseRef)
+          }
+      }
+      
+      // Check all values of the object recursively
+      Object.values(value).forEach(v => processValue(v))
+  }
+  
+  processValue(cType.properties)
+  return references
+}
+
+
 /**
  * Validates the claims in the VC's `credentialSubject` against a CType definition.
+ * Supports both nested and non-nested CType validation.
+ *
+ * For non-nested CTypes:
+ * - Validates claims directly against the CType schema
+ * 
+ * For nested CTypes:
+ * - Automatically detects nested structure through $ref properties
+ * - Fetches referenced CTypes from the blockchain
+ * - Performs validation against the main CType and all referenced CTypes
  *
  * @param credential A {@link KiltCredentialV1} type verifiable credential.
  * @param credential.credentialSubject The credentialSubject to be validated.
  * @param credential.type The credential's types.
  * @param options Options map.
- * @param options.cTypes One or more CType definitions to be used for validation. If `loadCTypes` is set to `false`, validation will fail if the definition of the credential's CType is not given.
- * @param options.loadCTypes A function to load CType definitions that are not in `cTypes`. Defaults to using the {@link newCachingCTypeLoader | CachingCTypeLoader}. If set to `false` or `undefined`, no additional CTypes will be loaded.
+ * @param options.cTypes One or more CType definitions to be used for validation. If loadCTypes is set to false, validation will fail if the definition of the credential's CType is not given.
+ * @param options.loadCTypes A function to load CType definitions that are not in cTypes. Defaults to using the {@link newCachingCTypeLoader | CachingCTypeLoader}. If set to false or undefined, no additional CTypes will be loaded.
+ * 
+ * @throws {Error} If the credential type does not contain a valid CType id
+ * @throws {Error} If required CType definitions cannot be loaded
+ * @throws {Error} If claims do not follow the expected CType format
+ * @throws {Error} If referenced CTypes in nested structure cannot be fetched from the blockchain
+ * @throws {Error} If validation fails against the CType schema
  */
 export async function validateSubject(
   {
@@ -344,7 +406,7 @@ export async function validateSubject(
   }: Pick<KiltCredentialV1, 'credentialSubject' | 'type'>,
   {
     cTypes = [],
-    loadCTypes = cachingCTypeLoader,
+    loadCTypes = newCachingCTypeLoader(),
   }: { cTypes?: ICType[]; loadCTypes?: false | CTypeLoader } = {}
 ): Promise<void> {
   // get CType id referenced in credential
@@ -354,6 +416,7 @@ export async function validateSubject(
   if (!credentialsCTypeId) {
     throw new Error('credential type does not contain a valid CType id')
   }
+
   // check that we have access to the right schema
   let cType = cTypes?.find(({ $id }) => $id === credentialsCTypeId)
   if (!cType) {
@@ -385,6 +448,43 @@ export async function validateSubject(
       [key.substring(vocab.length)]: value,
     }
   }, {})
-  // validates against CType (also validates CType schema itself)
-  CType.verifyClaimAgainstSchema(claims, cType)
+
+  // Connect to blockchain
+  const api = Kilt.ConfigService.get('api')
+
+  // Bizim eklediğimiz doğrulama mantığı
+  const isNested = cTypeTypeFinder(cType)
+  
+  if (!isNested) {
+    await CType.verifyClaimAgainstNestedSchemas(
+      cType,
+      [],
+      claims
+    )
+  } else {
+    const references = extractUniqueReferences(cType)
+    const referencedCTypes: ICType[] = []
+    
+    for (const ref of references) {
+      try {
+        const referencedCType = await CType.fetchFromChain(ref as any)
+        if (referencedCType.cType) {
+          referencedCTypes.push(referencedCType.cType)
+        }
+      } catch (error) {
+        console.error(`Failed to fetch CType for reference ${ref}:`, error)
+        throw new Error(`Failed to fetch CType from chain: ${ref}`)
+      }
+    }
+
+    if (referencedCTypes.length === references.size) {
+      await CType.verifyClaimAgainstNestedSchemas(
+        cType,
+        referencedCTypes,
+        claims
+      )
+    } else {
+      throw new Error("Some referenced CTypes could not be fetched")
+    }
+  }
 }
--- a/packages/credentials/src/V1/KiltCredentialV1.ts
+++ b/packages/credentials/src/V1/KiltCredentialV1.ts
@@ -452,8 +452,31 @@ export async function validateSubject(
     }
   }, {})
 
+  // Create a type-safe loader function
+  const effectiveLoader = async (id: string): Promise<ICType | undefined> => {
+    // Ensure the ID is in the correct format
+    if (!id.startsWith('kilt:ctype:0x')) {
+      throw new Error(`Invalid CType ID format: ${id}`)
+    }
+
+    const typedId = id as `kilt:ctype:0x${string}`
+
+    if (typeof loadCTypes === 'function') {
+      return loadCTypes(typedId)
+    }
+    const found = cTypes.find((ct) => ct.$id === typedId)
+    if (found) {
+      return found
+    }
+    throw new Error(
+      `CType ${id} not found in provided cTypes array and CType loading is disabled`
+    )
+  }
   // Load all nested CTypes
-  const referencedCTypes = await loadNestedCTypeDefinitions(cType, loadCTypes)
+  const referencedCTypes = await loadNestedCTypeDefinitions(
+    cType,
+    effectiveLoader
+  )
 
   // Convert Set to Array and filter out any undefined or null values
   const validCTypes = Array.from(referencedCTypes).filter(
--- a/packages/credentials/src/V1/KiltCredentialV1.ts
+++ b/packages/credentials/src/V1/KiltCredentialV1.ts
@@ -327,37 +327,59 @@ export function fromInput({
 
 const cachingCTypeLoader = newCachingCTypeLoader()
 
-// Single function to both check for references and extract them
-function extractUniqueReferences(
+async function loadNestedCTypeDefinitions(
   cType: ICType,
-  references: Set<string> = new Set<string>()
-): Set<string> {
-  if (typeof cType?.properties !== 'object' || cType.properties === null) {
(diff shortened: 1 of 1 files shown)


packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,43 @@ export async function validateSubject(
...
+    await CType.verifyClaimAgainstNestedSchemas(
+      cType,
+      [],
+      claims
+    )
+  } else {
+    const references = extractUniqueReferences(cType)
+    const referencedCTypes: ICType[] = []
+    
+    for (const ref of references) {
```
Dudleyneedham: I would suggest using a `Promise.all` here to reduce the code complexity and make this more efficient.

Additionally, change this from a `for..loop` with the await inside to a map. Using the `map` `const` you can use the `Promise.all`

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -327,15 +328,76 @@ export function fromInput({
...
+        return value.some(item => hasRef(item))
+      }
+      if (typeof value === 'object') {
+        return hasRef(value)
+      }
+      return false
+    })
+  }
+  return hasRef(cType.properties)
+}
```
Dudleyneedham: No need to make another `Named` function inside this function. 

```suggestion
// Helper function to check if CType is nested
function cTypeTypeFinder(cType: ICType): boolean {

    if (typeof cType.properties !== 'object' || cType.properties === null) return false
    
    if ('$ref' in cType.properties) return true
    
    return Object.values(cType.properties).some(value => {
      if (Array.isArray(value)) {
        return value.some(item => hasRef(item))
      }
      if (typeof value === 'object') {
        return hasRef(value)
      }
      return false
    })
}
```

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -327,15 +328,76 @@ export function fromInput({
...
+
+// Helper function to extract unique references from CType
+function extractUniqueReferences(cType: ICType): Set<string> {
```
Dudleyneedham: As above, change it.

Dudleyneedham (commented):
In general, this is bad practice to create a `Named` function into another `Named` function unless you are using it multiple times. 

There is a time and place for doing it, but it isn't here.

If you are worried about the typing you can cast or try another approach.

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -7,6 +7,7 @@
 
 import { hexToU8a } from '@polkadot/util'
 import { base58Encode } from '@polkadot/util-crypto'
+import * as Kilt from "@kiltprotocol/sdk-js"
```
rflechtner: Because the credentials package is imported by the sdk-js package this creates a circular dependency and will not work. We also don't do blanket imports in the sdk implementation. Be specific, only import what you need. CType stuff is implemented in the credentials package, this should not be necessary.

packages/credentials/src/V1/KiltCredentialV1.ts
rflechtner: I understand that you're defining a function to call it recursively, but can't you just do that with the outer function directly? Also, more generally, I think all that you need is the second function; if you're already recursing through the CType, collect all the references, we _will_ need them eventually.

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -327,15 +328,76 @@ export function fromInput({
...
+    })
+  }
+  return hasRef(cType.properties)
+}
+
+// Helper function to extract unique references from CType
+function extractUniqueReferences(cType: ICType): Set<string> {
+  const references = new Set<string>()
+  
+  function processValue(value: any) {
```
rflechtner: Don't use any, use `unknown`

packages/credentials/src/V1/KiltCredentialV1.ts
rflechtner: Here I can somewhat see the point of having an inner function; alternatively, you could solve it by taking an optional second argument:
 ```suggestion
function extractUniqueReferences(cType: ICType, references=new Set<string>()): Set<string> {
```

File: packages/credentials/src/V1/KiltCredentialV1.ts [hunk 6]
```diff
@@ -344,7 +406,7 @@ export async function validateSubject(
   }: Pick<KiltCredentialV1, 'credentialSubject' | 'type'>,
   {
     cTypes = [],
-    loadCTypes = cachingCTypeLoader,
+    loadCTypes = newCachingCTypeLoader(),
```
rflechtner: what's the reason for this change?

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,43 @@ export async function validateSubject(
       [key.substring(vocab.length)]: value,
     }
   }, {})
-  // validates against CType (also validates CType schema itself)
-  CType.verifyClaimAgainstSchema(claims, cType)
+
+  // Connect to blockchain
+  const api = Kilt.ConfigService.get('api')
```
rflechtner: shouldn't be needed, we have the loadCTypes argument to load ctypes

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,43 @@ export async function validateSubject(
...
+  // Bizim eklediğimiz doğrulama mantığı
+  const isNested = cTypeTypeFinder(cType)
+  
+  if (!isNested) {
+    await CType.verifyClaimAgainstNestedSchemas(
+      cType,
+      [],
+      claims
+    )
+  } else {
```
rflechtner: skip this step; call extractUniqueReferences directly, pass each ctype id to the ctype loader, then pass the resulting definitions to verifyClaimAgainstNestedSchemas. If there are no ctype references in the ctype, this result in an empty array, just like here.

packages/credentials/src/V1/KiltCredentialV1.ts
rflechtner: Yes, use array methods here please. E.g., `await Promise.all(Array.from(references).map(ctypeId => loadCTypes(ctypeId)))`

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,43 @@ export async function validateSubject(
...
+    const references = extractUniqueReferences(cType)
+    const referencedCTypes: ICType[] = []
+    
+    for (const ref of references) {
+      try {
+        const referencedCType = await CType.fetchFromChain(ref as any)
+        if (referencedCType.cType) {
+          referencedCTypes.push(referencedCType.cType)
+        }
```
rflechtner: again, use the loadCTypes argument. If it can't be found, throw an error saying that the ctype contains a reference to another ctype which cannot be resolved.

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -354,6 +416,7 @@ export async function validateSubject(
   if (!credentialsCTypeId) {
     throw new Error('credential type does not contain a valid CType id')
   }
+
   // check that we have access to the right schema
   let cType = cTypes?.find(({ $id }) => $id === credentialsCTypeId)
   if (!cType) {
```
rflechtner: We need to somehow handle both the ctypes passed in as the `cTypes` argument, and the `loadCtypes` function, if there is any. I suggest we create a combined ctype loader function from both, which first checks if the cType whose id is being looked up is contained in the cTypes array, and if not, forwards the id to the `loadCtypes` function, which then tries to look it up.

Hint: checking if we have a CType already in `cTypes` can be done quicker if we transform the array into a Map of ctype id to ctype definition once.

rflechtner (changes requested):

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at (i missed it that it already exists): https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/77b9c3d89f213c51b6b428ade650cc04c50efc2b

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: I am working on a new commit for this.

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,35 @@ export async function validateSubject(
...
-  // validates against CType (also validates CType schema itself)
-  CType.verifyClaimAgainstSchema(claims, cType)
+
+  try {
+    const references = extractUniqueReferences(cType)
+
+    const referencedCTypes = await Promise.all(
+      Array.from(references).map(async (ref) => {
+        try {
+          const referencedCType = await cachingCTypeLoader(ref as any)
```
rflechtner: please use the function passed as the `loadCtypes` argument

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,35 @@ export async function validateSubject(
...
+    )
+
+    if (validCTypes.length === references.size) {
+      await CType.verifyClaimAgainstNestedSchemas(cType, validCTypes, claims)
+    } else {
+      throw new Error('Some referenced CTypes could not be fetched')
+    }
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Validation error:', error)
```
rflechtner: we don't use console statements in library code, that's why the lint rule exists

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +448,35 @@ export async function validateSubject(
...
+    const references = extractUniqueReferences(cType)
+
+    const referencedCTypes = await Promise.all(
+      Array.from(references).map(async (ref) => {
+        try {
+          const referencedCType = await cachingCTypeLoader(ref as any)
+          return referencedCType
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error(`Failed to fetch CType for reference ${ref}:`, error)
```
rflechtner: we don't use console statements in library code, that's why the lint rule exists

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -327,15 +327,77 @@ export function fromInput({
...
+  if ('$ref' in (value as Record<string, unknown>)) {
+    return true
+  }
+
+  if (Array.isArray(value)) {
+    return value.some((item) => hasRef(item))
+  }
+
+  return Object.values(value as Record<string, unknown>).some((v) => hasRef(v))
+}
```
rflechtner: you're not using this function any more, are you?

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -327,15 +327,77 @@ export function fromInput({
...
+      const ref = objValue.$ref as string
+      if (ref.startsWith('kilt:ctype:')) {
+        references.add(ref.split('#/')[0])
+      }
+    }
+
+    if (Array.isArray(value)) {
+      value.forEach(processValue)
+    } else {
+      Object.values(objValue).forEach(processValue)
```
rflechtner: the point of passing down the set of references to extractUniqueReferences was that you can call extractUniqueReferences in recursion instead of an inner function. You don't necessarily need to do it that way, but if you're not using this parameter, you can also just pull it back into the function as a normal variable.

packages/credentials/src/V1/KiltCredentialV1.ts
rflechtner: this is still tbd

rflechtner (commented):
Looking way better already, but there's still a few things that need to be tackled

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/bbfced24e1897748e57f40ddc5fa06267ed4962c

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/bbfced24e1897748e57f40ddc5fa06267ed4962c

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/bbfced24e1897748e57f40ddc5fa06267ed4962c

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Yes forget to remove, fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/bbfced24e1897748e57f40ddc5fa06267ed4962c

packages/credentials/src/V1/KiltCredentialV1.ts
aybarsayan: Fixed at: https://github.com/KILTprotocol/sdk-js/pull/919/commits/bbfced24e1897748e57f40ddc5fa06267ed4962c

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +429,27 @@ export async function validateSubject(
...
+  const referencedCTypes = await Promise.all(
+    Array.from(references).map(async (ref) => {
+      if (typeof loadCTypes !== 'function') {
+        throw new Error(
+          `The definition for this credential's CType ${ref} has not been passed to the validator and CType loading has been disabled`
+        )
+      }
+      return loadCTypes(ref as any)
+    })
+  )
```
rflechtner: Sorry that I'm only realising that now, but we have two more issues here that need to be addressed:

1. There is a maximum recursion depth for synchronous code in JS, which means that deeply nested structures could end up triggering a max recursion depth reached error
2. The CTypes that we fetch could also contain references, so we need to check their definitions too and fetch any CTypes that they depend on

In light of these issues I think it's a good idea to refactor extractUniqueReferences so that it is `async` and also loads the ctype definitions directly. This function could be named `loadNestedCTypeDefinitions` and would
1. Take a CType definition `cType` and a CTypeLoader `cTypeLoader` as parameters
2. Initialise an empty Set `fetchedCTypeIds`
3. Initialise an empty Set `fetchedCTypeDefinitions`
4. Define a subroutine `processCTypeProperties` which accepts ICType["properties"] and:
    1. iterates over all properties in the object
    2. If the property value is an array, call and await `processCTypeProperties` on each item in it and continue
    3. If the property key is $ref, extract the CType id from it
    4. If the extracted CType id is not yet in `fetchedCTypeIds`, push the id to it, then call `cTypeLoader` passing the id as an argument
    5. await the Promise returned by the cTypeLoader; if it's a CType, call `processCTypeProperties` on its properties; else, throw an error
5. calls `processCTypeProperties` on the CType `properties` and awaits its completion
6. returns `fetchedCTypeDefinitions`

rflechtner:
Merged into #931

PR #924, opened 2024-12-20
feat: add metadata hash checks

ntn-x2:
Not a breaking change, users can opt in based on the new feature PR introduced in https://github.com/KILTprotocol/kilt-node/pull/835. I did not write any tests yet, but I have modified some and run them against a local instance with the new feature, and everything works flawlessly.

# Checklist

- [x] Fix `yarn.lock` issue
- [x] Fix `latest-develop` Docker image issue
- [x] Write basic integration tests against the `latest-develop` image for metadata hash checks

### DIFF
--- a/packages/chain-helpers/src/blockchain/Blockchain.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.ts
@@ -11,7 +11,7 @@ import type { SignerOptions } from '@polkadot/api-base/types'
 import type { Vec } from '@polkadot/types'
 import type { Call, Extrinsic } from '@polkadot/types/interfaces'
 import type { AnyNumber, IMethod } from '@polkadot/types/types'
-import { type BN } from '@polkadot/util'
+import { u8aToHex, type BN } from '@polkadot/util'
 import {
   type ExtraInfo,
   merkleizeMetadata,
@@ -20,6 +20,7 @@ import {
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- doing this instead of import '@kiltprotocol/augment-api' to avoid creating an import at runtime
 import type * as _ from '@kiltprotocol/augment-api'
 import type {
+  HexString,
   ISubmittableResult,
   KeyringPair,
   SubmittableExtrinsic,
@@ -172,11 +173,13 @@ export async function submitSignedTx(
 
 export const dispatchTx = submitSignedTx
 
-const metadataHashes = new Map<string, Uint8Array>()
+const metadataHashes = new Map<string, HexString>()
 
 // Returns the Merkle root of the metadata as stored in the `ConfigService` cache. If not present, it computes it, stores it in the cache for future retrievals, and returns it.
-async function getMetadataHash(api: ApiPromise): Promise<Uint8Array> {
-  const metadata = api.runtimeMetadata.asV15
+async function getMetadataHash(api: ApiPromise): Promise<HexString> {
+  const metadata = await api.call.metadata.metadataAtVersion(15)
+  // TODO: Find out why using this metadata here fails to decode when calculating the Merkle root.
+  // const metadata = api.runtimeMetadata.asV15
   const { specName, specVersion } = api.runtimeVersion
   const genesisHash = await api.genesisHash
   const cacheKey = blake2AsHex(
@@ -187,7 +190,7 @@ async function getMetadataHash(api: ApiPromise): Promise<Uint8Array> {
     ])
   )
   if (metadataHashes.has(cacheKey)) {
-    return metadataHashes.get(cacheKey) as Uint8Array
+    return metadataHashes.get(cacheKey) as HexString
   }
   const merkleInfo: ExtraInfo = {
     base58Prefix: api.consts.system.ss58Prefix.toNumber(),
@@ -197,7 +200,7 @@ async function getMetadataHash(api: ApiPromise): Promise<Uint8Array> {
     tokenSymbol: api.registry.chainTokens[0],
   }
   const merkleizedMetadata = merkleizeMetadata(metadata.toHex(), merkleInfo)
-  const metadataHash = merkleizedMetadata.digest()
+  const metadataHash = u8aToHex(merkleizedMetadata.digest())
   metadataHashes.set(cacheKey, metadataHash)
   return metadataHash
 }
@@ -218,7 +221,15 @@ export async function signTx(
   { tip, checkMetadata }: { tip?: AnyNumber; checkMetadata?: boolean } = {}
 ): Promise<SubmittableExtrinsic> {
   const signOptions: Partial<SignerOptions> = checkMetadata
-    ? { tip, metadataHash: await getMetadataHash(ConfigService.get('api')) }
+    ? {
+        tip,
+        // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
+        metadataHash: await getMetadataHash(ConfigService.get('api')),
+        // Used by external signers to to know there's additional data to be included in the payload (see link above).
+        withSignedTransaction: true,
+        // Forces the tx to fail of the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
+        mode: 1,
+      }
     : { tip }
 
   if ('address' in signer) {
@@ -227,9 +238,6 @@ export async function signTx(
 
   return tx.signAsync(signer.id, {
     ...signOptions,
-    // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
-    withSignedTransaction: true,
-    mode: 1,
     signer: Signers.getPolkadotSigner([signer]),
   })
 }
--- a/packages/chain-helpers/src/blockchain/Blockchain.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.ts
@@ -175,11 +175,11 @@ export const dispatchTx = submitSignedTx
 
 const metadataHashes = new Map<string, HexString>()
 
-// Returns the Merkle root of the metadata as stored in the `ConfigService` cache. If not present, it computes it, stores it in the cache for future retrievals, and returns it.
+// Returns the Merkle root of the metadata as stored in the local `metadataHashes` cache. If not present, it computes it, stores it in the cache for future retrievals, and returns it.
 async function getMetadataHash(api: ApiPromise): Promise<HexString> {
-  const metadata = await api.call.metadata.metadataAtVersion(15)
+  // const metadata = await api.call.metadata.metadataAtVersion(15)
   // TODO: Find out why using this metadata here fails to decode when calculating the Merkle root.
-  // const metadata = api.runtimeMetadata.asV15
+  const metadata = api.runtimeMetadata.asV15
   const { specName, specVersion } = api.runtimeVersion
   const genesisHash = await api.genesisHash
   const cacheKey = blake2AsHex(
@@ -222,14 +222,14 @@ export async function signTx(
 ): Promise<SubmittableExtrinsic> {
   const signOptions: Partial<SignerOptions> = checkMetadata
     ? {
-        tip,
-        // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
-        metadataHash: await getMetadataHash(ConfigService.get('api')),
-        // Used by external signers to to know there's additional data to be included in the payload (see link above).
-        withSignedTransaction: true,
-        // Forces the tx to fail of the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
-        mode: 1,
-      }
+      tip,
+      // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
+      metadataHash: await getMetadataHash(ConfigService.get('api')),
+      // Used by external signers to to know there's additional data to be included in the payload (see link above).
+      withSignedTransaction: true,
+      // Forces the tx to fail of the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
+      mode: 1,
+    }
     : { tip }
 
   if ('address' in signer) {
--- a/packages/chain-helpers/src/blockchain/Blockchain.ts
+++ b/packages/chain-helpers/src/blockchain/Blockchain.ts
@@ -225,7 +225,7 @@ export async function signTx(
       metadataHash: await getMetadataHash(ConfigService.get('api')),
       // Used by external signers to to know there's additional data to be included in the payload (see link above).
       withSignedTransaction: true,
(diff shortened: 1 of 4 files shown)


ntn-x2:
@rflechtner I am not understanding what I am doing wrong. I have inconsistent results running `yarn install` and `yarn build` on my machine and in the CI.

Dudleyneedham:
Check your node version. Also, check what version of Yarn you are running compared to the package.json

ntn-x2:
@rflechtner after changing the dependency type from peer dependency back to regular dependency, everything seems to be building ok and tests seem to pass (minor what was already broken on develop). I would then ask you to provide your feedback, or if you think the dependency should be a peer dependency, to try to fix it yourself as I honestly have no clue.

rflechtner:
> @rflechtner after changing the dependency type from peer dependency back to regular dependency, everything seems to be building ok and tests seem to pass (minor what was already broken on develop). I would then ask you to provide your feedback, or if you think the dependency should be a peer dependency, to try to fix it yourself as I honestly have no clue.

Peer dependencies have no effect besides issuing warnings in the CLI, so my guess the dependency was not actually being installed. I'm fine with leaving it as a regular dependency, given that it's from an entirely different stack of dependencies than the `@polkadot/api` related code. On this note though, is there no support from the `@polkadot/api` suite for checking the metadata hash? I'm not sure it's a great idea to start mixing the pjs and papi toolchains.

ntn-x2:
@rflechtner gotcha 👌

As for the library, unfortunately pjs itself does not offer the feature. The available options are the ones listed in [this](https://forum.polkadot.network/t/new-polkadot-ledger-app/8817#adding-support-of-the-new-ledger-app-to-your-wallet-10) Polkadot Forum. Even the Polkadot Apps code relies on the `@polkadot/api` library for this one feature: https://github.com/polkadot-js/apps/blob/f63da86b9a29bc9a99aaa18c75db61c242b9d9fd/packages/react-signer/src/signers/LedgerSigner.ts#L15.

ntn-x2:
Anyway, in order to test this, we first must fix an issue with our build pipeline that is failing to push new Docker images on DockerHub (just found out). @ggera is on it. Once that's resolved, I will try to push an empty commit to re-trigger the tests, and see if everything works fine. This should also fix the broken pipeline currently on `develop`, since the fixes for the missing runtime APIs have not been picked up cause there was no image pushed.

rflechtner:
> @rflechtner gotcha 👌
> 
> As for the library, unfortunately pjs itself does not offer the feature. The available options are the ones listed in [this](https://forum.polkadot.network/t/new-polkadot-ledger-app/8817#adding-support-of-the-new-ledger-app-to-your-wallet-10) Polkadot Forum. Even the Polkadot Apps code relies on the `@polkadot/api` library for this one feature: https://github.com/polkadot-js/apps/blob/f63da86b9a29bc9a99aaa18c75db61c242b9d9fd/packages/react-signer/src/signers/LedgerSigner.ts#L15.

I see! https://www.npmjs.com/package/merkleized-metadata seems lighter on additional dependencies, but choose whichever you think gets the job done.

rflechtner:
> Once that's resolved, I will try to push an empty commit to re-trigger the tests, and see if everything works fine.

You can also just hit the 'Re-run jobs' button on the actions status page (https://github.com/KILTprotocol/sdk-js/actions/runs/12633893068?pr=924)

Dudleyneedham:
@rflechtner The only concern with this package it doesn't seem maintained. I offered this yesterday but I think we should find another package if possible that is maintained or stick with Papi.

ntn-x2:
I think re-running the same job does not pull the latest version of a Docker image, does it? I think I tried this once, but sure I will try first to re-run the job once the new images are available.

ntn-x2:
> @rflechtner The only concern with this package it doesn't seem maintained. I offered this yesterday but I think we should find another package if possible that is maintained or stick with Papi.

+1 on this. I think we should stick with papi.

ntn-x2:
@rflechtner resetting to `develop` and re-installing dependencies still changes all hashes. So I force-pushed changes to the files and `package.json`, but not `yarn.lock`. Would you be able to fix that, and then I could go ahead with writing some integration tests for it?

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -24,6 +25,7 @@ import type {
 import { ConfigService } from '@kiltprotocol/config'
 import { SDKErrors, Signers } from '@kiltprotocol/utils'
 
+import { merkleizeMetadata } from '@polkadot-api/merkleize-metadata'
```
rflechtner: this could be moved to the block of other polkadot dependencies

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -174,19 +176,47 @@ export const dispatchTx = submitSignedTx
...
 ): Promise<SubmittableExtrinsic> {
+  const signOptions = await (async (): Promise<Partial<SignerOptions>> => {
+    if (!checkMetadata) {
+      return {
+        tip,
+      }
+    }
+    // If `checkMetadata` is enabled, include that when signing the tx.
+    const api = ConfigService.get('api')
+    const metadata = await api.call.metadata.metadataAtVersion(15)
```
rflechtner: the api should have the metadata cached, can't we use this one?

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -174,19 +176,47 @@ export const dispatchTx = submitSignedTx
...
+      specName: specName.toString(),
+      specVersion: specVersion.toNumber(),
+      tokenSymbol: api.registry.chainTokens[0],
+    }
+    const merkleizedMetadata = merkleizeMetadata(metadata.toHex(), merkleInfo)
+    const metadataHash = u8aToHex(merkleizedMetadata.digest())
+    return {
+      tip,
+      mode: 1,
+      withSignedTransaction: true,
```
rflechtner: what do these do?

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -174,19 +176,47 @@ export const dispatchTx = submitSignedTx
...
+    const { specName, specVersion } = api.runtimeVersion
+    const merkleInfo = {
+      base58Prefix: api.consts.system.ss58Prefix.toNumber(),
+      decimals: api.registry.chainDecimals[0],
+      specName: specName.toString(),
+      specVersion: specVersion.toNumber(),
+      tokenSymbol: api.registry.chainTokens[0],
+    }
+    const merkleizedMetadata = merkleizeMetadata(metadata.toHex(), merkleInfo)
+    const metadataHash = u8aToHex(merkleizedMetadata.digest())
```
rflechtner: the metadata hash calculation seems pretty computation heavy, we could maybe cache this in a map from genesis hash and runtime version to metadataHash?

ntn-x2: When would you populate the cache? First time we do this or we compute well known values statically? Where would we store such a map? In the config service?

rflechtner: I would keep it local to this package, and only populate the cache when a hash is requested but a cache miss occurs

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: Good point: https://github.com/KILTprotocol/sdk-js/pull/924/commits/7b9e1e4294504ba65287bf6db174849069d3a998.

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/924/commits/09c5876416ec955aaa0afa56427670de65798042

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: To be honest I am not 100% sure, but I know they are required to allow signers to include the metadata hash check. I added a comment to the only line I could find with a bit more details: https://github.com/KILTprotocol/sdk-js/pull/924/commits/09c5876416ec955aaa0afa56427670de65798042.

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: I gave it a try. What do you think? I'm using the `ConfigService` map (I could not think of anything else), and using (genesis hash, spec name, and spec version) as the key and the metadata hash as the value: https://github.com/KILTprotocol/sdk-js/pull/924/commits/dfd17af1f20716866adb67d07eb1d937a01f44e8

ntn-x2: Sorry, your comment came in after I wrote mine. I think we agree on how the cache should be populated, which is what I did here. Are you against using the `ConfigService` for that? Is the way I implemented it an abuse of it?

rflechtner: oh, yes, you can just declare a Map or object in the root of the file and use that instead - using the ConfigService is a bit of an abuse

rflechtner: e.g. `const cachedMetadataHashes = new Map<string, string>()`

ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/924/commits/22c3af8d938ed0f7c3acb1ec5f551731ccf5774b

ntn-x2:
After spending a lot of time on this and not understanding why tests would pass when using Chopsticks but not when using the standalone-node binary, I came across [this notice](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html#runtime-integration) that says that the metadata check extension does not work with native runtimes. At the moment, we still use the `NativeElseWasmExecutor` for both our binaries, which means this feature is untestable until we switch to a pure `WasmExecutor`. Actually, the feature is unusable in general until we switch to the `WasmExecutor`, but until this PR is merged, nothing breaks as the default behaviour is to ignore the metadata hash.

ntn-x2:
@rflechtner the CI is still broken due to missing runtime API decorations, but you can verify the changes work by running the new `Blockchain` integration tests locally.

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -167,26 +173,71 @@ export async function submitSignedTx(
 
 export const dispatchTx = submitSignedTx
 
+const metadataHashes = new Map<string, HexString>()
+
+// Returns the Merkle root of the metadata as stored in the `ConfigService` cache. If not present, it computes it, stores it in the cache for future retrievals, and returns it.
```
rflechtner: Comment is outdated

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -167,26 +173,71 @@ export async function submitSignedTx(
...
+async function getMetadataHash(api: ApiPromise): Promise<HexString> {
+  const metadata = await api.call.metadata.metadataAtVersion(15)
```
rflechtner: No need to fetch this if we have a cache hit, move down so it's skipped in that case

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -167,26 +173,71 @@ export async function submitSignedTx(
...
+    ? {
+        tip,
+        // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
+        metadataHash: await getMetadataHash(ConfigService.get('api')),
+        // Used by external signers to to know there's additional data to be included in the payload (see link above).
+        withSignedTransaction: true,
+        // Forces the tx to fail of the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
+        mode: 1,
+      }
+    : { tip }
```
rflechtner: Yess I like this so much better now

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -167,26 +173,71 @@ export async function submitSignedTx(
...
+    return metadataHashes.get(cacheKey) as HexString
+  }
+  const merkleInfo: ExtraInfo = {
+    base58Prefix: api.consts.system.ss58Prefix.toNumber(),
+    decimals: api.registry.chainDecimals[0],
+    specName: specName.toString(),
+    specVersion: specVersion.toNumber(),
+    tokenSymbol: api.registry.chainTokens[0],
+  }
+  const merkleizedMetadata = merkleizeMetadata(metadata.toHex(), merkleInfo)
```
rflechtner, line 200: This seems to be the earliest point where the metadata is needed

rflechtner (commented):
I have yet to test it, but the code looks great ! Just two little things you could adjust

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/924/commits/f5174bd3a7958e778d330a6b7828eaedccd16d64

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: Yes sorry I left the PR with one more open point I did not remember about. I changed the fetching here: https://github.com/KILTprotocol/sdk-js/pull/924/commits/4b05d7251ee8fc96bc36296cca0b92d20f75c95e.

There is still an error if we try to use the locally retrieved metadata, and I believe there should be one transformation step somehow involved, but I think for now we could go ahead and merge this, and we can optimise this in the future. Also, the fetching happens once and it's then cached, so not too big of a deal.

packages/chain-helpers/src/blockchain/Blockchain.ts
ntn-x2: https://github.com/KILTprotocol/sdk-js/pull/924#discussion_r1912965775

packages/chain-helpers/src/blockchain/Blockchain.ts
rflechtner: > I think for now we could go ahead and merge this, and we can optimise this in the future. Also, the fetching happens once and it's then cached, so not too big of a deal.

Yes, my thoughts exactly

packages/chain-helpers/src/blockchain/Blockchain.ts
rflechtner: https://github.com/KILTprotocol/sdk-js/pull/924#discussion_r1913013747

packages/chain-helpers/src/blockchain/Blockchain.ts
rflechtner: ok fair enough

packages/chain-helpers/src/blockchain/Blockchain.ts
```diff
@@ -167,26 +173,69 @@ export async function submitSignedTx(
...
+  { tip, checkMetadata }: { tip?: AnyNumber; checkMetadata?: boolean } = {}
 ): Promise<SubmittableExtrinsic> {
+  const signOptions: Partial<SignerOptions> = checkMetadata
+    ? {
+      tip,
+      // Required as described in https://github.com/polkadot-js/api/blob/109d3b2201ea51f27180e34dfd883ec71d402f6b/packages/api-base/src/types/submittable.ts#L79.
+      metadataHash: await getMetadataHash(ConfigService.get('api')),
+      // Used by external signers to to know there's additional data to be included in the payload (see link above).
+      withSignedTransaction: true,
+      // Forces the tx to fail of the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
```
rflechtner:
```suggestion
      // Forces the tx to fail if the metadata does not match (added for backward compatibility). See https://paritytech.github.io/polkadot-sdk/master/frame_metadata_hash_extension/struct.CheckMetadataHash.html.
```

rflechtner (approved):
💯

PR #930, opened 2025-01-14
feat: Implement 'revoke' for `Issuer`

abdulmth:
## replaces https://github.com/KILTprotocol/sdk-js/pull/918


packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
 import {
   assertMatchingConnection,
+  defaultTxSubmit,
   getDelegationNodeIdForCredential,
 } from './common.js'
-import type { KiltCredentialV1, KiltRevocationStatusV1 } from './types.js'
+import type { IssuerOptions } from '../interfaces.js'
+import type { KiltCredentialV1 } from './types.js'
+
+import { type KiltRevocationStatusV1 } from './types.js'
```
rflechtner:
```suggestion
import type { IssuerOptions } from '../interfaces.js'
import type { KiltCredentialV1, KiltRevocationStatusV1 } from './types.js'
```

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
 
 export type Interface = KiltRevocationStatusV1
 
 export const STATUS_TYPE = 'KiltRevocationStatusV1'
 
+/**
+ * @param credentialStatus The credential status propoerty of the Verifiable credential.
```
rflechtner:
```suggestion
 * @param credentialStatus The `credentialStatus` property of the Verifiable Credential.
```

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
+import type { IssuerOptions } from '../interfaces.js'
+import type { KiltCredentialV1 } from './types.js'
+
+import { type KiltRevocationStatusV1 } from './types.js'
 
 export type Interface = KiltRevocationStatusV1
 
 export const STATUS_TYPE = 'KiltRevocationStatusV1'
 
+/**
```
rflechtner, line 30: There's a general description of the function missing, such as "Revokes a Verifiable Credential containing a 'KiltRevocationStatusV1' type credential status method."

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
+ * @param credentialStatus The credential status propoerty of the Verifiable credential.
+ * @param opts Additional parameters.
+ * @param opts.api An optional polkadot-js/api instance connected to the blockchain network on which the credential is anchored.
+ * @param params.issuer Interfaces for interacting with the issuer identity.
+ * @param params.issuer.didDocument The DID Document of the issuer revoking the credential.
+ * @param params.issuer.signers Array of signer interfaces for credential authorization.
+ * @param params.issuer.submitter The submitter can be one of:
+ * - A MultibaseKeyPair for signing transactions
+ * - A Ed25519 type keypair for blockchain interactions
```
rflechtner: Why ed25519 specifically? Also you should specify that we're talking about a `KeyringPair` here

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
+    throw new Error(
+      `Cannot handle revocation status checks for asset type ${assetNamespace}:${assetReference}`
+    )
+  }
+  if (!assetInstance) {
+    throw new SDKErrors.CredentialMalformedError(
+      "The attestation record's CAIP-19 identifier must contain an asset index ('token_id') decoding to the credential root hash"
+    )
+  }
+  const rootHash = base58Decode(assetInstance)
```
rflechtner: This part is what's copied from the status check, right? Should we pull this out into a helper function to avoid duplication? Seems like something that could get out of sync, if we were to change the assetNamespace or -reference we expect for example, which would be bad.

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -9,23 +9,94 @@ import { u8aEq, u8aToHex, u8aToU8a } from '@polkadot/util'
...
+
+  let result = await transactionPromise
+  if ('status' in result) {
+    if (result.status !== 'confirmed') {
+      throw new SDKErrors.SDKError(
+        `Unexpected transaction status ${result.status}; the transaction should be "confirmed" for issuance to continue`
+      )
+    }
+    result = result.asConfirmed
+  }
```
rflechtner: I think we should inspect and raise any errors indicated by the handler, maybe like this:
```suggestion
  if ('status' in result) {
    let error: Error | undefined
    switch (result.status) {
      case 'confirmed':
        return
      case 'failed':
        error = result.asFailed.error
        break
      case 'rejected':
        error = result.asRejected.error
        break
      case 'unknown':
        error = result.asUnknown.error
        break
      default:
        break
    }
    throw (
      error ??
      new SDKErrors.SDKError(
        `Revocation failed with transaction status ${result?.status}`
      )
    )
  }
```

packages/credentials/src/issuer.ts
```diff
@@ -134,3 +137,42 @@ export async function issue({
...
+ * @param params.issuer.signers An array of signer interfaces, each allowing to request signatures made with a key associated with the issuer DID Document.
+ * The function will select the first signer that matches requirements around signature algorithm and relationship of the key to the DID as given by the DID Document.
+ * @param params.issuer.submitter Some proof types require making transactions to effect state changes on the KILT blockchain.
+ * The blockchain account whose address is specified here will be used to cover all transaction fees and deposits due for this operation.
+ * As transactions to the blockchain need to be signed, `signers` is expected to contain a signer interface where the `id` matches this address.
+ *
+ * Alternatively, you can pass a {@link SubmitOverride} callback that takes care of Did-authorizing and submitting the transaction.
+ * If you are using a service that helps you submit and pay for transactions, this is your point of integration to it.
+ *
+ *  @throws If the credential format is invalid or the  revocation fails.
```
rflechtner:
```suggestion
 *  @throws If the credential format is invalid or the revocation fails.
```

rflechtner (commented):
The docstrings need some love but otherwise I think this looks pretty damn good!

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: 053473dd2371bb15206e0aeed5bae4bebb87fda2

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: 053473dd2371bb15206e0aeed5bae4bebb87fda2 not sure the function name is good enough

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: 053473dd2371bb15206e0aeed5bae4bebb87fda2  That was a leftover from the old PR.

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -6,26 +6,95 @@
...
 } from './common.js'
+import type { IssuerOptions } from '../interfaces.js'
 import type { KiltCredentialV1, KiltRevocationStatusV1 } from './types.js'
 
 export type Interface = KiltRevocationStatusV1
 
 export const STATUS_TYPE = 'KiltRevocationStatusV1'
 
+/**
+ * Revokes a Verifiable Credential containing a 'KiltRevocationStatusV1.
```
rflechtner:
```suggestion
 * Revokes a Verifiable Credential containing a KiltRevocationStatusV1.
```

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -6,26 +6,95 @@
...
+ *
+ * @param credentialStatus The `credentialStatus` property of the Verifiable Credential.
+ * @param opts Additional parameters.
+ * @param opts.api An optional polkadot-js/api instance connected to the blockchain network on which the credential is anchored.
```
rflechtner: not sure if strictly necessary, but I think it's better if we keep the parameter documentation in the same order as the actual parameters

packages/credentials/src/V1/KiltRevocationStatusV1.ts
```diff
@@ -6,26 +6,95 @@
...
+ */
+export async function revoke(
+  credentialStatus: KiltRevocationStatusV1,
+  issuer: IssuerOptions,
+  opts: { api?: ApiPromise } = {}
+): Promise<void> {
+  const rootHash = getRootHash(credentialStatus, opts)
+  const { api = ConfigService.get('api') } = opts
+  const { didDocument, signers, submitter } = issuer
+
```
rflechtner, line 52: maybe add a TODO that in order to support revocations by someone else than the issuer (which can happen in case of delegations) the second parameter in this function would have to be populated as well.

packages/credentials/src/V1/common.ts
```diff
@@ -8,11 +8,26 @@
...
+import type {
+  HexString,
+  KiltAddress,
+  SharedArguments,
+  SignerInterface,
+} from '@kiltprotocol/types'
+import { Caip19, Caip2, SDKErrors, Signers } from '@kiltprotocol/utils'
 
-import type { KiltAttesterDelegationV1, KiltCredentialV1 } from './types.js'
+import { Extrinsic } from '@polkadot/types/interfaces/types.js'
```
rflechtner: importing from a nested file within a package is not a thing we should be doing in modern JS. You _can_ import from a subpath declared in the package's package.json (via the `exports` field), but if your import is coming from another package and has a file extension, you're doing something wrong.

packages/credentials/src/V1/common.ts
```diff
@@ -154,3 +169,90 @@ export function credentialIdFromRootHash(
...
   return `${KILT_CREDENTIAL_IRI_PREFIX}${base58Encode(bytes, false)}`
 }
+
+/**
+ * @param root0
+ * @param root0.didDocument
+ * @param root0.call
+ * @param root0.signers
+ * @param root0.submitter
+ */
```
rflechtner: remove or replace with actual documentation.

packages/credentials/src/V1/common.ts
```diff
@@ -154,3 +169,90 @@ export function credentialIdFromRootHash(
...
+  const result = await Blockchain.submitSignedTx(extrinsic, {
+    resolveOn: Blockchain.IS_FINALIZED,
+  })
+  const blockHash = result.status.asFinalized
+  return { block: { hash: blockHash.toHex() } }
+}
+
+/**
+ * @param credentialStatus
+ * @param issuer
```
rflechtner: I don't see an issuer parameter

packages/credentials/src/V1/common.ts
```diff
@@ -8,11 +8,26 @@
...
+  SharedArguments,
+  SignerInterface,
+} from '@kiltprotocol/types'
+import { Caip19, Caip2, SDKErrors, Signers } from '@kiltprotocol/utils'
 
-import type { KiltAttesterDelegationV1, KiltCredentialV1 } from './types.js'
+import { Extrinsic } from '@polkadot/types/interfaces/types.js'
+import { authorizeTx, signersForDid } from '@kiltprotocol/did'
+import { Blockchain } from '@kiltprotocol/chain-helpers'
+import { SimplifiedTransactionResult } from '../interfaces.js'
```
rflechtner:
```suggestion
import type { SimplifiedTransactionResult } from '../interfaces.js'
```

packages/credentials/src/issuer.ts
```diff
@@ -5,14 +5,17 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import type { Did, ICType, IClaimContents } from '@kiltprotocol/types'
-
 import { SDKErrors } from '@kiltprotocol/utils'
-import { KiltAttestationProofV1, KiltCredentialV1 } from './V1/index.js'
-import type { UnsignedVc, VerifiableCredential } from './V1/types.js'
-import type { CTypeLoader } from './ctype/index.js'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
rflechtner: this linter exception is disconnected from its original target now, so can probably go?

packages/credentials/src/V1/KiltRevocationStatusV1.ts
rflechtner: how about `getRootHashFromStatusId` ?

rflechtner (commented):
Just a few nitpicks now, then we're done here 😬

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/KiltRevocationStatusV1.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/common.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/common.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/issuer.ts
abdulmth: f1a54b26e2d9d9a452962f70c01ee00cf3d6e434

packages/credentials/src/V1/common.ts
abdulmth: bc6ad6351fae542b918141b3dc015fd1eefbaad4

packages/credentials/src/V1/common.ts
abdulmth: bc6ad6351fae542b918141b3dc015fd1eefbaad4

rflechtner (approved):

PR #931, opened 2025-01-16
fix: nested ctype loading

rflechtner:
## Fixes https://github.com/KILTprotocol/ticket/issues/3632

fork of https://github.com/KILTprotocol/sdk-js/pull/919

Integration tests are not currently working (https://github.com/KILTprotocol/sdk-js/pull/944)

## Checklist:

    * Link the PR/Ticket here


rflechtner:
@aybarsayan I moved nested CType loading to `CTypeLoader.ts` and saw an opportunity to reuse the caching CType loader for combining the CType definitions in `validateSubject`s `cTypes` parameter with its `loadCTypes` parameter.

packages/credentials/src/ctype/CTypeLoader.ts
```diff
@@ -32,9 +37,69 @@ export function newCachingCTypeLoader(
...
+    }
+
+    // Process all values in the object. Also works for arrays
+    await Promise.all(Object.values(value).map(extractRefsFrom))
+  }
+
+  await extractRefsFrom(cType.properties)
+
+  return fetchedCTypeDefinitions
+}
```
Dudleyneedham, line 105: What if we have self-referential? or a CType that references another one that references the first? 

We would be stuck in a loop. Can we prevent this by checking if a CType `$id` exists already? hmm, I guess this doesn't work.

What if we have a max depth? I mean, shouldn't we already have restrictions in the Meta-Schema or on-chain. Therefore, we should use this otherwise, we can get an infinite loop.

packages/credentials/src/ctype/CTypeLoader.ts
```diff
@@ -32,9 +37,69 @@ export function newCachingCTypeLoader(
...
+          }
+
+          await extractRefsFrom(referencedCType.properties)
+        }
+      }
+      return
+    }
+
+    // Process all values in the object. Also works for arrays
+    await Promise.all(Object.values(value).map(extractRefsFrom))
```
Dudleyneedham, line 99: I had asked chat about this. 

Not sure if we care to much about proformance. However, it gave me an interesting insight into this `Promise.all` and `map`.

```ts
await Promise.all(Object.values(value).map(extractRefsFrom))
```
- The issue is that extractRefsFrom is async, but map() does not handle async functions well—it creates a list of `Promise<void>`, which Promise.all then waits for.
- While this works in many cases, it may cause performance issues if many deep CType references are being processed simultaneously.
- Solution: You may consider sequential processing for deeply nested structures using for...of:


```ts
for (const subValue of Object.values(value)) {
  await extractRefsFrom(subValue)
}
```

This prevents too many concurrent fetches.

packages/credentials/src/ctype/CTypeLoader.ts
```diff
@@ -32,9 +37,69 @@ export function newCachingCTypeLoader(
   })
 
   async function getCType(id: ICType['$id']): Promise<ICType> {
```
Dudleyneedham, line 39: Randomly adding more soundness checks?

packages/credentials/src/ctype/CTypeLoader.ts
```diff
@@ -5,13 +5,15 @@
...
+import type { ICType } from '@kiltprotocol/types'
+import { SDKErrors } from '@kiltprotocol/utils'
 
 import { fetchFromChain } from './CType.chain.js'
+import { isICType, verifyDataStructure } from './CType.js'
 
 export type CTypeLoader = (id: ICType['$id']) => Promise<ICType>
 
-const loadCType: CTypeLoader = async (id) => {
+const chainCTypeLoader: CTypeLoader = async (id) => {
```
Dudleyneedham, line 16: Not entirely sure why the name change `loadCType` sounded better. I think people would understand that more.

I understand this fetches from the chain. But should this matter in the eyes of a user?

Dudleyneedham: I suggest we revert this. If I wanted to get a ctype I would have a hard time to find this.

The action first then subject second makes more sense, considering the user might fully understand the subject but the action is easy to grasp.

packages/credentials/src/V1/KiltCredentialV1.ts
```diff
@@ -385,6 +382,28 @@ export async function validateSubject(
       [key.substring(vocab.length)]: value,
     }
   }, {})
+
+  // Turn CType loader & ctypes array into combined loader function
+  const combinedCTypeLoader = newCachingCTypeLoader(
+    cTypes,
+    typeof loadCTypes === 'function'
```
Dudleyneedham, line 389: I never knew `function` had a typeof. 

Is this new?

Dudleyneedham (changes requested):
Minor comments and request to revert a name change.

packages/credentials/src/ctype/CTypeLoader.ts
rflechtner: > What if we have self-referential? or a CType that references another one that references the first?
> 
> We would be stuck in a loop. Can we prevent this by checking if a CType `$id` exists already? hmm, I guess this doesn't work.

It does work, and we do it on line 82. So no getting stuck in a loop.

packages/credentials/src/ctype/CTypeLoader.ts
rflechtner: Bizarre that chad thinks performance would be better if we process them sequentially. Async code exists exactly for this use case, where it doesn't make much sense to block execution while waiting for the full node to return historic blocks. In case there are issues with making too many requests at the same time, I'd recommend rate limiting on the ctype loader side, but not here.

packages/credentials/src/ctype/CTypeLoader.ts
rflechtner: I moved them from the code for loading nested ctype definitions, as they seemed misplaced there.

packages/credentials/src/ctype/CTypeLoader.ts
rflechtner: I can't tell you exactly why I renamed this, probably because it hated the look of the default argument definition
```
cTypeLoader = loadCType
```
Doesn't really tell you much, while `chainCTypeLoader` at least explains where the CTypes are loaded from.
That said, besides maybe showing up in the function signature as the default argument, this does not come up in user land as it is not exported from this file. But if you hate it I can revert.

packages/credentials/src/V1/KiltCredentialV1.ts
rflechtner: Mhh I haven't checked, but my assumption is that it's ancient

packages/credentials/src/ctype/CTypeLoader.ts
Dudleyneedham: Ok, I did a lot of digging to see how it all works. I understand now. My brain hurts.

packages/credentials/src/ctype/CTypeLoader.ts
Dudleyneedham: Yes, it is strange. There wasn't any issues on my side, but I like to run it through chat afterwards. Seeing what interesting things it comes up with. I sent the test results. It basically a pointless comment without the context of what we are doing in the project.

packages/credentials/src/ctype/CTypeLoader.ts
Dudleyneedham: Ok fair enough.

packages/credentials/src/ctype/CTypeLoader.ts
Dudleyneedham: Hmm, but it seems worse to me now `CTypeLoader.chainCTypeLoader`. 

Let us put this into the dev channel and see.

packages/credentials/src/V1/KiltCredentialV1.ts
Dudleyneedham: Ok, I checked it was around since JS was around. Never knew.

Dudleyneedham (approved):
LGTM

packages/credentials/src/ctype/CTypeLoader.ts
rflechtner: again, it's not exported, so there's neither `CTypeLoader.chainCTypeLoader` nor `CTypeLoader.loadCType`. It's not accessible from the outside at all

Dudleyneedham: Ok, let us leave it then :)

PR #929, opened 2025-01-13
fix: CI pipeline

ntn-x2:
Disable compression for dependencies for reproducible install steps, and fix the CI issue with bundle test preparation.

Integration tests with `develop` are still broken since we have not yet released the updated type definitions.

PR #921, opened 2024-12-05
chore: pull image via ecr pull cache

ggera:
## fixes KILTProtocol/ticket#3600
`standalone-node` is pushed only to docker-hub & to avoid rate limit I introduced `ecr-pull-cache`
## How to test:
 `docker pull $AWS_ACCOUNT.dkr.ecr.$REGION.amazonaws.com/dh/kiltprotocol/standalone-node`
## Checklist:

- [ ] I have verified that the code works
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)
    * Either PR or Ticket to update [the Docs](https://github.com/KILTprotocol/docs)
    * Link the PR/Ticket here

PR #903, opened 2024-09-04
chore: replace all uses of tx.balances.transfer()

rflechtner:
## re/ KILTProtocol/ticket#3582

Replaces all uses of the deprecated balances.transfer extrinsic. 

## Checklist:

- [x] I have verified that the code works
- [ ] I have verified that the code is easy to understand
  - [ ] If not, I have left a well-balanced amount of inline comments
- [ ] I have [left the code in a better state](https://deviq.com/principles/boy-scout-rule)
- [ ] I have documented the changes (where applicable)
    * Either PR or Ticket to update [the Docs](https://github.com/KILTprotocol/docs)
    * Link the PR/Ticket here

PR #932, opened 2025-01-16
test: fix tests & ci

rflechtner:
Enables and fixes CI/tests for PRs to `v0-lts`, the legacy sdk support branch.
Includes backports of #929, #921, #903 & https://github.com/KILTprotocol/sdk-js/pull/871.

It seems that the type definitions here are already out of date, leading to CI failure. This is fixed in https://github.com/KILTprotocol/sdk-js/pull/933.

## Checklist:

    * Link the PR/Ticket here

### DIFF
--- a/.github/actions/cached-image-pull/action.yml
+++ b/.github/actions/cached-image-pull/action.yml
@@ -27,7 +27,7 @@ runs:
 
     - name: check image cache
       id: image-cache
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: ${{ steps.image-file.outputs.path }}
         key: image-cache-${{ inputs.image }}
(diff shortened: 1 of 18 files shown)


ntn-x2 (approved):

ntn-x2:
It seems the Docker image does not exist?

rflechtner:
> It seems the Docker image does not exist?

The docker image seems fine to me but we are getting the same errors on `latest` which we started getting on `latest-develop` since the update to the new did runtime apis. These disappear when using the the current release of type-definitions. Not entirely sure what's going on here, as this would mean that all our applications should already be broken in production?

rflechtner:
> It seems the Docker image does not exist?

```
2025-01-16 17:36:41        API/INIT: kilt-kestrel/11405: Not decorating unknown runtime apis: 0x26609555c0656603/2, 0xa47b7d544994c99b/1, 0xfbc577b9d747efd6/1
```

That's the hint; the reason it can't find the right definitions might be that we changed the chain name for the standalone chain and hadn't updated this in the types for the legacy sdk. That would also explain why the production apps didn't break.

PR #945, opened 2025-02-20
chore: use api v15

rflechtner:
We have the opportunity to bump pjs dependencies on the v0 branch to v15 (v13 for `common` packages) before the release of 0.36.0, which would give us a `@polkadot/api` version with support for metadata v15. This eliminates the need for updates to the type-definitions package, and thus does away with having to bump these in all apps whenever a new runtime api version is released.

So far we planned to release with v11, which is the minimum version compatible with the new runtime, but this seems short-sighted; I'm sure that now that an sdk upgrade is required to support the new runtime, people would prefer to update once and be done with it for the near future, instead of having to make package bumps again when we make changes to runtime apis.


Dudleyneedham (approved):
LGTM. 

Not much to check

PR #949, opened 2025-03-06
feat: adding public key to the multikey signer generate keypair

Dudleyneedham:
Adding the public key u8a to the multikey signer generate keypair.


rflechtner (changes requested):
I think we talked about this: the `MultibaseKeyPair` type shouldn't be changed, instead extend the type for the return type of `generateKeypair`

